PR #8, opened 2018-05-04
added a Travis configuration file, fixed Clippy lints

afck:


DrPeterVanNostrand:
If you can't apt-get the protoc binary, I have an open PR with protoc installation instructions that you can add to a shell script `./.travis/install_protoc.sh` and then add the following to the travis.yml:

```
before_install:
    ./.travis/install_protoc.sh
```

vkomenda:
Hi @DrPeterVanNostrand! There is no .travis/ in https://github.com/poanetwork/hbbft/pull/9/
Please update your PR.

DrPeterVanNostrand:
Hi @vkomenda ! I meant that when installing binaries with travis, generally you have to do some combination of curl/wget + chmod + mv + rm, it is common to put these instructions in a small shell script in the directory `.travis`.

So, if apt-get install didn't work, @afck could just copy the Debian `protoc` installation instructions in the README from [#9](https://github.com/poanetwork/hbbft/pull/9) into a shell script (for example `.travis/install_protoc.sh`, but the file could be located anywhere), and just call that script in `before_install` step.

afck:
Thanks for the information! You're right, the version in Ubuntu seems to be too old (on my Debian stable here it worked just fine). I'll follow your instructions.

afck:
That passed: https://travis-ci.com/poanetwork/hbbft/builds/72582202
(Not sure why GitHub isn't showing it.)

This would now also make Travis check Clippy and Rustfmt, but if you find them annoying, I'm happy with just the tests, too, of course.

PR #11, opened 2018-05-07
The ACS and BA algorithms

vkomenda:
I've defined the ACS algorithm and an interface to it consisting of 3 functions: new, send_proposed_value and on_input. The latter is a message handler for RB and BA messages. These two types of message are therefore joined in the types of message inputs and outputs.

Note about BA: A common coin implementation is not provided in this pull request. The coin value is constant 1 at https://github.com/poanetwork/hbbft/blob/15353e8d27398fa745a99a148f74276c62ec1d11/src/agreement.rs#L174.

### DIFF
--- a/proto/message.proto
+++ b/proto/message.proto
@@ -45,8 +45,9 @@ message LemmaProto {
 }
 
 message AgreementProto {
+  uint32 epoch = 1;
   oneof payload {
-    bool bval = 1;
-    bool aux = 2;
+    bool bval = 2;
+    bool aux = 3;
   }
 }
--- a/proto/message.proto
+++ b/proto/message.proto
@@ -1,53 +1,52 @@
 syntax = "proto3";
 
 message MessageProto {
-  oneof payload {
-    BroadcastProto broadcast = 1;
-    AgreementProto agreement = 2;
-  }
+    oneof payload {
+        BroadcastProto broadcast = 1;
+        AgreementProto agreement = 2;
+    }
 }
 
 message BroadcastProto {
-  oneof payload {
-    ValueProto value = 1;
-    EchoProto echo = 2;
-    ReadyProto ready = 3;
-  }
+    oneof payload {
+        ValueProto value = 1;
+        EchoProto echo = 2;
+        ReadyProto ready = 3;
+    }
 }
 
 message ValueProto {
-  ProofProto proof = 1;
+    ProofProto proof = 1;
 }
 
 message EchoProto {
-  ProofProto proof = 1;
+    ProofProto proof = 1;
 }
 
 message ReadyProto {
-  bytes root_hash = 1;
+    bytes root_hash = 1;
 }
 
 message ProofProto {
-  bytes root_hash = 1;
-  LemmaProto lemma = 2;
-  bytes value = 3;
+    bytes root_hash = 1;
+    LemmaProto lemma = 2;
+    bytes value = 3;
 }
 
 message LemmaProto {
-  bytes node_hash = 1;
-  LemmaProto sub_lemma = 2;
+    bytes node_hash = 1;
+    LemmaProto sub_lemma = 2;
 
-  oneof sibling_hash {
-    bytes left_sibling_hash = 3;
-    bytes right_sibling_hash = 4;
-  }
-
+    oneof sibling_hash {
+        bytes left_sibling_hash = 3;
+        bytes right_sibling_hash = 4;
+    }
 }
 
 message AgreementProto {
-  uint32 epoch = 1;
-  oneof payload {
-    bool bval = 2;
-    bool aux = 3;
-  }
+    uint32 epoch = 1;
+    oneof payload {
+        bool bval = 2;
+        bool aux = 3;
+    }
 }
--- a/proto/message.proto
+++ b/proto/message.proto
@@ -1,52 +1,52 @@
 syntax = "proto3";
 
 message MessageProto {
-    oneof payload {
-        BroadcastProto broadcast = 1;
-        AgreementProto agreement = 2;
-    }
+  oneof payload {
+    BroadcastProto broadcast = 1;
+    AgreementProto agreement = 2;
+  }
 }
 
 message BroadcastProto {
-    oneof payload {
-        ValueProto value = 1;
-        EchoProto echo = 2;
-        ReadyProto ready = 3;
-    }
+  oneof payload {
+    ValueProto value = 1;
+    EchoProto echo = 2;
+    ReadyProto ready = 3;
+  }
 }
 
 message ValueProto {
-    ProofProto proof = 1;
+  ProofProto proof = 1;
 }
 
 message EchoProto {
-    ProofProto proof = 1;
+  ProofProto proof = 1;
 }
 
 message ReadyProto {
-    bytes root_hash = 1;
+  bytes root_hash = 1;
 }
 
 message ProofProto {
-    bytes root_hash = 1;
-    LemmaProto lemma = 2;
-    bytes value = 3;
+  bytes root_hash = 1;
+  LemmaProto lemma = 2;
+  bytes value = 3;
 }
 
 message LemmaProto {
-    bytes node_hash = 1;
-    LemmaProto sub_lemma = 2;
+  bytes node_hash = 1;
+  LemmaProto sub_lemma = 2;
 
-    oneof sibling_hash {
-        bytes left_sibling_hash = 3;
-        bytes right_sibling_hash = 4;
-    }
+  oneof sibling_hash {
+    bytes left_sibling_hash = 3;
+    bytes right_sibling_hash = 4;
+  }
 }
 
 message AgreementProto {
-    uint32 epoch = 1;
-    oneof payload {
-        bool bval = 2;
-        bool aux = 3;
-    }
+  uint32 epoch = 1;
+  oneof payload {
+    bool bval = 2;
+    bool aux = 3;
+  }
 }
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -1,5 +1,6 @@
 //! Binary Byzantine agreement protocol from a common coin protocol.
 
+use itertools::Itertools;
 use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
 use std::hash::Hash;
 
@@ -210,17 +211,12 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
     /// can, however, expect every good node to send an AUX value that will
     /// eventually end up in our bin_values.
     fn count_aux(&self) -> (usize, BTreeSet<bool>) {
-        let mut count = 0;
-        let vals: BTreeSet<bool> = self.received_aux
+        let (vals_cnt, vals) = self.received_aux
             .values()
-            .filter(|b| {
-                count += 1;
-                self.bin_values.contains(b)
-            })
-            .cloned()
-            .collect();
+            .filter(|b| self.bin_values.contains(b))
+            .tee();
 
-        (count, vals)
+        (vals_cnt.count(), vals.cloned().collect())
     }
 
     /// Waits until at least (N − f) AUX_r messages have been received, such that
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -48,7 +48,7 @@ impl AgreementMessage {
 }
 
 pub struct Agreement<NodeUid> {
-    /// The UID of the corresponding node.
+    /// The UID of the corresponding proposer node.
     uid: NodeUid,
     num_nodes: usize,
     num_faulty_nodes: usize,
@@ -113,27 +113,27 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
     /// to remote nodes. There can be up to 2 messages.
     pub fn on_input(
         &mut self,
-        uid: NodeUid,
+        sender_id: NodeUid,
         message: &AgreementMessage,
     ) -> Result<AgreementOutput, Error> {
         match *message {
             // The algorithm instance has already terminated.
             _ if self.terminated => Err(Error::Terminated),
 
-            AgreementMessage::BVal((epoch, b)) if epoch == self.epoch => self.on_bval(uid, b),
+            AgreementMessage::BVal((epoch, b)) if epoch == self.epoch => self.on_bval(sender_id, b),
 
-            AgreementMessage::Aux((epoch, b)) if epoch == self.epoch => self.on_aux(uid, b),
+            AgreementMessage::Aux((epoch, b)) if epoch == self.epoch => self.on_aux(sender_id, b),
 
             // Epoch does not match. Ignore the message.
             _ => Ok((None, VecDeque::new())),
         }
     }
 
-    fn on_bval(&mut self, uid: NodeUid, b: bool) -> Result<AgreementOutput, Error> {
+    fn on_bval(&mut self, sender_id: NodeUid, b: bool) -> Result<AgreementOutput, Error> {
         let mut outgoing = VecDeque::new();
 
         self.received_bval
-            .entry(uid)
+            .entry(sender_id)
             .or_insert_with(BTreeSet::new)
             .insert(b);
         let count_bval = self.received_bval
@@ -179,11 +179,11 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
         }
     }
 
-    fn on_aux(&mut self, uid: NodeUid, b: bool) -> Result<AgreementOutput, Error> {
+    fn on_aux(&mut self, sender_id: NodeUid, b: bool) -> Result<AgreementOutput, Error> {
         let mut outgoing = VecDeque::new();
 
         self.received_aux
-            .entry(uid)
+            .entry(sender_id)
             .or_insert_with(BTreeSet::new)
             .insert(b);
         if !self.bin_values.is_empty() {
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -178,8 +178,9 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
                 self.received_aux.insert(self.uid.clone(), b);
             }
 
-            outgoing.extend(self.try_coin());
-            Ok((self.output, outgoing))
+            let (decision, maybe_message) = self.try_coin();
+            outgoing.extend(maybe_message);
+            Ok((decision, outgoing))
         }
         // upon receiving BVAL_r(b) messages from f + 1 nodes, if
         // BVAL_r(b) has not been sent, multicast BVAL_r(b)
@@ -200,8 +201,9 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
         self.received_aux.insert(sender_id.clone(), b);
         let mut outgoing = VecDeque::new();
         if !self.bin_values.is_empty() {
-            outgoing.extend(self.try_coin());
-            Ok((self.output, outgoing))
+            let (decision, maybe_message) = self.try_coin();
+            outgoing.extend(maybe_message);
+            Ok((decision, outgoing))
         } else {
             Ok((None, outgoing))
         }
@@ -233,14 +235,14 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
     /// of either an AUX_r or a BVAL_r message).
     ///
     /// Once the (N - f) messages are received, gets a common coin and uses it
-    /// to compute the next decision estimate and, optionally, sets the output
-    /// decision value. The function may start the next epoch. In that case, it
+    /// to compute the next decision estimate and outputs the optional decision
+    /// value.  The function may start the next epoch. In that case, it also
     /// returns a message for broadcast.
-    fn try_coin(&mut self) -> VecDeque<AgreementMessage> {
+    fn try_coin(&mut self) -> (Option<bool>, VecDeque<AgreementMessage>) {
         let (count_aux, vals) = self.count_aux();
         if count_aux < self.num_nodes - self.num_faulty_nodes {
             // Continue waiting for the (N - f) AUX messages.
-            return VecDeque::new();
+            return (None, VecDeque::new());
         }
 
         // FIXME: Implement the Common Coin algorithm. At the moment the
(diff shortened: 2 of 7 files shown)


src/common_subset.rs
```diff
@@ -116,57 +113,95 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
+                if let Some(mut agreement_instance) = self.agreement_instances.get_mut(&uid) {
+                    // Optional output of agreement and outgoing agreement
+                    // messages to remote nodes.
+                    result = if agreement_instance.terminated() {
+                        // This instance has terminated and does not accept input.
+                        Ok((None, VecDeque::new()))
+                    } else {
+                        // Send the message to the agreement instance.
+                        agreement_instance
+                            .on_input(uid.clone(), &amessage)
```
afck: That shouldn't be the same `uid` as above, should it? If I understand correctly:

* `self.agreement_instances.get_mut(&uid)` is the instance that decides whether the element proposed by node `uid` should be part of the common subset in the end.
* The parameter of `Agreement::on_input` should be the `uid` of the node that sent this particular message.

src/common_subset.rs
```diff
@@ -116,57 +113,95 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
+                            .map_err(Error::from)
+                    }
+                }
+
+                if let Ok((output, mut outgoing)) = result {
+                    if let Some(b) = output {
+                        outgoing.append(&mut self.on_agreement_result(uid, b));
```
afck: Is that `&mut` a typo?

src/common_subset.rs
```diff
@@ -116,57 +113,95 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
         // Upon delivery of value 1 from at least N − f instances of BA, provide
         // input 0 to each instance of BA that has not yet been provided input.
         if result {
-            self.agreement_true_outputs.insert(uid);
+            self.agreement_results.insert(uid, result);
+            // The number of instances of BA that output 1.
+            let results1: usize =
+                self.agreement_results
+                    .iter()
+                    .fold(0, |count, (_, v)| if *v { count + 1 } else { count });
```
afck: That could also be simplified with `filter(|v| **v).count()`, I think.

src/common_subset.rs
```diff
@@ -176,24 +211,19 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
-            .collect();
-        let completed_uids: HashSet<NodeUid> = self.agreement_results
+        if self.agreement_instances
             .iter()
-            .map(|(k, _)| k.clone())
-            .collect();
-        if instance_uids == completed_uids {
-            // All instances of Agreement that delivered `true`.
-            let delivered_1: HashSet<NodeUid> = self.agreement_results
+            .all(|(_, instance)| instance.terminated())
```
afck: If only the values are needed, `values()` should be preferred over `iter()`:

```rust
self.agreement_instances.values().all(Agreement::terminated)
```

src/proto/mod.rs
```diff
@@ -74,10 +74,12 @@ impl<T: Send + Sync + fmt::Debug> fmt::Debug for BroadcastMessage<T> {
...
 /// Messages sent during the binary Byzantine agreement stage.
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum AgreementMessage {
-    BVal(bool),
-    Aux(bool),
+    /// BVAL message with an epoch.
+    BVal((u32, bool)),
+    /// AUX message with an epoch.
+    Aux((u32, bool)),
```
afck: Maybe we should move the `AgreementMessage` type into the `agreement` module. I think the individual algorithms' modules could be pretty self-contained.

proto/message.proto
```diff
@@ -45,8 +45,9 @@ message LemmaProto {
 }
 
 message AgreementProto {
+  uint32 epoch = 1;
   oneof payload {
-    bool bval = 1;
-    bool aux = 2;
+    bool bval = 2;
+    bool aux = 3;
```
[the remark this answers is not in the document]
vkomenda: rustfmt doesn't work with anything but Rust sources:
```
$ rustfmt proto/message.proto 
error: expected one of `!` or `::`, found `=`
 --> /home/vk/src/poanetwork/hbbft/proto/message.proto:1:8
  |
1 | syntax = "proto3";
  |        ^ expected one of `!` or `::` here

```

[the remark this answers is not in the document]
afck: Ah, sorry! You're right, of course, and indentation 2 is fine in a proto file!

src/agreement.rs
```diff
@@ -1,42 +1,241 @@
...
 
+use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
+use std::hash::Hash;
+
 use proto::AgreementMessage;
-use std::collections::{BTreeSet, VecDeque};
 
-#[derive(Default)]
-pub struct Agreement {
+pub struct Agreement<NodeUid> {
```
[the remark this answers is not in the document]
vkomenda: Added.

proto/message.proto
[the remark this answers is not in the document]
vkomenda: Changed indentation to 4 spaces and committed.

src/agreement.rs
```diff
@@ -1,42 +1,241 @@
...
+    /// Bin values. Reset on every epoch update.
+    bin_values: BTreeSet<bool>,
+    /// Values received in BVAL messages. Reset on every epoch update.
+    received_bval: HashMap<NodeUid, BTreeSet<bool>>,
+    /// Sent BVAL values. Reset on every epoch update.
+    sent_bval: BTreeSet<bool>,
+    /// Values received in AUX messages. Reset on every epoch update.
+    received_aux: HashMap<NodeUid, BTreeSet<bool>>,
+    /// All the output values in all epochs.
+    outputs: BTreeMap<u32, bool>,
```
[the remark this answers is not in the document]
vkomenda: Agreement can output multiple times. That's why the termination criterion is more complex than whether the instance has output a value. I have to correct the code to reflect that. At the moment only one value is stored per agreement instance.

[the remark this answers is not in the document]
afck: Right, that's a bit ambiguous in the paper. I think the reason for the complex termination criterion is just that a node needs to keep sending messages to help the other nodes output, too. The formulation "then output b" should really be read as "then output b, unless you already have done so". For the user of the module it will certainly be more convenient if it only outputs once?

src/agreement.rs
```diff
@@ -1,42 +1,241 @@
...
+    /// of either an AUX_r or a BVAL_r message).
+    ///
+    /// `try_coin` output an optional combination of the agreement value and the
+    /// agreement broadcast message.
+    fn try_coin(&mut self) -> (Option<bool>, Option<AgreementMessage>) {
+        let (count_aux, vals) = self.count_aux();
+        if count_aux >= self.num_nodes - self.num_faulty_nodes {
+            // FIXME: Implement the Common Coin algorithm. At the moment the
+            // coin value is constant `true`.
+            let coin: u64 = 1;
```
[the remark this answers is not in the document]
vkomenda: I've made it random over {true, false}. This way BA is essentially the same as in Mostefaoui et al. [42].

src/agreement.rs
```diff
@@ -1,42 +1,241 @@
...
+
+            let coin2 = coin % 2 != 0;
```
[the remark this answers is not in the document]
vkomenda: Done.

vkomenda:
@afck, I've changed `on_input_agreement` considerably while working on your passing remark on `uid`. Please check whether you find any issues with the new version. The rationale behind the new version is that any incoming BA message should be delivered to all local BA instances since all BA messages are broadcasts, hence there is no explicit destination of a BA message.

proto/message.proto
```diff
@@ -1,52 +1,52 @@
...
 
 message MessageProto {
-  oneof payload {
-    BroadcastProto broadcast = 1;
-    AgreementProto agreement = 2;
-  }
+    oneof payload {
+        BroadcastProto broadcast = 1;
+        AgreementProto agreement = 2;
+    }
```
afck: Sorry about my remark! You were right, of course, and we should leave the indentation at 2 spaces instead of 4 here, as is the standard for proto files.

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
+    /// so waiting for N - f agreeing messages would not always terminate. We
+    /// can, however, expect every good node to send an AUX value that will
+    /// eventually end up in our bin_values.
+    fn count_aux(&self) -> (usize, BTreeSet<bool>) {
+        let vals = BTreeSet::new();
+        (
+            self.received_aux
+                .values()
+                .filter(|values| values.is_subset(&self.bin_values))
+                .map(|values| vals.union(values))
```
afck: `BTreeSet::union` doesn't modify the receiver, so `vals` will remain empty. This should probably say:

```rust
vals.extend(values)
```

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
     input: Option<bool>,
-    _bin_values: BTreeSet<bool>,
+    /// Bin values. Reset on every epoch update.
+    bin_values: BTreeSet<bool>,
+    /// Values received in BVAL messages. Reset on every epoch update.
+    received_bval: HashMap<NodeUid, BTreeSet<bool>>,
+    /// Sent BVAL values. Reset on every epoch update.
+    sent_bval: BTreeSet<bool>,
+    /// Values received in AUX messages. Reset on every epoch update.
+    received_aux: HashMap<NodeUid, BTreeSet<bool>>,
```
afck: We could make this a `HashMap<NodeUid, bool>`, since multiple `Aux` messages from the same node can be safely ignored. (I think… can they?)

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
+    /// messages are received, thus this condition may be triggered upon arrival
+    /// of either an AUX_r or a BVAL_r message).
+    ///
+    /// `try_coin` outputs an optional combination of the agreement value and
+    /// the agreement broadcast message.
+    fn try_coin(&mut self) -> (Option<bool>, Option<AgreementMessage>) {
+        let (count_aux, vals) = self.count_aux();
+        if count_aux < self.num_nodes - self.num_faulty_nodes {
+            // Continue waiting for the (N - f) AUX messages.
+            (None, None)
```
afck: You could make this an early `return (None, None)`, then the rest of the method wouldn't have to be indented.

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
+        } else {
+            // FIXME: Implement the Common Coin algorithm. At the moment the
+            // coin value is random and local to each instance of Agreement.
+            let coin2 = random::<bool>();
```
afck: Using `random` means the coin won't be "common" at all: each node will have a different value. I think that breaks the guarantee that it terminates, even without any malicious nodes. The instances take the coin value as a hint when to terminate, so an unlucky node may outlive all its remote counterparts, and never be able to output anything.

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
+            // some round r' > r."
+            self.terminated = self.terminated || self.estimated.values().any(|b| *b == coin2);
+
+            // Prepare to start the next epoch.
+            self.bin_values.clear();
+
+            if vals.len() != 1 {
+                // Start the next epoch.
+                self.epoch += 1;
+                (None, Some(self.set_input(coin2)))
```
afck: This could also `return`, removing some indentation below.

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
+                let mut message = None;
+                // NOTE: `vals` has exactly one element due to `vals.len() == 1`
+                let output: Vec<Option<bool>> = vals.into_iter()
+                    .take(1)
+                    .map(|b| {
+                        message = Some(self.set_input(b));
+
+                        if b == coin2 {
+                            // Record the output to perform a termination check later.
+                            self.estimated.insert(self.epoch, b);
```
afck: I'm confused about the terminology: Is `input` here the "estimate" in the paper, and `self.estimated` the "output" in the paper?

src/agreement.rs
```diff
@@ -1,42 +1,283 @@
...
+                // Start the next epoch.
+                self.epoch += 1;
+                (None, Some(self.set_input(coin2)))
+            } else {
+                let mut message = None;
+                // NOTE: `vals` has exactly one element due to `vals.len() == 1`
+                let output: Vec<Option<bool>> = vals.into_iter()
+                    .take(1)
+                    .map(|b| {
+                        message = Some(self.set_input(b));
```
afck: Does `self.epoch` need to be incremented before this call? I imagine the `BVal` messages this sends should have the _next_ epoch.

src/common_subset.rs
```diff
@@ -8,28 +8,33 @@ use std::fmt::{Debug, Display};
...
 /// Input from a remote node to Common Subset.
 pub enum Input<NodeUid> {
     /// Message from a remote node `uid` to the broadcast instance `uid`.
     Broadcast(NodeUid, BroadcastMessage<ProposedValue>),
-    /// Message from a remote node `uid` to the agreement instance `uid`.
-    Agreement(NodeUid, AgreementMessage),
+    /// Message from a remote node `message_sender_id` concerning the common
+    /// subset element proposed by the node `element_proposer_id`.
+    Agreement {
+        message_sender_id: NodeUid,
```
afck: The message sender's ID doesn't need to be part of the message itself, I think, since it's passed into `on_input` anyway.

src/common_subset.rs
```diff
@@ -101,11 +105,8 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
 
     /// Upon delivery of v_j from RBC_j, if input has not yet been provided to
     /// BA_j, then provide input 1 to BA_j. See Figure 11.
-    pub fn on_broadcast_result(
-        &mut self,
-        uid: &NodeUid,
-    ) -> Result<Option<AgreementMessage>, Error> {
-        if let Some(agreement_instance) = self.agreement_instances.get_mut(uid) {
+    fn on_broadcast_result(&mut self, uid: &NodeUid) -> Result<Option<AgreementMessage>, Error> {
+        if let Some(agreement_instance) = self.agreement_instances.get_mut(&uid) {
```
afck: I wonder whether it's dangerous to change an agreement instance's estimate value later than round 0. Could this prevent termination? (Not sure about that.)

src/common_subset.rs
```diff
@@ -116,57 +117,117 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
+        &mut self,
+        msg_sender_id: NodeUid,
+        element_proposer_id: NodeUid,
+        amessage: &AgreementMessage,
+    ) -> Result<CommonSubsetOutput<NodeUid>, Error> {
+        // The result defaults to error.
+        let mut result = Err(Error::NoSuchAgreementInstance);
+
+        // Send the message to the local instance of Agreement
+        if let Some(mut agreement_instance) = self.agreement_instances.get_mut(&element_proposer_id)
```
afck: Instead of the mutable `result`, you could just return here on error:

```rust
let agreement_instance = self.agreement_instances
    .get_mut(&element_proposer_id)
    .ok_or(Error::NoSuchAgreementInstance)?;
```

src/common_subset.rs
```diff
@@ -116,57 +117,117 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
-            self.agreement_true_outputs.insert(uid);
+            self.agreement_results.insert(element_proposer_id, result);
+            // The number of instances of BA that output 1.
+            let results1 = self.agreement_results.values().filter(|v| **v).count();
 
-            if self.agreement_true_outputs.len() >= self.num_nodes - self.num_faulty_nodes {
-                let instances = &mut self.agreement_instances;
-                for (_uid0, instance) in instances.iter_mut() {
+            if results1 >= self.num_nodes - self.num_faulty_nodes {
+                for instance in self.agreement_instances.values_mut() {
```
afck: See above: I'm worried that this could break the agreement if the instance has already progressed beyond the first round. (But it's not quite clear in the paper and I haven't thought much about it yet.)

src/agreement.rs
vkomenda: I think an agreement instance can output different AUX values in different epochs. Note that the values in AUX messages depend on the current state of bin_values which is subject to change.

afck: Sorry, I meant multiple `Aux` messages _in a single epoch_ can be safely ignored. This container should be cleared in each round change, shouldn't it?

src/agreement.rs
vkomenda: `Agreement::estimated` is the mapping `est` in the paper. I renamed `estimated` from `outputs` following your earlier comment in a previous commit. `Agreement::input` is the input from ACS. The corresponding construct to "output" from BA in the paper is the first component `Option<bool>` in `AgreementOutput`. I hope this clarifies the terminology. I'm also going to add a few comments.

src/common_subset.rs
vkomenda: I think you are confusing CommonSubset::on_input which does not receive the ID as argument with Agreement::on_input which does. The latter gets the sender ID from the message received by CommonSubset::on_input.

src/agreement.rs
vkomenda: You are right. It makes sense to increment the epoch earlier.

src/common_subset.rs
afck: Why can't `CommonSubset::on_input` just pass on the sender ID to `Agreement::on_input`? It looks dangerous to me to have it in the message, since it would basically allow the sender to lie about their identity.

vkomenda: Ah, no, the sender ID is in `common_subset::Input` which is not a message but rather a type of input to the Common Subset algorithm. As we discussed before, `AgreementMessage` does not have any IDs.

afck: Right, I see! But then what will actually be transmitted over the network, i.e. what's the "common subset" message type? Because that will need to contain `element_proposer_id`, but not `message_sender_id`, I think. So wouldn't it be more natural to replace `Input` with `Message`, and change the signature of `CommonSubset::on_input` so that it takes the sender ID separately from the message?

src/common_subset.rs
vkomenda: I implemented this in a stricter way. There is a check `has_input` which returns true iff the first estimated value has been written. `set_input` will return an error if the input has already been provided.

src/agreement.rs
```diff
@@ -216,12 +210,13 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
...
-            .fold(BTreeSet::new(), |vals, values| {
-                vals.union(values).cloned().collect()
-            });
+        let mut vals: BTreeSet<bool> = BTreeSet::new();
+
+        for b in self.received_aux.values() {
+            if self.bin_values.contains(b) {
+                vals.insert(b.clone());
+            }
+        }
```
afck: But now `vals.len()` will always be `<= 2` (number of values instead of number of entries). Maybe it should be:

```rust
let aux_count = self.received_aux.values()
    .filter(|b| self.bin_values.contains(b))
    .count();
(aux_count, vals)
```

vkomenda: How do you define `vals`?

afck: The same way you already do.

vkomenda: If so then `aux_count == vals.len()`. So, `aux_count` is redundant.

src/common_subset.rs
vkomenda: Actually the mutable result helps to avoid a multiple mutable borrow conflict.

src/agreement.rs
```diff
@@ -210,15 +210,17 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
...
-        for b in self.received_aux.values() {
-            if self.bin_values.contains(b) {
-                vals.insert(b.clone());
-            }
-        }
+        let mut count = 0;
+        let vals: BTreeSet<bool> = self.received_aux
+            .values()
+            .filter(|b| {
+                count += 1;
```
afck: Now it will count the total number of entries in `received_aux`, instead of just the ones whose values are in `bin_values`.

src/common_subset.rs
vkomenda: In the paper, BA instances can set the estimated value only for the next epoch. So, if an instance hasn't been provided with input `true` from the ACS instance, the estimated value of the epoch 0 will remain empty until the ACS instance fills it in using `Agreement::set_input(false)`. Then that BA can terminate as soon as the coin value is `false` due to the termination criterion.

src/agreement.rs
```diff
@@ -1,42 +1,278 @@
...
+        }
+
+        // FIXME: Implement the Common Coin algorithm. At the moment the
+        // coin value is common across different nodes but not random.
+        let coin2 = (self.epoch % 2) == 0;
+
+        // Check the termination condition: "continue looping until both a
+        // value b is output in some round r, and the value Coin_r' = b for
+        // some round r' > r."
+        self.terminated = self.terminated || self.estimated.values().any(|b| *b == coin2);
```
afck: It's not sufficient that any previous estimate agrees with the coin. The _output_ has to agree with it.

src/agreement.rs
```diff
@@ -1,42 +1,278 @@
...
+    /// Values received in BVAL messages. Reset on every epoch update.
+    received_bval: HashMap<NodeUid, BTreeSet<bool>>,
+    /// Sent BVAL values. Reset on every epoch update.
+    sent_bval: BTreeSet<bool>,
+    /// Values received in AUX messages. Reset on every epoch update.
+    received_aux: HashMap<NodeUid, bool>,
+    /// Estimates of the decision value in all epochs. The first estimated value
+    /// is provided as input by Common Subset using the `set_input` function
+    /// which triggers the algorithm to start.
+    estimated: BTreeMap<u32, bool>,
```
afck: I don't think we actually need to remember the previous estimates: Once we move on to epoch `i + 1`, the value from epoch `i` can safely be forgotten.

I think we could even just make `estimate` a simple `bool` and initialize it with `false` (without sending messages). `Agreement::on_input` would then do nothing if `epoch > 0` or input has already been provided before (i.e. we'd need another `has_input: bool`).

We'd also need a `has_output: bool`. After that's `true`, the estimate can't change again (we don't even need to cater for that, as it's guaranteed by the algorithm itself, I think), and we can terminate as soon as in any future round `has_output && coin == estimate`.

vkomenda: For the termination criterion, I mostly agree but I think the paper asks for `self.output == Some(coin)`. In that case do we need `estimated` at all? It is never read.

vkomenda: My mistake, `estimated` is used for broadcast.

afck: No, I think you're right: It _is_ used for broadcasting `BVal`, but only at the point where we transition to a new epoch (or receive user input in epoch 0). After that we can forget it, so we don't actually have to store it in the struct at all. (But then we'd have to make `output` an `Option<bool>`, as you say. Either way is fine with me.)

vkomenda: Please have a look at my last commit. I added a BVal message at the start of every epoch r, r > 0. I think the paper clearly requires those.

src/agreement.rs
```diff
@@ -198,11 +198,12 @@ impl<NodeUid: Clone + Eq + Hash> Agreement<NodeUid> {
 
     fn handle_aux(&mut self, sender_id: &NodeUid, b: bool) -> Result<AgreementOutput, Error> {
         self.received_aux.insert(sender_id.clone(), b);
+        let mut outgoing = VecDeque::new();
         if !self.bin_values.is_empty() {
-            self.try_coin();
-            Ok((self.output, VecDeque::new()))
+            outgoing.extend(self.try_coin());
+            Ok((self.output, outgoing))
```
afck: We need to make sure we output only once.

vkomenda:
Yes, I fixed the once-per-instance property of output too.

PR #13, opened 2018-05-08
Fix broadcast and example, enable more tests.

afck:
This fixes several issues with the broadcast algorithm and enables the
previously ignored tests that now pass:
* Don't decide on a root hash based on anyone's `Echo` message.
* Work around the `merkle` crate's inability to produce the proof of the
  `i`-th leaf for a given index `i`.
* Ignore messages from unknown nodes.
* Avoid decoding multiple times.
* Document the full algorithm.
* Don't count multiple `Echo` or `Ready` messages coming from the same
  node.
* Fix index computation for a given proof.
* Move `BroadcastMessage` into `broadcast` to make the module more
  self-contained.

The example now only executes a single broadcast instance, expecting the
first node (the one with the lexicographically lowest address) to
propose a value. A shell script is added that runs for example nodes.

Use env_logger instead of simple_logger, so the log level can be controlled
with an environment variable. You can e.g. log all output from the broadcast
test and the crate itself in debug mode with:
RUST_LOG=hbbft=debug,broadcast=debug

Some debugging messages are more concise now and use hexadecimal
notation instead of printing arrays of decimal values.

An indentation error in the Travis script is also fixed.

### DIFF
--- a/examples/run-consensus-nodes.sh
+++ b/examples/run-consensus-nodes.sh
@@ -3,11 +3,13 @@
 export RUST_LOG=hbbft=debug
 
 cargo build --example consensus-node
-cargo run --example consensus-node -- --bind-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5002 --remote-address=127.0.0.1:5003 --value=Foo &
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5002 --remote-address=127.0.0.1:5003 --remote-address=127.0.0.1:5004 --value=Foo &
 sleep 1
-cargo run --example consensus-node -- --bind-address=127.0.0.1:5001 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5002 --remote-address=127.0.0.1:5003 &
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5001 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5002 --remote-address=127.0.0.1:5003 --remote-address=127.0.0.1:5004 &
 sleep 1
-cargo run --example consensus-node -- --bind-address=127.0.0.1:5002 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5003 &
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5002 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5003 --remote-address=127.0.0.1:5004 &
 sleep 1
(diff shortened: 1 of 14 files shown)


afck:
That's odd: `cargo fmt` fails and says it can't find `src/proto/message.rs`, even though the file is clearly there (tests pass, etc.) and it also works locally for me. I'll disable it for now and investiage separately.

examples/run-consensus-nodes.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
```
vkomenda: I suggest calling every node with the environment parameter that enables logging. E.g.
`RUST_LOG=hbbft=debug`. Otherwise there is no output.

src/proto/mod.rs
```diff
@@ -326,3 +298,19 @@ impl LemmaProto {
         }
     }
 }
+
+/// The path of the lemma, as a binary string
+fn path_of_lemma(mut lemma: &Lemma) -> String {
```
vkomenda: I would suggest having a custom return type instead of String. Such as a vector of bool wrapped in a newtype.
```
struct BinaryPath(Vec<bool>);
```
Adding a custom Display type will achieve your goal of succinct debug output.

afck:
The example still hangs sporadically: Messages sent to "all" nodes sometimes only reach some of them. I suspect that the reason is this [issue](https://github.com/poanetwork/hbbft/blob/afck--remove-channels/src/proto_io.rs#L104), where protobuf skips a message and just sets the current position to the end of the stream. If more than one message is in the buffer, that would swallow the second one.

vkomenda:
protobuf's function `recv` doesn't seek to the next byte after the end of the received message. This is a sure bug.

vkomenda:
In fact I meant this function: https://github.com/stepancheg/rust-protobuf/blob/99ed2af856559759f7be0abca29d3fa6704c8461/protobuf/src/stream.rs#L702

examples/run-consensus-nodes.sh
```diff
@@ -0,0 +1,13 @@
...
+
+cargo build --example consensus-node
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5002 --remote-address=127.0.0.1:5003 --value=Foo &
+sleep 1
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5001 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5002 --remote-address=127.0.0.1:5003 &
+sleep 1
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5002 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5003 &
+sleep 1
+cargo run --example consensus-node -- --bind-address=127.0.0.1:5003 --remote-address=127.0.0.1:5000 --remote-address=127.0.0.1:5001 --remote-address=127.0.0.1:5002 &
+wait
```
vkomenda: Can you add a fifth node? That would help demonstrating the lemma index fix!

afck: Done.

examples/run-consensus-nodes.sh
afck: Done.

PR #14, opened 2018-05-10
Simplify the message types.

afck:
This moves the remaining channel/crossbeam logic to the examples and reorganizes the message types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,18 +5,18 @@ authors = ["Vladimir Komendantskiy <komendantsky@gmail.com>"]
 
 [dependencies]
 env_logger = "0.5.10"
+itertools = "0.7"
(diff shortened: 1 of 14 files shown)


afck:
I agree, but I'd like to try and simplify the example and maybe the message types even further before we add more functionality to it again. Since it's just an example I think readability and simplicity are more important than efficiency, and I'm hoping it could all basically just be one small loop, without any channels or multithreading:
Wait for incoming messages, deserialize them, pass them into the algorithm (possibly selecting among several instances), serialize the returned messages and send them out; repeat.

On the other hand it would be nice to have a more sophisticated example, of course, like a little chat app where Honey Badger guarantees that the messages appear in the same order on everyone's screen. But I guess we should focus on the library itself first. :smile:

PR #15, opened 2018-05-12
Start implementing the top-level Honey Badger algorithm.

afck:
This also contains a few fixes for the `common_subset` module:
* Rename `common_subset::Output` to `Message` to avoid confusing it
  with the value that the algorithm outputs as a result.
* Implement dispatch of messages to the right instance within
  `CommonSubset`, in a way that is transparent to the user.

And it generalizes `TargetedMessage`, so that it can replace e.g. `TargetedBroadcastMessage`.

### DIFF
--- a/src/common_subset.rs
+++ b/src/common_subset.rs
@@ -11,23 +11,50 @@ use agreement;
 use agreement::{Agreement, AgreementMessage};
 
 use broadcast;
-use broadcast::{Broadcast, BroadcastMessage, TargetedBroadcastMessage};
+use broadcast::{Broadcast, BroadcastMessage};
+
+use messaging::{Target, TargetedMessage};
 
 // TODO: Make this a generic argument of `Broadcast`.
 type ProposedValue = Vec<u8>;
 // Type of output from the Common Subset message handler.
-type CommonSubsetOutput<NodeUid> = (Option<HashSet<ProposedValue>>, VecDeque<Output<NodeUid>>);
+type CommonSubsetOutput<NodeUid> = (
+    Option<HashSet<ProposedValue>>,
+    VecDeque<TargetedMessage<Message<NodeUid>, NodeUid>>,
+);
 
-/// Output from Common Subset to remote nodes.
-pub enum Output<NodeUid> {
-    /// A broadcast message to be sent to the destination set in the
-    /// `TargetedBroadcastMessage`.
-    Broadcast(TargetedBroadcastMessage<NodeUid>),
-    /// An agreement message to be broadcast to all nodes. There are no
-    /// one-to-one agreement messages.
-    Agreement(AgreementMessage),
+/// Message from Common Subset to remote nodes.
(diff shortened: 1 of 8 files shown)


src/common_subset.rs
```diff
@@ -138,7 +188,9 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
         input_result.map(|mut queue| {
             if let Some(agreement_message) = opt_message {
                 // Append the message to agreement nodes to the common output queue.
-                queue.push_back(Output::Agreement(agreement_message))
+                queue.push_back(
+                    Target::All.message(Message::Agreement(proposer_id.clone(), agreement_message)),
```
vkomenda: Here we have the proposer ID back as part of the message. Is this better than adding the proposer ID at the top level? The proposer ID is associated to the instance at the top level anyway.

src/common_subset.rs
```diff
@@ -173,22 +225,19 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
-            // error
-            result
-                .map(|(_, messages)| (None, messages.into_iter().map(Output::Agreement).collect()))
+        // Process Agreement outputs.
+        if let Some(b) = output {
+            outgoing.append(&mut self.on_agreement_result(proposer_id, b)?);
         }
+
+        // Check whether Agreement has completed.
+        let into_msg = |a_msg| Target::All.message(Message::Agreement(proposer_id.clone(), a_msg));
```
vkomenda: There are no `Target::Node` Agreement messages. Annotating the message may fit into the common message delivery system on one hand, but on the other, the messaging system may take into account that any Agreement message is a multicast.

src/common_subset.rs
```diff
@@ -105,25 +133,47 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
...
+            if let Some(broadcast_instance) = self.broadcast_instances.get_mut(proposer_id) {
                 broadcast_instance
                     .handle_broadcast_message(sender_id, bmessage)
                     .map(|(opt_value, queue)| {
                         instance_result = opt_value;
-                        queue.into_iter().map(Output::Broadcast).collect()
+                        queue
+                            .into_iter()
+                            .map(|msg| {
+                                msg.map(|b_msg| Message::Broadcast(proposer_id.clone(), b_msg))
```
vkomenda: Looks a bit involved but I guess this is how adding the proposer to the message was supposed to work anyway.

afck: I'm hoping that the `TargetedMessage` and its `map` method makes the distinction explicit between the message content, which higher-level algorithms will probably wrap into their own message types, and the addressee, to whom the wrapped message needs to be sent.
But maybe we can find a better way to do that.

PR #16, opened 2018-05-14
Binary Agreement test module

vkomenda:
This pull request adds a simple test module of Binary Agreement. There is no notion of adversary at the moment. Asynchrony between BA instances is not modelled yet either.

### DIFF
--- a/tests/agreement.rs
+++ b/tests/agreement.rs
@@ -1,4 +1,7 @@
-//! Tests of the Binary Byzantine Agreement protocol.
+//! Tests of the Binary Byzantine Agreement protocol. Only one proposer instance
+//! is tested. Each of the nodes in the simulated network run only one instance
+//! of Agreement. This way we only test correctness of the protocol and not
+//! message dispatch between multiple proposers.
 //!
 //! There are three properties that are tested:
 //!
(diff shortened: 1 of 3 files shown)


tests/agreement.rs
```diff
@@ -0,0 +1,196 @@
...
+/// Message pertaining to a particular Agreement instance.
+type InstanceMessage = (NodeId, AgreementMessage);
+type MessageQueue = VecDeque<InstanceMessage>;
+
+struct TestNode {
+    /// Sender ID.
+    id: NodeId,
+    /// Map from proposer IDs into agreement instances.
+    agreements: BTreeMap<NodeId, Agreement<NodeId>>,
+    /// Queue of triples (sender_id, proposer_id, message).
```
afck: Rust code should be in backticks in comments (at least in doc comments with `///`).

afck: (Sorry for the spam; I intended to include this with the other comments in the code review.)

tests/agreement.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Message pertaining to a particular Agreement instance.
+type InstanceMessage = (NodeId, AgreementMessage);
+type MessageQueue = VecDeque<InstanceMessage>;
+
+struct TestNode {
+    /// Sender ID.
+    id: NodeId,
+    /// Map from proposer IDs into agreement instances.
+    agreements: BTreeMap<NodeId, Agreement<NodeId>>,
```
afck: I'd actually just test a single instance here: All the dispatching based on the proposer ID is implemented here in the test and not in the production code, so all this does is running several agreement tests in parallel.

In the production code, the dispatching is implemented in `common_subset`, so the tests for that module will verify that part of the code.

tests/agreement.rs
vkomenda: OK, triples should go once I revert to a single proposer test.

tests/agreement.rs
vkomenda: This makes sense. I can revert to testing a single proposer.

tests/agreement.rs
```diff
@@ -0,0 +1,169 @@
...
+    while network.nodes.values().any(|node| node.outputs.is_empty()) {
+        let (NodeId(id), output) = network.step();
+        if let Some(value) = output {
+            debug!("Node {} output {}", id, value);
+        }
+    }
+    network.nodes
+}
+
+/// Test 3 correct and 1 faulty node. The faulty node simply negates all other
```
afck: Node 3 isn't really faulty, it just starts with a different value but follows the protocol correctly.

PR #21, opened 2018-05-14
Define a common DistAlgorithm trait.

afck:
This will allow us to deduplicate network simulations in tests for the
different algorithms. More generally, it facilitates writing general
tools and applying them to all distributed algorithms.

### DIFF
--- a/src/broadcast.rs
+++ b/src/broadcast.rs
@@ -94,7 +94,7 @@ pub struct Broadcast<N> {
     num_nodes: usize,
     num_faulty_nodes: usize,
     data_shard_num: usize,
-    coding: ReedSolomon,
+    coding: Coding,
     /// Whether we have already multicast `Echo`.
     echo_sent: bool,
     /// Whether we have already multicast `Ready`.
@@ -170,7 +170,7 @@ impl<N: Eq + Debug + Clone + Ord> Broadcast<N> {
         let num_faulty_nodes = (num_nodes - 1) / 3;
         let parity_shard_num = 2 * num_faulty_nodes;
         let data_shard_num = num_nodes - parity_shard_num;
-        let coding = ReedSolomon::new(data_shard_num, parity_shard_num)?;
+        let coding = Coding::new(data_shard_num, parity_shard_num)?;
 
         Ok(Broadcast {
             our_id,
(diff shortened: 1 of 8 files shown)


src/broadcast.rs
```diff
@@ -89,19 +86,68 @@ pub struct Broadcast<NodeUid> {
...
+    /// The outgoing message queue.
+    messages: VecDeque<TargetedMessage<BroadcastMessage, N>>,
+    /// The output, if any.
+    output: Option<Vec<u8>>,
 }
 
-impl<NodeUid: Eq + Hash + Debug + Clone + Ord> Broadcast<NodeUid> {
+impl<N: Eq + Debug + Clone + Ord> DistAlgorithm for Broadcast<N> {
+    type NodeUid = N;
+    type Input = Vec<u8>; // TODO: Allow anything serializable.
```
vkomenda: Can you comment a bit more on what you mean by "serializable". Is it `Serde::Serialize`?

src/broadcast.rs
```diff
@@ -89,19 +86,68 @@ pub struct Broadcast<NodeUid> {
...
+    type Output = Self::Input;
+    type Message = BroadcastMessage;
+    type Error = Error;
+
+    fn input(&mut self, input: Self::Input) -> Result<(), Self::Error> {
+        if self.our_id != self.proposer_id {
+            return Err(Error::UnexpectedMessage);
```
vkomenda: Possibly the error message name is obsolete. Whether the argument `input` arrives as part of a message depends on the context of application of the function `input`. Can the message name reflect why the input is rejected? E.g., `InstanceCannotPropose`?

src/broadcast.rs
afck: Yes, and deserializable, too; something like I already did in honey_badger.rs: `T: serde::Serialize + serde::de::DeserializeOwned`.

The downside of this is that it depends on serde unconditionally. I think that's okay, since serde is extremely widely used anyway, but alternatively we could make this also depend on a feature flag later, and e.g. optionally use `T: protobuf::Message` instead.

src/broadcast.rs
```diff
@@ -89,19 +86,68 @@ pub struct Broadcast<NodeUid> {
     /// Whether we have already output a value.
     has_output: bool,
```
vkomenda: Is `has_output` equivalent to `output.is_some()`?

src/broadcast.rs
```diff
@@ -89,19 +86,68 @@ pub struct Broadcast<NodeUid> {
...
+    fn next_message(&mut self) -> Option<TargetedMessage<Self::Message, N>> {
+        self.messages.pop_front()
+    }
+
+    fn next_output(&mut self) -> Option<Self::Output> {
+        self.output.take()
+    }
+
+    fn terminated(&self) -> bool {
+        self.has_output
```
vkomenda: Can this line be `self.output.is_some()` instead? And `has_output` removed?

src/broadcast.rs
afck: No, because the `next_output` method empties `self.output`: It should return `Some` only once, but we need to remember whether we already output something before.

src/broadcast.rs
```diff
@@ -89,19 +86,68 @@ pub struct Broadcast<NodeUid> {
...
+            BroadcastMessage::Echo(p) => self.handle_echo(sender_id, p),
+            BroadcastMessage::Ready(ref hash) => self.handle_ready(sender_id, hash),
+        }
+    }
+
+    fn next_message(&mut self) -> Option<TargetedMessage<Self::Message, N>> {
+        self.messages.pop_front()
+    }
+
+    fn next_output(&mut self) -> Option<Self::Output> {
```
vkomenda: In what sense does this provide a "next" output given that there should be at most one output? I think of the output more in terms of generators: https://doc.rust-lang.org/1.23.0/std/ops/trait.Generator.html
I think this function is performing a `Return` action of a generator `resume`.

tests/broadcast.rs
```diff
@@ -68,37 +63,43 @@ enum MessageScheduler {
 
 impl MessageScheduler {
     /// Chooses a node to be the next one to handle a message.
-    fn pick_node(&self, nodes: &BTreeMap<NodeId, TestNode>) -> NodeId {
+    fn pick_node<D: DistAlgorithm>(&self, nodes: &BTreeMap<D::NodeUid, TestNode<D>>) -> D::NodeUid {
```
vkomenda: This is great. The scheduler looks generic enough to be made into a separate module. I wonder if that makes sense at the moment.

vkomenda: Nevermind, that would require abstracting over `TestNode` and that goes beyond the scope of this PR.

afck: Yes, let's move it to something like `tests/network` when we use it for other tests. (I didn't do it in this PR, to keep the diff readable.)

src/broadcast.rs
afck: I think it can't. (See above.)

src/broadcast.rs
afck: Done.

afck:
I added a fix: broadcast now also works with 1, 2 or 3 nodes.

afck:
I'm happy with locally renaming `has_output` to `decided` in the `broadcast` implementation. (I agree that `has_output` is ambiguous. I meant "output" as the past tense of the verb, i.e. "has already done some outputting", not as the noun "has an output available".)

Regarding the `Vec`/`Iterator`: How would you feel about having a `DistAlgorithm::output_iter` instead that would return an iterator over the currently available outputs? I already added that for messages anyway, so it would only be natural for outputs, too.

vkomenda:
`DistAlgorithm::output_iter` is a good idea. Please add it.

How about a name `made_output` or `produced_output` instead of `has_output`? Those may be less ambiguous while still not too specific not to include the multiple input case.

afck:
I also added the missing values into `Agreement::sent_bval`. (Thanks to @anthdm for spotting this!)

vkomenda:
Yes, I'd prefer `made_output` instead of `decided` to include the multiple output trait instances.

PR #22, opened 2018-05-14
Make protobuf support optional, add serde support.

### LINKED ISSUE #19: Put protobuf serialization behind a feature flag.
This would make it easier for users who don't want to use protobuf to compile the crate. It would remove the requirement to have `protoc` installed.

afck:
Protobuf support is now only built if the feature
"serialization-protobuf" is enabled.

Close #19


src/util.rs
```diff
@@ -0,0 +1,82 @@
+use merkle::proof::{Lemma, Positioned, Proof};
```
vkomenda: Can you give a more descriptive name to the `util` module?

Cargo.toml
```diff
@@ -8,17 +8,26 @@ bincode = "1.0.0"
...
+[features]
+serialization-protobuf = [ "protobuf", "protoc-rust" ]
+serialization-serde = [ "merkle/serialization-serde", "serde_derive" ]
 
 [build-dependencies]
-protoc-rust = "1.6.0"
+protoc-rust = { version = "1.6.0", optional = true }
 
 [dev-dependencies]
 crossbeam = "0.3.2"
```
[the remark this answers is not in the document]
afck: The example still requires it and I think it's not possible to define example-only (but not test) dependencies in `Cargo.toml` at the moment.

src/util.rs
afck: Good point! Maybe `fmt`, since it now contains only formatting? I renamed it, but I'm open to better suggestions.

PR #23, opened 2018-05-14
Cleaned up imports.

DrPeterVanNostrand:
- Standardized import ordering (from top of file downwards) to:
```
extern <crate name>
mod <mod name>
use:std:: ...
use::<extern crate name>:: ...
use::<mod name>:: ...
```
- Removed `*` imports.
- Removed redundant and/or unnecessary namespace (`::`) operators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,3 +49,28 @@ Once you have verified that the `protoc` binary is in your `$PATH`, you can
 build `hbbft` using cargo:
 
     $ cargo build [--release]
+
+# Contributing
+
+Before submitting PRs to `hbbft`, check that your code is properly
+formatted using `rustfmt`.
+
+You can check to see whether or not you have `rustfmt` installed by running
+the following:
+
+    $ cargo fmt --help
+
+If you see an error message similiar to the following, you will need to
+install `rustfmt`:
+
+    error: toolchain '<you rust target archetecture>' does not have the binary `cargo-fmt`
+
+To install `rustfmt` run the following:
+
+    $ rustup component add rustfmt-preview
+
+To run `rustfmt` on `hbbft` run the following:
+
+    $ cargo fmt
+
+Commit and push your changes, then submit your PR!
--- a/src/common_subset.rs
+++ b/src/common_subset.rs
@@ -319,4 +319,3 @@ impl<NodeUid: Clone + Debug + Display + Eq + Hash + Ord> CommonSubset<NodeUid> {
         }
     }
 }
-
--- a/src/common_subset.rs
+++ b/src/common_subset.rs
@@ -7,21 +7,40 @@ use std::collections::{HashMap, HashSet, VecDeque};
 use std::fmt::{Debug, Display};
 use std::hash::Hash;
 
-use agreement;
-use agreement::{Agreement, AgreementMessage};
-
-use broadcast;
-use broadcast::{Broadcast, BroadcastMessage};
-
+use agreement::{self, Agreement, AgreementMessage};
+use broadcast::{self, Broadcast, BroadcastMessage};
 use messaging::{Target, TargetedMessage};
 
 // TODO: Make this a generic argument of `Broadcast`.
 type ProposedValue = Vec<u8>;
+
 // Type of output from the Common Subset message handler.
 type CommonSubsetOutput<NodeUid> = (
     Option<HashSet<ProposedValue>>,
     VecDeque<TargetedMessage<Message<NodeUid>, NodeUid>>,
 );
+
+#[derive(Clone, Debug)]
+pub enum Error {
+    UnexpectedMessage,
+    NotImplemented,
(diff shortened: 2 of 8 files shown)


src/proto/mod.rs
```diff
@@ -1,18 +1,29 @@
...
+use broadcast::BroadcastMessage;
+use proto::message::{
+    AgreementProto,
+    BroadcastProto,
+    EchoProto,
+    LemmaProto,
+    ProofProto,
+    ReadyProto,
+    ValueProto
+};
```
afck: Is rustfmt okay with this? (We're using `rustfmt 0.4.2-nightly (dd807e2 2018-04-18)` and would like to enforce it in Travis, but last time I tried, a bug that I could not reproduce locally was preventing it.)

DrPeterVanNostrand: Oh, yea. Rustfmt complains about multi-line imports. Rustfmt will ask you to line wrap at `max_width`, so assuming max_width defaults to 100, it should look like:

```
use proto::message::{
    AgreementProto, BroadcastProto, EchoProto, LemmaProto, ProofProto, ReadyProto, ValueProto,
};
```

I can change the imports to work with rustfmt 0.4.2-nightly.

README.md
```diff
@@ -49,3 +49,28 @@ Once you have verified that the `protoc` binary is in your `$PATH`, you can
 build `hbbft` using cargo:
 
     $ cargo build [--release]
+
```
vkomenda: @DrPeterVanNostrand: Can you please add the formatting guidelines here. You have applied a coding standard to the code. Please share that standard with us.

src/common_subset.rs
```diff
@@ -7,22 +7,41 @@ use std::collections::{HashMap, HashSet, VecDeque};
...
 // TODO: Make this a generic argument of `Broadcast`.
 type ProposedValue = Vec<u8>;
+
 // Type of output from the Common Subset message handler.
 type CommonSubsetOutput<NodeUid> = (
     Option<HashSet<ProposedValue>>,
     VecDeque<TargetedMessage<Message<NodeUid>, NodeUid>>,
 );
 
+#[derive(Clone, Debug)]
```
vkomenda: Is there a rule in your coding standard for `Error` to be moved over here? You mentioned only imports and namespace operators in the PR description. Does this change make code any better?

src/honey_badger.rs
```diff
@@ -9,6 +9,40 @@ use serde::Serialize;
 use common_subset::{self, CommonSubset};
 use messaging::TargetedMessage;
 
+/// A Honey Badger error.
```
vkomenda: Likewise, I don't see why these definitions should be moved.

src/proto_io.rs
```diff
@@ -1,10 +1,11 @@
...
 
-use protobuf::{self, Message};
-use std::io::{Read, Write};
+use std::cmp;
+use std::io::{self, Read, Write};
 use std::marker::PhantomData;
 use std::net::TcpStream;
-use std::{cmp, io};
+
+use protobuf::{self, CodedOutputStream, Message, ProtobufError};
```
vkomenda: Modules may have their own `Message` types. Referring to those types as `module::Message` adds readability, in my opinion.

src/common_subset.rs
DrPeterVanNostrand: But this change is a totally subjective, and is one that I don't feel strongly about. If you have a preference one way or another,  I can change it back without a problem.

src/honey_badger.rs
DrPeterVanNostrand: See above comment for my reasoning on putting errors at the top of the file, especially in circumstances like the following:

```
use some_crate;

pub enum Error {
    SomeCrate(some_crate::Error)
}
```
I think that it is another (loosely held) convention to place types that are composed of other types closer to each other in the file as it aids readability. For example:

```
struct A;
struct B;
type MyTuple = (A, B);
```

is far more readable than something like:

```
struct A;
struct B;
... 100 lines of code ...
type MyTuple = (A, B);
```

DrPeterVanNostrand: I don't feel very strongly about this change, if you have a preference I am happy to change things around.

src/proto_io.rs
DrPeterVanNostrand: Ok, so you are saying that it would be more clear to write:

```
use protobuf::{self, CodedOutputStream, ProtobufError};
...
impl<S: Read + Write, M: protobuf::Message> ProtoIo<S, M>
...
```

Instead of:

```
use protobuf::{self, CodedOutputStream, Message, ProtobufError};
...
impl<S: Read + Write, M: Message> ProtoIo<S, M>
```

I agree with the argument that there are multiple `Message` types in the `hbbft` crate, and adding `protobuf::` before `Message` would make things clearer.

PR #30, opened 2018-05-17
The BA module now defines a DistAlgorithm instance

vkomenda:
I've updated the BA and ACS modules so that BA is now an instance of `DistAlgorithm`. I also modified one of the BA tests to test for termination when more than 2/3 nodes propose.

### DIFF
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -268,22 +268,20 @@ impl<NodeUid: Clone + Debug + Eq + Hash> Agreement<NodeUid> {
             self.uid, self.epoch
         );
 
-        let decision = if vals.len() != 1 {
+        if vals.len() != 1 {
             self.estimated = Some(coin);
-            None
         } else {
             // NOTE: `vals` has exactly one element due to `vals.len() == 1`
             let v: Vec<bool> = vals.into_iter().collect();
             let b = v[0];
             self.estimated = Some(b);
             // Outputting a value is allowed only once.
-            if self.output.is_none() && b == coin {
+            if self.decision.is_none() && b == coin {
                 // Output the agreement value.
                 self.output = Some(b);
+                // Latch the decided state.
+                self.decision = Some(b);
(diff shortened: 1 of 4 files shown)


src/agreement.rs
```diff
@@ -268,8 +289,13 @@ impl<NodeUid: Clone + Debug + Eq + Hash> Agreement<NodeUid> {
 
         let b = self.estimated.unwrap();
         self.sent_bval.insert(b);
-        let bval_msg = AgreementMessage::BVal(self.epoch, b);
-        (decision, vec![bval_msg])
+        self.messages
+            .push_back(AgreementMessage::BVal(self.epoch, b));
+        self.output = decision;
```
afck: Should it say `if self.decision.is_none()` in line 280 now?

src/common_subset.rs
```diff
@@ -12,7 +12,7 @@ use agreement::{Agreement, AgreementMessage};
 use broadcast;
 use broadcast::{Broadcast, BroadcastMessage};
 use fmt::HexBytes;
-use messaging::{DistAlgorithm, Target, TargetedMessage};
+use messaging::{DistAlgorithm, MessageIter, OutputIter, Target, TargetedMessage};
```
afck: It shouldn't be necessary to import the iterator types, since `DistAlgorithm` has convenience methods for them.

src/agreement.rs
```diff
@@ -42,12 +38,74 @@ pub struct Agreement<NodeUid> {
...
+            _ if self.terminated => Err(Error::Terminated),
+
+            AgreementMessage::BVal(epoch, b) if epoch == self.epoch => {
+                self.handle_bval(sender_id, b)
+            }
+
+            AgreementMessage::Aux(epoch, b) if epoch == self.epoch => self.handle_aux(sender_id, b),
+
+            // Epoch does not match. Ignore the message.
+            _ => Ok(()),
```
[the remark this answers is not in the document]
vkomenda: Good point! That will need a bit of design.

PR #32, opened 2018-05-17
Remove protoc dep, use new MerkleTree methods, fix Agreement.

afck:


src/agreement.rs
```diff
@@ -71,18 +84,20 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> DistAlgorithm for Agreement<NodeU
         sender_id: &Self::NodeUid,
         message: Self::Message,
     ) -> Result<(), Self::Error> {
+        if self.terminated {
+            return Err(Error::Terminated);
+        }
+        if message.epoch() < self.epoch {
+            return Ok(()); // Message is obsolete: We are already in a later epoch.
+        }
+        if message.epoch() > self.epoch {
```
vkomenda: Could be `else if`.

src/agreement.rs
```diff
@@ -178,32 +190,35 @@ impl<NodeUid: Clone + Debug + Eq + Hash> Agreement<NodeUid> {
...
 
+    fn send_bval(&mut self, b: bool) -> Result<(), Error> {
+        // Record the value `b` as sent.
+        self.sent_bval.insert(b);
+        // Multicast BVAL.
+        self.messages
+            .push_back(AgreementMessage::BVal(self.epoch, b));
+        // Receive the BVAL message locally.
+        let our_uid = self.uid.clone();
+        self.handle_bval(&our_uid, b)
```
vkomenda: Well spotted!

src/agreement.rs
```diff
@@ -178,32 +190,35 @@ impl<NodeUid: Clone + Debug + Eq + Hash> Agreement<NodeUid> {
                 self.messages
                     .push_back(AgreementMessage::Aux(self.epoch, b));
                 // Receive the AUX message locally.
-                self.received_aux.insert(self.uid.clone(), b);
+                let our_uid = self.uid.clone();
+                self.handle_aux(&our_uid, b)?;
```
vkomenda: Makes sense too.

src/agreement.rs
```diff
@@ -286,9 +302,12 @@ impl<NodeUid: Clone + Debug + Eq + Hash> Agreement<NodeUid> {
         };
 
         let b = self.estimated.unwrap();
-        self.sent_bval.insert(b);
-        self.messages
-            .push_back(AgreementMessage::BVal(self.epoch, b));
+        self.send_bval(b)?;
+        let queued_msgs = mem::replace(&mut self.incoming_queue, Vec::new());
+        for (sender_id, msg) in queued_msgs {
+            self.handle_message(&sender_id, msg)?;
```
vkomenda: Here is one possible place for introducing spam control. I've no idea though how to avoid spam apart from using a set instead of `Vec` of queued messages and filtering out messages with unknown sender IDs. The latter is already done at the top level.

afck: That's true. Basically we just need to store per epoch: who sent us which `BVal`s and which `Aux`.

We still might want to limit the number of future epochs somehow. On the other hand, that wouldn't be strictly correct…

afck: I created a new issue for this: https://github.com/poanetwork/hbbft/issues/43

PR #34, opened 2018-05-19
Extend tests and fix CommonSubset.

afck:
Verify termination and more outputs.

`CommonSubset` now instantiates `Agreement` with the correct ID.

### DIFF
--- a/src/common_subset.rs
+++ b/src/common_subset.rs
@@ -33,6 +33,28 @@ pub enum Message<NodeUid> {
     Agreement(NodeUid, AgreementMessage),
 }
 
+/// The queue of outgoing messages in a `CommonSubset` instance.
+#[derive(Deref, DerefMut)]
+struct MessageQueue<NodeUid>(VecDeque<TargetedMessage<Message<NodeUid>, NodeUid>>);
+
+impl<NodeUid: Clone + Debug + Eq + Hash + Ord> MessageQueue<NodeUid> {
+    /// Appends to the queue the messages from `agr`, wrapped with `proposer_id`.
+    fn extend_agreement(&mut self, proposer_id: &NodeUid, agr: &mut Agreement<NodeUid>) {
+        let convert = |msg: TargetedMessage<AgreementMessage, NodeUid>| {
+            msg.map(|a_msg| Message::Agreement(proposer_id.clone(), a_msg))
+        };
+        self.extend(agr.message_iter().map(convert));
+    }
+
+    /// Appends to the queue the messages from `bc`, wrapped with `proposer_id`.
+    fn extend_broadcast(&mut self, proposer_id: &NodeUid, bc: &mut Broadcast<NodeUid>) {
+        let convert = |msg: TargetedMessage<BroadcastMessage, NodeUid>| {
+            msg.map(|b_msg| Message::Broadcast(proposer_id.clone(), b_msg))
+        };
+        self.extend(bc.message_iter().map(convert));
+    }
+}
+
 /// Asynchronous Common Subset algorithm instance
 ///
 /// The Asynchronous Common Subset protocol assumes a network of `N` nodes that send signed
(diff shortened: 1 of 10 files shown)


src/common_subset.rs
```diff
@@ -132,7 +151,7 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> CommonSubset<NodeUid> {
         // Create all agreement instances.
         let mut agreement_instances: BTreeMap<NodeUid, Agreement<NodeUid>> = BTreeMap::new();
         for uid0 in all_uids {
-            agreement_instances.insert(uid0.clone(), Agreement::new(uid0.clone(), num_nodes));
+            agreement_instances.insert(uid0.clone(), Agreement::new(uid.clone(), num_nodes));
```
vkomenda: I'm interested in how the change works. `uid0` is the proposer ID ranging over all nodes. `uid` is the sender ID and is the ID of the node the ACS instance is set up to run on. If you create a new Agreement each time with the proposer ID `uid`, all the Agreement instances will be associated with this node and none with any remote nodes. Can you please give some insight of why this might not be the case?

tests/network/mod.rs
```diff
@@ -27,6 +27,12 @@ impl<D: DistAlgorithm> TestNode<D> {
         &self.outputs
     }
 
+    /// Returns whether the algorithm has terminated.
+    #[allow(unused)] // Not used in all tests.
```
vkomenda: So, your preference is to define methods as opposed to `pub` fields?

tests/common_subset.rs
```diff
@@ -31,21 +30,23 @@ fn test_common_subset<A: Adversary<CommonSubset<NodeUid>>>(
...
+    for node in network.nodes.values() {
+        if let Some(output) = expected.as_ref() {
+            assert!(iter::once(output).eq(node.outputs()));
+            continue;
         }
+        assert_eq!(1, node.outputs().len());
+        expected = Some(node.outputs()[0].clone());
+    }
+    let output = expected.unwrap();
+    assert!(output.len() * 3 > inputs.len() * 2);
```
vkomenda: Can you please make a comment on what condition this assertion is checking?

tests/broadcast.rs
```diff
@@ -82,12 +82,12 @@ fn test_broadcast<A: Adversary<Broadcast<NodeUid>>>(
     network.input(NodeUid(0), proposed_value.to_vec());
 
     // Handle messages in random order until all nodes have output the proposed value.
-    while network.nodes.values().any(|node| node.outputs().is_empty()) {
-        let id = network.step();
-        if !network.nodes[&id].outputs().is_empty() {
-            assert_eq!(vec![proposed_value.to_vec()], network.nodes[&id].outputs());
-            debug!("Node {:?} received", id);
-        }
+    while !network.nodes.values().all(TestNode::terminated) {
```
vkomenda: I agree, it's better this way.

src/common_subset.rs
afck: The first argument of `Agreement::new` is the node's own ID. The `Agreement` algorithm itself has no notion of a proposer, since every node can propose a boolean value.

The proposer ID exists only on the `CommonSubset` level and defines what the corresponding agreement instance is _about_, i.e. whose item's inclusion into the subset we are trying to agree about.

tests/network/mod.rs
afck: In general I'd tend towards yes: Encapsulate more, expose less. In this particular case I just thought it was nice to be able to write `nodes.values().all(TestNode::terminated)`. If you prefer, I'm happy to revert this.

tests/common_subset.rs
afck: If my understanding is correct, the Common Subset algorithm guarantees that the output contains more than ⅔ of the values, so we should assert that.

src/common_subset.rs
vkomenda: Right! Thanks for reminding about that. I still mix up the two roles of IDs sometimes. Possibly we should say either `proposer_id` or `sender_id` instead of just `uid`. The latter is so ambiguous.

afck: Good idea, I changed it.

PR #36, opened 2018-05-21
Update Rust, re-enable rustfmt.

afck:


afck:
To update to the same versions locally, use:

```sh
rustup default nightly-2018-05-19
rustup component add rustfmt-preview
cargo install clippy -f --vers=0.0.202
```

afck:
I think I know why CI failed: `rustfmt` expects the `message.rs` file, which is generated by `build.rs`. I moved the `rustfmt` check to after the build.

PR #44, opened 2018-05-23
Added a CONF round

vkomenda:
In order to guarantee termination, the common coin is invoked in the CONF round only for the subset of `bin_values` committed to in the CONF message. The Cobalt fix used the whole set of `bin_values`, which led to non-termination in our tests.

### DIFF
--- a/src/agreement.rs
+++ b/src/agreement.rs
@@ -24,7 +24,7 @@ error_chain!{
 /// A lattice-valued description of the state of `bin_values`, essentially the same as the set of
 /// subsets of `bool`.
 #[cfg_attr(feature = "serialization-serde", derive(Serialize))]
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum BinValues {
     None,
     False,
@@ -94,12 +94,12 @@ impl BinValues {
         }
     }
 
-    pub fn is_subset(&self, other: &BinValues) -> bool {
+    pub fn is_subset(&self, other: BinValues) -> bool {
         match self {
             BinValues::None => true,
-            BinValues::False if *other == BinValues::False || *other == BinValues::Both => true,
-            BinValues::True if *other == BinValues::True || *other == BinValues::Both => true,
-            BinValues::Both if *other == BinValues::Both => true,
+            BinValues::False if other == BinValues::False || other == BinValues::Both => true,
+            BinValues::True if other == BinValues::True || other == BinValues::Both => true,
+            BinValues::Both if other == BinValues::Both => true,
             _ => false,
         }
     }
@@ -110,6 +110,12 @@ impl BinValues {
             BinValues::True => Some(true),
             _ => None,
         }
+    }
+}
+
+impl Default for BinValues {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
@@ -308,7 +314,7 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
         // upon receiving BVAL_r(b) messages from 2f + 1 nodes,
         // bin_values_r := bin_values_r ∪ {b}
         if count_bval == 2 * self.num_faulty_nodes + 1 {
-            let previous_bin_values = self.bin_values.clone();
+            let previous_bin_values = self.bin_values;
             let _bin_values_changed = self.bin_values.insert(b);
 
             // wait until bin_values_r != 0, then multicast AUX_r(w)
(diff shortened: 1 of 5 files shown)


vkomenda:
I cannot repeat the CI error on my machine. All tests pass.

afck:
It was the broadcast test that failed with a `signal: 4, SIGILL: illegal instruction`; so it probably doesn't have anything to do with your change. First I suspected that some dependency broke, but `cargo update` didn't reproduce the problem on my machine either. Maybe let's just restart the Travis build and see whether it happens again?

src/agreement.rs
```diff
@@ -19,6 +21,116 @@ error_chain!{
     }
 }
 
+/// A lattice-valued description of the state of `bin_values`, essentially the same as the set of
+/// subsets of `bool`.
+#[cfg_attr(feature = "serialization-serde", derive(Serialize))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum BinValues {
```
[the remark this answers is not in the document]
vkomenda: I tried [bool; 2] at first but replaced it with a specialised enum because you need to define operations anyway, `insert`, `union`, and also `from_iter` for `count_conf`, and those actually take the most of code. Maybe move the enum and the methods to a separate module?

src/agreement.rs
```diff
@@ -220,9 +345,57 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
...
+        // Receive the CONF message locally.
+        let our_uid = self.uid.clone();
+        self.handle_conf(&our_uid, v)
+    }
+
+    /// Waits until at least (N − f) AUX_r messages have been received, such that
+    /// the set of values carried by these messages, vals, are a subset of
+    /// bin_values_r (note that bin_values_r may continue to change as BVAL_r
+    /// messages are received, thus this condition may be triggered upon arrival
+    /// of either an AUX_r or a BVAL_r message).
```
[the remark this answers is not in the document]
vkomenda: Good point. This paragraph was actually copied from the Honey Badger paper. I can lift the epoch in this PR. It's no biggie.

src/agreement.rs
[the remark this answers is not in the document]
afck: Yes, if we need this I'd also put it in a separate module. (Maybe a submodule of `agreement` if it's only used here.)

src/agreement/mod.rs
```diff
@@ -19,21 +23,44 @@ error_chain!{
     }
 }
 
-/// Messages sent during the binary Byzantine agreement stage.
 #[cfg_attr(feature = "serialization-serde", derive(Serialize))]
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum AgreementMessage {
+pub enum AgreementContent {
```
afck: The docs of the variants need to be updated: they still say "with an epoch".

src/agreement/mod.rs
```diff
@@ -46,13 +73,15 @@ pub struct Agreement<NodeUid> {
...
+    bin_values: BinValues,
+    /// Values received in `BVal` messages. Reset on every epoch update.
+    received_bval: BTreeMap<NodeUid, BTreeSet<bool>>,
     /// Sent BVAL values. Reset on every epoch update.
     sent_bval: BTreeSet<bool>,
-    /// Values received in AUX messages. Reset on every epoch update.
-    received_aux: HashMap<NodeUid, bool>,
+    /// Values received in `Aux` messages. Reset on every epoch update.
+    received_aux: BTreeMap<NodeUid, bool>,
+    /// Received CONF messages. Reset on every epoch update.
```
afck: There's still a "`CONF`" here, and a "`BVAL`" four lines above.

src/agreement/mod.rs
```diff
@@ -189,92 +223,152 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
...
-            .push_back(AgreementMessage::BVal(self.epoch, b));
-        // Receive the BVAL message locally.
+            .push_back(AgreementMessage::bval(self.epoch, b));
+        // Receive the `BVal` message locally.
         let our_uid = self.uid.clone();
         self.handle_bval(&our_uid, b)
     }
 
+    fn send_conf(&mut self) -> AgreementResult<()> {
+        if self.conf_round {
```
afck: Should this be called `sent_conf`, analogously to the other two `sent_*`?

src/agreement/mod.rs
```diff
@@ -189,92 +223,152 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
...
 
-    /// AUX_r messages such that the set of values carried by those messages is
-    /// a subset of bin_values_r. Outputs this subset.
+    /// The count of `Aux` messages such that the set of values carried by those messages is a
+    /// subset of bin_values_r.
     ///
-    /// FIXME: Clarify whether the values of AUX messages should be the same or
-    /// not. It is assumed in `count_aux` that they can differ.
+    /// FIXME: Clarify whether the values of `Aux` messages should be the same or not. It is assumed
+    /// in `count_aux` that they can differ.
```
afck: Can that `FIXME` be removed? The following paragraph seems to answer this question.

src/agreement/mod.rs
vkomenda: That might imply a wrong reading of `sent_conf` as the actual sent value. Instead, `conf_round` is a property whether `Agreement` is currently performing a `Conf` round.

PR #46, opened 2018-05-28
Implement threshold signatures and encryption.

### LINKED ISSUE #38: Implement a threshold signature scheme.
The common coin is based on threshold signatures: https://eprint.iacr.org/2002/118.pdf (section 4.2)

### LINKED ISSUE #40: Threshold Encryption
Honey Badger requires threshold encryption to avoid sending the proposed transactions as plaintext. http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.119.1717&rep=rep1&type=pdf

afck:
Close #38, close #40

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -1,12 +1,16 @@
 mod error;
 
 use byteorder::{BigEndian, ByteOrder};
+use init_with::InitWith;
 
 use pairing::{CurveAffine, CurveProjective, Engine, Field, PrimeField};
 use rand::{ChaChaRng, Rand, Rng, SeedableRng};
 use ring::digest;
 
 use self::error::{ErrorKind, Result};
+
+/// The number of words (`u32`) in a ChaCha RNG seed.
+const CHACHA_RNG_SEED_SIZE: usize = 8;
 
 /// Returns a hash of the given message in `G2`.
 pub fn hash_g2<E, M>(msg: M) -> E::G2
@@ -16,14 +20,10 @@ where
     M: AsRef<[u8]>,
 {
     let digest = digest::digest(&digest::SHA256, msg.as_ref());
-    // The `pairing` crate's `G2` implements `Rand`. We initialize a seedable RNG with the SHA256
-    // digest, and use it to generate the element.
-    let mut msg_u32: Vec<u32> = Vec::with_capacity(256 / 32);
-    for chunk in digest.as_ref().chunks(4) {
-        let word = BigEndian::read_u32(chunk);
-        msg_u32.push(word);
-    }
-    let mut rng = ChaChaRng::from_seed(&msg_u32);
+    let seed = <[u32; CHACHA_RNG_SEED_SIZE]>::init_with_indices(|i| {
+        BigEndian::read_u32(&digest.as_ref()[(4 * i)..(4 * i + 4)])
+    });
+    let mut rng = ChaChaRng::from_seed(&seed);
     rng.gen()
 }
 
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -1,5 +1,7 @@
 mod error;
 
+use byteorder::{BigEndian, ByteOrder};
+
 use pairing::{CurveAffine, CurveProjective, Engine, Field, PrimeField};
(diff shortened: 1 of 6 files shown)


src/crypto/mod.rs
```diff
@@ -0,0 +1,293 @@
...
+            result.add_assign(&summand);
+        }
+        Ok(Signature(result))
+    }
+}
+
+/// A secret key and an associated set of secret key shares.
+pub struct SecretKeySet<E: Engine> {
+    /// The coefficients of a polynomial whose value at `0` is the "master key", and value at
+    /// `i + 1` is key share number `i`.
```
vkomenda: I see that you refer to coefficients by their positions in the `Vec`. Wouldn't it be more descriptive to refer here to the polynomial instead of the vector that stores the coefficients? E.g., "the master key is the 0-th coefficient of the polynomial and, for 0 < i <= N, the i-th coefficient of the polynomial is the (i-1)-th key share".

afck: But the `(i - 1)`-th key share isn't the `i`-th coefficient; it's the value at `i`:

```rust
coeff[0] + i * coeff[1] + i * i * coeff[2] + …
```

Not sure what would be a better formulation.

src/crypto/mod.rs
```diff
@@ -0,0 +1,293 @@
...
+where
+    E: Engine,
+    <E as Engine>::G2: Rand,
+    M: AsRef<[u8]>,
+{
+    let digest = digest::digest(&digest::SHA512, msg.as_ref());
+    // The `pairing` crate's `G2` implements `Rand`. We initialize a seedable RNG with the SHA512
+    // digest, and use it to generate the element.
+    let mut msg_u32: Vec<u32> = Vec::with_capacity((digest.as_ref().len() + 3) / 4);
+    for chunk in digest.as_ref().chunks(4) {
```
DrPeterVanNostrand: If I'm reading this correctly, you are serializing a byte-array into a `Vec<u32>` here. The `byteorder` crate has a convenient function for doing just this, if you're not aware of it already, it would allow you simplify this code block into:

```rust
use byteorder::{BigEndian, ByteOrder};
...
for chunk in digest.as_ref().chunks(4) {
    let word = BigEndian::read_u32(chunk);
    msg_u32.push(word);
}

```

afck: Unfortunately that panics if `word.len() < 4`, so we'd have to add padding; not sure if the code would still be as concise with that?

afck: Done.

src/crypto/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+
+/// Returns a hash of the given message in `G2`.
+pub fn hash_g2<E, M>(msg: M) -> E::G2
+where
+    E: Engine,
+    <E as Engine>::G2: Rand,
+    M: AsRef<[u8]>,
+{
+    let digest = digest::digest(&digest::SHA256, msg.as_ref());
+    // The `pairing` crate's `G2` implements `Rand`. We initialize a seedable RNG with the SHA512
```
vkomenda: Now the reference to SHA512 is obsolete in the comment.

afck: Thanks! I fixed it.

src/crypto/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+pub fn hash_g2<E, M>(msg: M) -> E::G2
+where
+    E: Engine,
+    <E as Engine>::G2: Rand,
+    M: AsRef<[u8]>,
+{
+    let digest = digest::digest(&digest::SHA256, msg.as_ref());
+    // The `pairing` crate's `G2` implements `Rand`. We initialize a seedable RNG with the SHA256
+    // digest, and use it to generate the element.
+    let mut msg_u32: Vec<u32> = Vec::with_capacity(256 / 32);
```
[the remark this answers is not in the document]
afck: Good idea! I changed it.
I went a bit overboard and used [init_with](https://docs.rs/init_with/1.1.0/init_with/) to avoid the `0u32` initialization. (But the bottleneck is certainly either `digest` or `from_seed`.)

PR #49, opened 2018-05-29
Shared node ID and network size info among subalgorithms

vkomenda:


tests/honey_badger.rs
```diff
@@ -55,8 +57,13 @@ where
     // TODO: Verify that all nodes output the same epochs.
 }
 
-fn new_honey_badger(id: NodeUid, all_ids: BTreeSet<NodeUid>) -> HoneyBadger<usize, NodeUid> {
-    HoneyBadger::new(id, all_ids, 12, 0..5).expect("Instantiate honey_badger")
+// FIXME: It is possible to pass `netinfo` to `HoneyBadger::new` in the same way it is passed to
+// other algorithms. That way `new_honey_badger` would be unified with `new_agreement`, etc. However
+// a `netinfo` constructor argument would arguably make the top-level interface a bit convoluted.
```
afck: Would it?
I think it conceptually makes sense for the user to first create the `NetworkInfo` object, and then pass it into the algorithm they want to run. Especially since it will only get more complicated with the signing and encryption keys that we'll add to it. It would be nicer if changes like that only affected `NetworkInfo`, without changing the `honey_badger` module at all if it doesn't use them.

vkomenda: That would also fix the current lint warning. So, fine by me!

afck: Looks good! That note can probably be removed now.

PR #53, opened 2018-05-31
Add bandwidth and CPU to the simulation.

afck:


afck:
This (either contains a bug or) exposes a bug in `HoneyBadger`: If I run it with

```sh
cargo run --example simulation --all-features -- --tx-size 100
```

I get a bincode error:

```rust
Bincode(Io(Custom { kind: UnexpectedEof, error: StringError("failed to fill whole buffer") }))
```

I'll look into that in the afternoon.

afck:
Sorry, ignore the previous message! I just forgot to rebase. https://github.com/poanetwork/hbbft/pull/52 actually fixed this.

PR #56, opened 2018-06-07
Added Term messages in Agreement and enabled early termination

### LINKED ISSUE #55: Add a `Terminate` message to ABA?
I think our current `Agreement` implementation is not guaranteed to terminate: It _will_ eventually output, but after that it could hang.

Let's say the `f` faulty nodes just never send any messages, one node, Alice, decides and outputs :zero: in round `1`, and the coin shows :zero: in round `2` again. Then Alice will _terminate_ in round `2` and honest Bob will _output_ in round `2`. But Bob must continue until the coin comes up :zero: again. He will be stuck in round `3`, where he won't receive more than `N - f - 1` `Aux` messages.

Am I missing something? But even if I am, it would be nice to be able to terminate immediately when we output.

We should add a new message type `AgreementContent::Terminate(bool)`, which counts as `Aux`, `BVal` and `Conf` for all following rounds. Also, if a node receives `f + 1` of them, it can immediately terminate, too.

Mostéfaoui, Moumen and Raynal wrote [a follow-up paper](https://hal.archives-ouvertes.fr/hal-01176110/document) where they deal with termination. There are some other modifications to ABA in there which we should also have a closer look at. Also: How do the [Python](https://github.com/amiller/HoneyBadgerBFT/issues/63), [Erlang](https://github.com/helium/erlang-hbbft) and [Go](https://github.com/anthdm/hbbft) implementations handle the above? If this is indeed a problem, we should file bugs there, too.

vkomenda:
Close #55 

A terminated instance continues handling messages after termination according to the previous semantics of termination. However a `Term` message is sent only once.

### DIFF
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -352,18 +352,21 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
     /// f agreeing messages would not always terminate. We can, however, expect every good node to
     /// send an `Aux` value that will eventually end up in our `bin_values`.
     fn count_aux(&self) -> usize {
-        let count_aux = self
+        let mut aux: BTreeMap<_, _> = self
             .received_aux
-            .values()
-            .filter(|&&b| self.bin_values.contains(b))
-            .count();
-        let count_term = self
-            .received_term
-            .values()
-            .filter(|&&b| self.bin_values.contains(b))
-            .count();
+            .iter()
+            .filter(|(_, &b)| self.bin_values.contains(b))
+            .collect();
 
-        count_aux + count_term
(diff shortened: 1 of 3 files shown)


src/agreement/mod.rs
```diff
@@ -32,6 +32,18 @@ pub enum AgreementContent {
...
+}
+
+impl AgreementContent {
+    /// Creates an message with a given epoch number.
+    pub fn with_epoch(self, epoch: u32) -> AgreementMessage {
+        AgreementMessage {
+            epoch,
+            content: self,
+        }
+    }
```
afck: Good idea, that's much more concise! :+1:

afck:
Looks like Travis failed to download protoc. Probably some connection issue; retrying…

PR #59, opened 2018-06-13
Common coin

### LINKED ISSUE #39: Common Coin
Honey Badger's common coin implementation requires threshold signatures: https://github.com/poanetwork/hbbft/issues/38

vkomenda:
Close #39 

This might also require a further commit to solve a few merge conflicts after we merge #58.

### DIFF
--- a/examples/network/node.rs
+++ b/examples/network/node.rs
@@ -35,17 +35,15 @@
 //! consensus node will be the same `result`.
 
 use crossbeam;
-use std::collections::HashSet;
+use std::collections::{BTreeSet, HashSet};
 use std::fmt::Debug;
 use std::marker::{Send, Sync};
 use std::net::SocketAddr;
 use std::rc::Rc;
 use std::{io, iter, process, thread, time};
 
-use pairing::bls12_381::Bls12;
-use rand;
-
 use hbbft::broadcast::{Broadcast, BroadcastMessage};
+use hbbft::crypto::keygen::Poly;
 use hbbft::crypto::SecretKeySet;
 use hbbft::messaging::{DistAlgorithm, NetworkInfo, SourcedMessage};
 use hbbft::proto::message::BroadcastProto;
@@ -101,17 +99,25 @@ impl<T: Clone + Debug + AsRef<[u8]> + PartialEq + Send + Sync + From<Vec<u8>> + 
             .chain(connections.iter().map(|c| c.node_str.clone()))
             .collect();
         node_strs.sort();
-        debug!("Nodes:  {:?}", node_strs);
-        let proposer_id = 0;
         let our_id = node_strs.binary_search(&our_str).unwrap();
-        let num_nodes = connections.len() + 1;
+        let all_ids: BTreeSet<_> = (0..node_strs.len()).collect();
 
-        if value.is_some() != (our_id == proposer_id) {
+        // FIXME: This example doesn't call algorithms that use cryptography. However the keys are
+        // required by the interface to all algorithms in Honey Badger. Therefore we set placeholder
+        // keys here. A fully-featured application would need to take appropriately initialized keys
+        // from elsewhere.
+        let secret_key_set = SecretKeySet::from(Poly::zero());
+        let secret_key = secret_key_set.secret_key_share(our_id as u64);
+        let public_key_set = secret_key_set.public_keys();
+
+        let netinfo = NetworkInfo::new(our_id, all_ids.clone(), secret_key, public_key_set);
+
+        if value.is_some() != (our_id == 0) {
             panic!("Exactly the first node must propose a value.");
         }
 
         // Initialise the message delivery system and obtain TX and RX handles.
-        let messaging: Messaging<BroadcastMessage> = Messaging::new(num_nodes);
+        let messaging: Messaging<BroadcastMessage> = Messaging::new(all_ids.len());
         let rxs_to_comms = messaging.rxs_to_comms();
         let tx_from_comms = messaging.tx_from_comms();
         let rx_to_algo = messaging.rx_to_algo();
@@ -128,18 +134,8 @@ impl<T: Clone + Debug + AsRef<[u8]> + PartialEq + Send + Sync + From<Vec<u8>> + 
             // corresponding to this instance, and no dedicated comms task. The
             // node index is 0.
             let broadcast_handle = scope.spawn(move || {
-                let all_ids = (0..num_nodes).collect();
-                let sk_set =
-                    SecretKeySet::<Bls12>::new((num_nodes - 1) / 3, &mut rand::thread_rng());
-                let pk_set = sk_set.public_keys();
-                let netinfo = Rc::new(NetworkInfo::new(
-                    our_id,
-                    all_ids,
-                    sk_set.secret_key_share(our_id as u64),
-                    pk_set.clone(),
-                ));
                 let mut broadcast =
-                    Broadcast::new(netinfo, proposer_id).expect("failed to instantiate broadcast");
+                    Broadcast::new(Rc::new(netinfo), 0).expect("failed to instantiate broadcast");
 
                 if let Some(v) = value {
                     broadcast.input(v.clone().into()).expect("propose value");
@@ -210,13 +206,6 @@ impl<T: Clone + Debug + AsRef<[u8]> + PartialEq + Send + Sync + From<Vec<u8>> + 
                 .unwrap();
 
             process::exit(0);
-
-            // TODO: Exit cleanly.
-            // match msg_handle.join() {
-            //     Ok(()) => debug!("Messaging stopped OK"),
-            //     Err(e) => debug!("Messaging error: {:?}", e),
-            // }
-            // Err(Error::NotImplemented)
         }) // end of thread scope
     }
 }
--- a/examples/network/node.rs
+++ b/examples/network/node.rs
@@ -42,7 +42,11 @@ use std::net::SocketAddr;
 use std::rc::Rc;
 use std::{io, iter, process, thread, time};
 
+use pairing::bls12_381::Bls12;
+use rand;
+
 use hbbft::broadcast::{Broadcast, BroadcastMessage};
+use hbbft::crypto::SecretKeySet;
 use hbbft::messaging::{DistAlgorithm, NetworkInfo, SourcedMessage};
 use hbbft::proto::message::BroadcastProto;
 use network::commst;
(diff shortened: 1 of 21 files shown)


src/common_coin.rs
```diff
@@ -0,0 +1,217 @@
...
+            let received_shares = &self.received_shares;
+            if received_shares.len() > self.netinfo.num_faulty() {
+                // Pass the indices of sender nodes to `combine_signatures`.
+                let shares: BTreeMap<&u64, &Signature<Bls12>> = self
+                    .netinfo
+                    .all_uids()
+                    .iter()
+                    .map(|id| (&node_indices[id], received_shares.get(id)))
+                    .filter(|(_, share)| share.is_some())
+                    .map(|(n, share)| (n, share.unwrap()))
```
afck: You can combine these two lines:

```rust
.filter_map(|(n, opt_share)| opt_share.map(|share| (n, share)))
```

Or even all three:

```rust
.filter_map(|id| received_shares.get(id).map(|share| (&node_indices[id], share)))
```

src/agreement/mod.rs
```diff
@@ -387,53 +459,36 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
...
-            return Ok(());
-        }
+#[derive(Clone)]
+struct Nonce(Vec<u8>);
 
-        self.start_next_epoch();
+impl Nonce {
+    pub fn new(invocation_id: &[u8], hb_epoch: u64, agreement_epoch: u32) -> Self {
+        Nonce(Vec::from(format!(
+            "Nonce for Honey Badger {:?} @ epoch {}:{}",
```
afck: It would be slightly more efficient to just use `"{:?}@{}:{}"` here (and then the proposer ID, too, of course).

src/common_subset.rs
```diff
@@ -148,7 +148,7 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> DistAlgorithm for CommonSubset<No
 }
 
 impl<NodeUid: Clone + Debug + Eq + Hash + Ord> CommonSubset<NodeUid> {
-    pub fn new(netinfo: Rc<NetworkInfo<NodeUid>>) -> CommonSubsetResult<Self> {
+    pub fn new(netinfo: Rc<NetworkInfo<NodeUid>>, hb_epoch: u64) -> CommonSubsetResult<Self> {
```
afck: Maybe also `session_id: u64`?

src/crypto/mod.rs
```diff
@@ -52,18 +55,42 @@ impl<E: Engine> PublicKey<E> {
...
+
 impl<E: Engine> PartialEq for Signature<E> {
     fn eq(&self, other: &Signature<E>) -> bool {
         self.0 == other.0
     }
 }
 
+impl<E: Engine> Signature<E> {
+    pub fn parity(&self) -> bool {
+        let uncomp = self.0.into_affine().into_uncompressed();
```
afck: I wonder whether that's unique, i.e. does `a == b` imply `a.into_affine().into_uncompressed() == b.into_affine().into_uncompressed()`? (I assume it does, but I'm not sure.) Even if that holds true for `Bls12` but is not a guarantee of `Engine` in general, I'd maybe make this method `Bls12`-only.

src/messaging.rs
```diff
@@ -167,4 +181,28 @@ impl<NodeUid: Ord> NetworkInfo<NodeUid> {
...
+        &self.secret_key
+    }
+
+    pub fn public_key_set(&self) -> &PublicKeySet<Bls12> {
+        &self.public_key_set
+    }
+
+    /// The canonical numbering of all nodes.
+    ///
+    /// FIXME: To avoid multiple computations of the same result, caching should be introduced.
```
afck: If we do indeed need this at all, I'd vote for computing it only once, in the constructor.

src/messaging.rs
```diff
@@ -167,4 +181,28 @@ impl<NodeUid: Ord> NetworkInfo<NodeUid> {
...
+            .iter()
+            .enumerate()
+            .map(|(n, id)| (id, n as u64))
+            .collect()
+    }
+
+    /// Returns the unique ID of the Honey Badger invocation.
+    pub fn invocation_id(&self) -> Vec<u8> {
+        self.public_key_set.public_key().to_bytes()
+    }
```
afck: Maybe we should have another separate invocation ID; otherwise we make it unsafe to start two instances with the same key set.

tests/common_coin.rs
```diff
@@ -0,0 +1,130 @@
...
+    let num_samples_f64 = num_samples as f64;
+    let gain = num_samples_f64.log2().min(max_gain);
+    let step = EXPECTED_SHARE / max_gain;
+    let min_throws = (num_samples_f64 * gain * step) as usize;
+    info!(
+        "Expecting a minimum of {} throws for each coin side. Throws of true: {}. Throws of false: {}.",
+        min_throws, count_true, count_false
+    );
+    assert!(count_true > min_throws);
+    assert!(count_false > min_throws);
```
afck: What's the probability of this legitimately failing? If that's not negligible, maybe we should either `#[ignore]` this test by default and in CI, or use a fixed random seed.

Also, agreement doesn't depend on the coin being totally fair. If it's a 40%-60% coin, it would probably still work (almost) just as well. (But I agree that it makes sense to have this test in general. Anything other than 50%-50% would be worth investigating.)

examples/network/node.rs
```diff
@@ -124,7 +128,16 @@ impl<T: Clone + Debug + AsRef<[u8]> + PartialEq + Send + Sync + From<Vec<u8>> +
             // corresponding to this instance, and no dedicated comms task. The
             // node index is 0.
             let broadcast_handle = scope.spawn(move || {
-                let netinfo = Rc::new(NetworkInfo::new(our_id, (0..num_nodes).collect()));
+                let all_ids = (0..num_nodes).collect();
+                let sk_set =
+                    SecretKeySet::<Bls12>::new((num_nodes - 1) / 3, &mut rand::thread_rng());
```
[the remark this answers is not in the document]
vkomenda: Or I could deserialize the keys from a file in this example.

src/agreement/mod.rs
```diff
@@ -305,6 +328,53 @@ impl<NodeUid: Clone + Debug + Eq + Hash + Ord> Agreement<NodeUid> {
...
+    /// Handles a Common Coin message. If there is output from Common Coin, starts the next
+    /// epoch. The function may output a decision value.
+    fn handle_coin(&mut self, sender_id: &NodeUid, msg: CommonCoinMessage) -> AgreementResult<()> {
+        self.common_coin.handle_message(sender_id, msg)?;
+        self.extend_common_coin();
+
+        if let Some(coin) = self.common_coin.next_output() {
+            // Check the termination condition: "continue looping until both a value b is output in some
+            // round r, and the value Coin_r' = b for some round r' > r."
+            self.terminated = self.terminated || self.decision == Some(coin);
```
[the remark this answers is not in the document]
vkomenda: You are right. It's dead code now.

examples/network/node.rs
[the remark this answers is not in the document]
vkomenda: There was some work involved in creating an `Rc<NetworkInfo<_>>` outside the `spawn` context and I didn't want to make `SecretKey` clonable to overcome the issue. After several attempts I just made a FIXME note like you asked in the first place.

src/crypto/mod.rs
afck: No, I'm worried about [`E::G2`'s `PartialEq` implementation](https://github.com/ebfull/pairing/blob/master/src/bls12_381/ec.rs#L45) itself.
I'm wondering whether each node will compute the same value of `uncomp` here: If there are different valid in-memory representations for the same point on the affine curve, it could happen that `a == b`, even though they are not bitwise equal. ([Projective coordinates](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography#Projective_coordinates) are not unique.) The question is whether the same point can even have two different valid compressed representations.
But I think it can't, so we should be fine: The `pairing` documentation doesn't explain this but it looks like it is just [the x coordinate plus a few flags](https://github.com/ebfull/pairing/blob/master/src/bls12_381/ec.rs#L1510), which should all be unique.

src/messaging.rs
vkomenda: I'm trying to find out if the following modification of `NetworkInfo` is workable:
```
pub struct NetworkInfo<'a, NodeUid: 'a> {
    our_uid: &'a NodeUid,
    all_uids: BTreeSet<&'a NodeUid>,
    num_nodes: usize,
    num_faulty: usize,
    secret_key: SecretKey<Bls12>,
    public_key_set: PublicKeySet<Bls12>,
    /// An ordered list of node IDs for indexing purposes.
    node_list: Vec<&'a NodeUid>,
}
```
Here, `node_list` is computed in the constructor. After a straightforward refactoring, the main library compiles. However tests and examples would require some restructuring because many functions rely on passing node IDs by value. Pass-by-reference definitely works in the library but becomes quite complicated outside it.

afck: I think it's okay to require `NodeUid: Clone` everywhere, to remove the lifetime parameter.

But a `Vec<NodeUid>` still isn't optimal for index lookup (linear time). A `HashMap<NodeUid, usize>` would find the index in constant time.

tests/common_coin.rs
vkomenda: These tests with 200 and 50 samples require 42.5% and 31.4% expectations respectively. I set the maximum expectation to 48% if someone tests with 400 samples or more. I agree that this still might fail sometimes and we don't want that. I can tune down the maximum expectation to 40-45% which should be more relaxed.

src/crypto/mod.rs
vkomenda: The wikipedia article call this system Jacobian as opposed to projective.

vkomenda: This `PartialEq` instance is only required for a few assertion tests at the moment where coincidence of two points via their projective coordinates would be highly unlikely. I tried to introduce conditional compilation with `#[cfg(debug_assertions)] but that made the instance not compile in test `--release`  configuration where the debug assertions are used.
An alternative could be making the private field of `Signature` accessible from outside the module by a method or just by making the field public, and then defining the equality relation in the tests, where it does not interfere with the library.

afck: My concern was not the `PartialEq`, but whether the coin is actually common: If the uncompressed representations that two nodes compute (even though they refer to the same point of the curve) have different bytes, then they might get different coin values.
I'm pretty sure that can't happen, and they _are_ guaranteed to be equal, but the `pairing` docs don't say that explicitly.

afck: Actually, it looks like for `G2Affine` (and `Fq2`, `Fq` and `FqRepr`), `PartialEq` is derived, so they do seem to have unique representations. This should be fine then. :+1: 
(I think `x.c0 + x.c1 * u` _is_ unique, because `u` is not in the base field `Fq`.)

vkomenda: Glad we looked into it! If so, we are done with this PR?

PR #60, opened 2018-06-14
Add some crate-level documentation.

afck:


src/lib.rs
```diff
@@ -1,7 +1,87 @@
...
+//! nodes.
+//!
+//! [**Common Subset**](common_subset/index.html)
+//!
+//! Each node inputs one item. The output is a set of at least `N - f` nodes' items and will be the
+//! same in every node.
+//!
+//! This is the main building block of Honey Badger: In each epoch, every node proposes a number of
+//! transactions. Using the Common Subset protocol, they agree on at least `N - f` of those
+//! proposals. The batch contains the union of these sets of transactions.
```
vkomenda: In the code, the definition of output of `CommonSubset` is slightly more technical but I agree that the description of  output as something that _contains_ the union of sets of transactions is totally valid.

src/lib.rs
```diff
@@ -1,7 +1,87 @@
...
+//!
+//! [**Reliable Broadcast**](broadcast/index.html)
+//!
+//! One node, the _proposer_, inputs an item, and every node receives that item as an output. Even
+//! if the proposer is faulty it is guaranteed that either all nodes output the same item or none
```
vkomenda: "all nodes" -> "all correct nodes"

src/lib.rs
```diff
@@ -1,7 +1,87 @@
...
+//! at all.
```
vkomenda: The disjunction is rather more tautological: either all correct nodes output the item, or no correct nodes output it. I think that "none at all" is not correct here because faulty nodes may do whatever they like!

src/lib.rs
```diff
@@ -1,7 +1,87 @@
...
+//!
+//! This is used in Common Subset to send each node's proposal to the other nodes.
+//!
+//! [**Binary Agreement**](agreement/index.html)
+//!
+//! Each node inputs a binary value: `true` or `false`. As output, either all nodes receive `true`
```
vkomenda: "all nodes" -> "all honest nodes"

src/lib.rs
afck: You're right, strictly speaking we're not outputting a set of proposals, but a set of pairs `(proposer, proposal)`…

Maybe expanding the text (the paragraph before this one) as follows would be more accurate; or would that be unnecessarily complicated?
"The output is a set of at least `N - f` nodes' IDs, together with their proposed items, and will be the same in every node."

src/lib.rs
afck: I'll add it.

That's probably another thing we should choose consistent wording for in general: Shall we try to use "correct" everywhere, and avoid "honest"?

vkomenda: On one hand, "honest" sounds less formal. On the other, a node may be honest yet stalled... This is something that is usually covered by eventual delivery assumption in papers. So even though "correct" sounds more formal, I think it's more appropriate, provided that it is explained as well.

src/lib.rs
vkomenda: The current phrasing conveys the meaning of the output mapping very well and relates it with the paper. The mapping from IDs to proposals is just our way to keep some useful identification for test purposes that would have otherwise been lost.

src/lib.rs
```diff
@@ -1,7 +1,87 @@
...
+//! databases and blockchains.
+//!
+//! Honey Badger is modular, and composed of several other algorithms that can also be used
+//! independently. All of them are Byzantine fault tolerant and asynchronous.
+//!
+//! ## Algorithms
+//!
+//! [**Honey Badger BFT**](honey_badger/index.html)
+//!
+//! The nodes input any number of _transactions_ (any user-defined type) and outputs a sequence of
```
andogro: output

src/lib.rs
```diff
@@ -1,7 +1,87 @@
-//! # HoneyBadgerBFT
+//! # Honey Badger BFT
 //!
-//! Library of asynchronous Byzantine fault tolerant consensus known as "the
-//! honey badger of BFT protocols" after a paper with the same title.
+//! An implementation of [The Honey Badger of BFT Protocols](https://eprint.iacr.org/2016/199.pdf),
+//! an asynchronous, Byzantine fault tolerant consensus algorithm.
+//!
+//! This crate only implements the protocol itself and is meant to be instantiated in a number of
```
andogro: I have a number of recommendations for rearranging and consolidating the text - I will send an email and we can decide the best way to proceed with these types of changes. I don't want to clutter with a bunch of comments, but that may be the preferred method.

src/lib.rs
```diff
@@ -1,7 +1,87 @@
...
+//!
+//! This is used in Common Subset to send each node's proposal to the other nodes.
+//!
+//! [**Binary Agreement**](agreement/index.html)
+//!
+//! Each node inputs a binary value: `true` or `false`. As output, either all nodes receive `true`
+//! or all nodes receive `false`. The output is guaranteed to be a value that was input by at least
+//! one _honest_ node.
+//!
+//! This is used in Common Subset to decide for each node's proposal whether it should be included
```
andogro: Tighten wording - "This is used in Subset to decide whether each node's proposal should be included in the subset or not."

afck:
Thanks! I changed the text according to your suggestions.

PR #61, opened 2018-06-15
Zero out `SecretKey` memory when dropped

DrPeterVanNostrand:
- Added the `Drop` trait to `SecretKey`.

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -8,6 +8,8 @@ mod serde_impl;
 use self::keygen::{Commitment, Poly};
 
 use std::fmt;
+use std::mem;
+use std::ptr;
 
 use byteorder::{BigEndian, ByteOrder};
 use init_with::InitWith;
@@ -105,6 +107,17 @@ pub struct SecretKey<E: Engine>(E::Fr);
 impl<E: Engine> PartialEq for SecretKey<E> {
     fn eq(&self, other: &SecretKey<E>) -> bool {
         self.0 == other.0
+    }
+}
+
+impl<E: Engine> Drop for SecretKey<E> {
+    fn drop(&mut self) {
+        let size = mem::size_of_val(self);
+        unsafe {
+            let p = self as *mut Self;
+            ptr::write_bytes(p as *mut u8, 0, size);
+            ptr::write(p, SecretKey(E::Fr::zero()));
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


DrPeterVanNostrand:
Yup, no problem.

src/crypto/mod.rs
```diff
@@ -108,6 +110,17 @@ impl<E: Engine> PartialEq for SecretKey<E> {
     }
 }
 
+impl<E: Engine> Drop for SecretKey<E> {
+    fn drop(&mut self) {
+        let size = mem::size_of_val(self);
+        unsafe {
+            let p = self as *mut Self;
+            ptr::write_bytes(p as *mut u8, 0, size);
+            ptr::write(p, SecretKey(E::Fr::zero()));
```
vkomenda: In the `Bls12` implementation this is indeed represented by all zeros:
https://github.com/ebfull/pairing/blob/09b6e6f9212020f385218e5cf5287e381ccd312b/src/bls12_381/fr.rs#L327
What if it is implemented differently for another curve? Maybe the `Drop` instance should be specified only for `SecretKey<Bls12>` since this is the only one where this instuction is guaranteed to work as we expect?

Or maybe the `write` instruction is not needed after all?

DrPeterVanNostrand: The call to `write` was meant to re-instantiate a valid SecretKey object, I used `zero()` just because it was the easiest way to create a new valid object. The `zero()` call wasn't meant to zero-out memory (that's what `write_bytes()` is doing).

I thought that re-instantiating was necessary because after `SecretKey::drop()` is called,  Rust will call `std::mem::drop()` behind the scenes, which takes ownership of an argument `T` (in our case `T` is a `SecretKey`), and then frees the memory of `T`.

But, I just tried the above tests with `write()` commented out, and everything builds and passes. So, I think that you are correct, in that the call to `write()` is unnecessary.

afck: It's not really important that it's all zeros, is it? As long as every byte is overwritten with anything, we know that we can't leak a private key.

But it sounds like [clear_on_drop](https://crates.io/crates/clear_on_drop) would be the safest choice.

yrashk: My understanding is that what is also important with regards to preventing the leakage or secret keys is avoiding moving variables. Every time a variable is moved it's copied. Hopefully the upcoming Pin/Unpin thing will be of use for this case https://github.com/rust-lang/rfcs/blob/master/text/2349-pin.md

PR #62, opened 2018-06-17
Overwrite `SecretKey` memory on drop

DrPeterVanNostrand:
- Added the dependency `clear_on_drop`, which ensures that the compiler doesn't elide `SecretKey`'s destructor.
- Added ` impl Default` for `SecretKey`.
    * `Default` is a trait bound for `ClearOnDrop` type.
- Added `impl Drop` to `NetworkInfo` to overwrite `SecretKey` memory when going out of scope (uses `ClearOnDrop` type).


DrPeterVanNostrand:
If you don't like this method of clearing the secret-keys; define the `Drop` trait for each type that has a `SecretKey` as one of the types for its fields; I think something like the following would work too. We could define a type `SecretKeyWithDropSafety` (which wraps a heap allocated `SecretKey`), and we switch each structure that requires a `SecretKey` to use a `SecretKeyWithDropSafety`. This would exclude the need for defining `Drop` traits.

Both methods (i think...) remove any copying/cloning/moving of secrets and overcome any compiler optimizations that may keep keys around in memory.

In `crypto/mod.rs`:

```rust
use std::ops::{Deref, DerefMut};

pub struct SecretKeyWithDropSafety<E: Engine>(ClearWithDrop<Box<SecretKey<E>>>);

impl<E: Engine> SecretKeyWithDropSafety<E> {
    pub fn new<R: Rng>(rng: &mut R) -> Self {
        ClearWithDrop::new(Box::new(SecretKey::new(rng)))
    }
}

impl<E: Engine> Deref for SecretKeyWithDropSafety<E> {
    type Target = ClearWithDrop<Box<SecretKey<E>>>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl<E: Engine> DerefMut for SecretKeyWithDropSafety<E> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}
```

In `messaging.rs`:

```rust
#[derive(Debug)]
pub struct NetworkInfo<NodeUid: Clone + Eq + Hash> {
    ...
    secret_key: SecretKeyWithDropSafety<Bls12>,
    ...
}
```

afck:
I'd prefer not having a secret key without drop safety at all. Why not just implement it for `SecretKey` itself instead of `NetworkInfo`?
Also, is it possible just call [clear](https://docs.rs/clear_on_drop/0.2.3/clear_on_drop/clear/trait.Clear.html#tymethod.clear) instead of creating a new `ClearOnDrop` instance?

```rust
use clear_on_drop::clear::Clear;
impl<E: Engine> Drop for SecretKey<E> {
    fn drop(&mut self) {
        self.clear();
    }
}
```

DrPeterVanNostrand:
This was the first thing that I tried, but it leads to hitting the `max_recursion_limit`. The `.clear()` method calls its instance's `.drop()` method (`ptr::drop_in_place(ptr)`), so you would do endless recursive calls to `.drop()` if you manually call `.clear()` from a destructor. 

```rust
fn clear(&mut self) {
        let size = mem::size_of_val(self);
        unsafe {
            let ptr = self as *mut Self;
            ptr::drop_in_place(ptr);
            ptr::write_bytes(ptr as *mut u8, 0, size);
            hide_mem_impl::<Self>(ptr);
            Self::initialize(ptr);
        }
}
```

This is the main logic for all the clear-on-drop functionality. What happens is:

1. We get the size in bytes of the value that we have a mutable reference to (ie. a `SecretKey`)
2. Get a raw pointer to the `SecretKey`
3. Call `SecretKey.drop()`
4. Overwrite the `SecretKey` memory with zeros
5. "Hide" this location from the compiler using some inline asm magic
6. Re-instantiate a `SecretKey` at the address of the raw pointer using `SecretKey::default()`

DrPeterVanNostrand:
This also raises the question, why can't we do something like this:

```rust
pub struct SecretKey<E: Engine>(ClearOnDrop<Box<E::Fr>>);
```

We can't do this because `E::Fr` does not implement `Default`, which is a bound for the `Clear` trait.

vkomenda:
Unless the `SecretKey` is used after `clear`ing - which is the case in `drop` - I think that you don't need to initialize it again at after clearing to a default value. In the code for `clear` above, `Self::initialize(ptr)` makes sense if you continue using the object after the call to `clear`. `drop` should only overwrite the memory area of a `SecretKey` with zeros. Tell me if I'm missing something.

afck:
Ah, I see! And it works for [`ClearOnDrop` itself](https://github.com/cesarb/clear_on_drop/blob/master/src/clear_on_drop.rs#L144) because it calls `P::Target`s `clear`, and not its own…

But maybe we can add `Default` to `pairing`? https://github.com/ebfull/pairing/issues/88

DrPeterVanNostrand:
@vkomenda Yup, I agree. We don't need the re-initialization after zeroing out the memory. This is partly why I created a custom `Drop` implementation for `SecretKey` prior to switching to the `clear_on_drop` crate. I believe that the reason why the `.clear()` method from the `clear_on_drop` crate re-initializes after zeroing is to allow for a scenario such as the following, where you would want to clear/zero-out the state of a structure which contains secrets:

```rust
let mut rng = MyRng::init_with_random_seed();

{
    let clear = ClearOnDrop::new(&mut rng);
    write_to_my_secret_file(clear.gen());
}

assert_eq!(rng.state, ZERO_RNG_STATE);
rng.update_seed_with_random_value();

{
    let clear = ClearOnDrop::new(&mut rng);
    write_to_my_secret_file(clear.gen());
}
```

Sorry, if that reads like gibberish.

@afck Thank you for putting that issue in. I considered this as well. It seems like it would solve all the issues we have. The question is, should a cryptographic library implement defaults for such a specific use case as `clear_on_drop`?

afck:
> The question is, should a cryptographic library implement defaults for such a specific use case

I think clearing on drop is something that's _especially_ relevant to cryptography.
It's kind of an implementation detail that it has to do with `Default` in this case, but that, too, feels like a natural trait for an algebraic `Field`.

DrPeterVanNostrand:
Yup, that's good enough justification for me. I agree with your `pairing` issue, that numeric types in Rust default to zero, therefore it is reasonable that Fields do too.

DrPeterVanNostrand:
I guess I'd prefer to wait a day until we hear back from the `pairing` authors, a quick look at the other PRs show a response time of <= 1 day. If we don't hear back anything, we can merge with the caveat that this may be changed in the future.

afck:
No response so far. I'm fine with merging this. But one more idea:

Why not just put a `ClearOnDrop<SecretKey<_>>` into `NetworkInfo`? If even the constructor expects that instead of a `SecretKey<_>`, it would also remind the user to ideally never store a secret key outside a `ClearOnDrop`.

DrPeterVanNostrand:
Yup that's possible, but we'd have to change `NetworkInfo` to use either:

1. `ClearOnDrop<Box<SecretKey>>`
2. `ClearOnDrop<&mut SecretKey>`

`SecretKey` doesn't implement `DerefMut<Target = SecretKey>` which is required by `ClearOnDrop<SecretKey>`.

In case (1), we'd have to be explicit with user's that they have to call
`ClearOnDrop::new(Box::new( ... ))`
on their secret keys before passing them to `NetworkInfo::new()`.

In case (2), we would be introducing a lifetime into the `NetworkInfo` struct.

vkomenda:
@DrPeterVanNostrand, case (1) looks fine. Let @afck correct me but I think this was the whole idea that the user is reminded to use `ClearOnDrop` in their code.

DrPeterVanNostrand:
Yea, I think that using `Box` is the best option too.

afck:
`Looks<A<Bit<Weird>>>`, but I also like the `Box` option (1) better, yes.

DrPeterVanNostrand:
Ok, I'll make some changes to this PR.

Issue #51, opened 2018-05-30
Report faulty nodes.


afck:
Keep a log of events that prove that a particular node is faulty, and propagate that information up, e.g. via a `DistAlgorithm::get_fault_log` method. (See #24.)

vkomenda:
Errors can be logged at least in `HoneyBadger` in those places where there are TODOs to start logging something. So far there are 3 such TODOs:

1) an incorrect message from a known sender:
https://github.com/poanetwork/hbbft/blob/e0a85a54bbaf968107e4784ba2f8ca2c9e012fc4/src/honey_badger.rs#L238

2) an incorrect ciphertext computed for a known proposer node:
https://github.com/poanetwork/hbbft/blob/e0a85a54bbaf968107e4784ba2f8ca2c9e012fc4/src/honey_badger.rs#L412

3) decryption failure for a konwn proposer:
https://github.com/poanetwork/hbbft/blob/e0a85a54bbaf968107e4784ba2f8ca2c9e012fc4/src/honey_badger.rs#L435

each related to failures related to processing encrypted messages or ciphertexts. However the same logging improvement can be applied to other places possibly in other algorithms too where we, say, return `Ok(())` or simply `()` when some check fails, and the reason for that failure is likely to be some remote node that we can identify on the spot.

afck:
(Note that the log shouldn't just be command line output; the user should be able to programmatically learn about malicious nodes and act on it.)

PR #68, opened 2018-06-20
Encryption of proposals in HoneyBadger

### LINKED ISSUE #41: Encrypt proposed transactions in Honey Badger
Using https://github.com/poanetwork/hbbft/issues/40, encrypt the proposed batches of transactions in `HoneyBadger`.

vkomenda:
Close #41

### DIFF
--- a/src/honey_badger.rs
+++ b/src/honey_badger.rs
@@ -95,7 +95,7 @@ where
                 self.handle_common_subset_message(sender_id, epoch, cs_msg)
             }
             MessageContent::DecryptionShare { proposer_id, share } => {
-                self.handle_decryption_share_message(sender_id, epoch, &proposer_id, &share)
+                self.handle_decryption_share_message(sender_id, epoch, proposer_id, share)
             }
         }
     }
@@ -227,33 +227,20 @@ where
         &mut self,
         sender_id: &NodeUid,
         epoch: u64,
-        proposer_id: &NodeUid,
-        share: &DecryptionShare<Bls12>,
+        proposer_id: NodeUid,
+        share: DecryptionShare<Bls12>,
     ) -> HoneyBadgerResult<()> {
-        // Insert the share.
-        self.received_shares
-            .entry(epoch)
-            .and_modify(|epoch_entry| {
-                epoch_entry
-                    .entry(proposer_id.clone())
-                    .and_modify(|proposer_entry| {
-                        proposer_entry.insert(sender_id.clone(), share.clone());
-                    })
-                    .or_insert_with(|| {
-                        vec![(sender_id.clone(), share.clone())]
-                            .into_iter()
-                            .collect()
-                    });
-            })
-            .or_insert_with(|| {
-                vec![(
-                    proposer_id.clone(),
-                    vec![(sender_id.clone(), share.clone())]
-                        .into_iter()
-                        .collect(),
-                )].into_iter()
-                    .collect()
-            });
+        {
+            // Insert the share.
+            let received_shares = self
+                .received_shares
+                .entry(epoch)
+                .or_insert_with(BTreeMap::new);
+            let proposer_shares = received_shares
+                .entry(proposer_id.clone())
+                .or_insert_with(BTreeMap::new);
+            proposer_shares.insert(sender_id.clone(), share);
+        }
 
         if epoch == self.epoch && self.try_decrypt_proposer_selection(proposer_id) {
             self.try_output_batch()?;
@@ -276,7 +263,7 @@ where
             .decrypted_selections
             .iter()
             .flat_map(|(proposer_id, ser_batch)| {
-                // If serialization fails, the proposer of that batch is faulty. Ignore it.
+                // If deserialization fails, the proposer of that batch is faulty. Ignore it.
                 bincode::deserialize::<Vec<Tx>>(&ser_batch)
                     .ok()
                     .map(|proposed| (proposer_id.clone(), proposed))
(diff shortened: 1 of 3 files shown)


src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+                // TODO: Log the incorrect node `j`.
+                continue;
+            }
+            if let Some(share) = self.netinfo.secret_key().decrypt_share(&ciphertext) {
+                // Send the share to remote nodes.
+                self.messages.0.push_back(
+                    Target::All.message(
+                        MessageContent::DecryptionShare {
+                            proposer_id: proposer_id.clone(),
+                            share: share.clone(),
```
afck: Those shares can be large. I wonder whether we should actually put them into an `Rc` here instead of cloning them? (Probably rather some optimization to try out and benchmark later…)

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+                self.handle_decryption_share_message(&our_id, epoch, &proposer_id, &share)?;
+            } else {
+                warn!("Share decryption failed for proposer {:?}", proposer_id);
+                // TODO: Log the decryption failure.
+                continue;
+            }
+            let ciphertexts = match self.ciphertexts.entry(self.epoch) {
+                Entry::Occupied(entry) => entry.into_mut(),
+                Entry::Vacant(entry) => entry.insert(BTreeMap::new()),
+            };
```
afck: This should be equivalent to `or_insert_with(BTreeMap::new)`.

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+            ciphertexts.insert(proposer_id, ciphertext);
```
afck: Isn't it sufficient to just store the proposer ID, without the ciphertext?

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+        }
+        Ok(())
+    }
+
+    /// Checks whether the current epoch has output, and if it does, advances the epoch and
+    /// proposes a new batch.
+    fn process_output(&mut self) -> Result<(), Error> {
+        if let Some(cs_output) = self.take_current_output() {
+            self.send_decryption_shares(cs_output)?;
+            // TODO: May also check that there is no further output from Common Subset.
```
afck: Why? It should output exactly once. Or do you mean for later epochs?

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+                        .collect(),
+                )].into_iter()
+                    .collect()
+            });
+
+        if epoch == self.epoch && self.try_decrypt_proposer_selection(proposer_id) {
+            self.try_output_batch()?;
+        }
+
+        Ok(())
```
afck: We need to `pk.verify_decryption_share(&share, &ciphertext)` with the public key share `pk` of the sender as soon as we have both the ciphertext and the share. Then we drop all invalid shares, so we can be sure that any set of `f + 1` share will successfully yield the proposer's input.

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+            })
+            .or_insert_with(|| {
+                vec![(
+                    proposer_id.clone(),
+                    vec![(sender_id.clone(), share.clone())]
+                        .into_iter()
+                        .collect(),
+                )].into_iter()
+                    .collect()
+            });
```
afck: I think this could be simplified:

```rust
self.received_shares
    .entry(epoch)
    .or_insert_with(BTreeSet::new)
    .entry(proposer_id.clone())
    .or_insert_with(BTreeSet::new)
    .insert(sender_id.clone(), share.clone());
```

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+
+        // Deserialize the output.
+        let transactions: BTreeMap<NodeUid, Vec<Tx>> = self
+            .decrypted_selections
+            .iter()
+            .flat_map(|(proposer_id, ser_batch)| {
+                // If serialization fails, the proposer of that batch is faulty. Ignore it.
+                bincode::deserialize::<Vec<Tx>>(&ser_batch)
+                    .ok()
+                    .map(|proposed| (proposer_id.clone(), proposed))
```
afck: Something to keep in mind for #51: We could add another field (`faulty_nodes`?) to `Batch`, to report the nodes whose input didn't deserialize.

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+        if let Some(shares) = self.received_shares.get(&self.epoch) {
+            proposer_ids = Some(shares.keys().cloned().collect());
+        }
+        if let Some(proposer_ids) = proposer_ids {
+            // Try to output a batch if there is a non-empty set of proposers for which we have already received
+            // decryption shares.
+            if !proposer_ids.is_empty()
+                && proposer_ids
+                    .iter()
+                    .all(|proposer_id| self.try_decrypt_proposer_selection(proposer_id))
```
afck: The criteria here should probably be:

* `CommonSubset` has already output in this epoch.
* For every proposer from `CommonSubset`'s output, we have received `f + 1` valid decryption shares.

If there are at most `f` faulty nodes, it should be guaranteed that that eventually happens _and_ that decryption succeeds.

Calling `try_decrypt_proposer_selection` every time looks like it might be expensive: wouldn't that repeatedly decrypt and overwrite the same thing? (Or should that method just return `true` immediately, if there's already an entry in `decrypted_selections`?)

src/honey_badger.rs
```diff
@@ -200,43 +222,213 @@ where
...
+    /// this epoch.
+    fn all_selections_decrypted(&mut self) -> bool {
+        let ciphertexts = match self.ciphertexts.entry(self.epoch) {
+            Entry::Occupied(entry) => entry.into_mut(),
+            Entry::Vacant(entry) => entry.insert(BTreeMap::new()),
+        };
+        let all_ciphertext_proposers: BTreeSet<_> = ciphertexts.keys().collect();
+        let all_decrypted_selection_proposers: BTreeSet<_> =
+            self.decrypted_selections.keys().collect();
+        all_ciphertext_proposers.is_subset(&all_decrypted_selection_proposers)
```
afck: Can this be a proper subset or would it just be equal?

src/honey_badger.rs
vkomenda: I think not? A ciphertext is required for `PublicKeySet::decrypt` in `HoneyBadger::try_decrypt_proposer_selection`.

src/honey_badger.rs
vkomenda: This is exactly what I mean. We may output an error if there is a second output from the `CommonSubset`. This is the place where you had the `while` loop going through all `Some` outputs of the `CommonSubset`.

src/honey_badger.rs
```diff
@@ -75,10 +85,18 @@ where
...
+        if epoch < self.epoch {
+            // Ignore all messages from past epochs.
+            return Ok(());
+        }
+        match content {
+            MessageContent::CommonSubset(cs_msg) => {
                 self.handle_common_subset_message(sender_id, epoch, cs_msg)
             }
+            MessageContent::DecryptionShare { proposer_id, share } => {
+                self.handle_decryption_share_message(sender_id, epoch, &proposer_id, &share)
```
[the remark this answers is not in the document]
vkomenda: Then `clippy` starts complaining on "needless" pass-by-value.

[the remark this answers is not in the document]
afck: But not if you remove the `.clone()` in `handle_decryption_share_message`.

src/honey_badger.rs
vkomenda: I created a separate issue to test my proposed implementation: #75. I suggest adding that test after we finish working on this PR.

src/honey_badger.rs
```diff
@@ -60,6 +60,9 @@ pub struct HoneyBadger<Tx, NodeUid> {
     decrypted_selections: BTreeMap<NodeUid, Vec<u8>>,
     /// Ciphertexts output by Common Subset in an epoch.
     ciphertexts: BTreeMap<u64, BTreeMap<NodeUid, Ciphertext<Bls12>>>,
+    /// Those shares in `received_shares` that could not be verified on reception due to the absense
```
afck: Is keeping track about those even necessary? When we receive the ciphertext, doesn't that mean that none of the existing shares have been verified yet? I.e. won't this set always either be equal to the senders whose shares we've already received (if we don't have the ciphertext yet), or empty (if we have it)?

(Also typo: `absence`)

src/honey_badger.rs
vkomenda: I've been experimenting and this does indeed seem possible. Please check how you like the extra trait bounds!

I have one extra `clone` though, here in `send_decryption_shares`, where I move the share to an `Rc` and clone the `Rc` in the messages, and then I have to use the same share to pass by value to `handle_decryption_share_message`.

Also I will have to fix the tests tomorrow.

src/honey_badger.rs
vkomenda: This is indeed the case. Thanks for spotting this.

src/honey_badger.rs
```diff
@@ -104,7 +104,10 @@ where
     }
 
     fn next_message(&mut self) -> Option<TargetedMessage<Self::Message, NodeUid>> {
-        self.messages.pop_front()
+        let convert = |msg: TargetedMessage<InternalMessage<NodeUid>, NodeUid>| {
+            msg.map(|int_msg| int_msg.into_external())
```
afck: That probably doesn't need a closure: `map(InternalMessage::into_external)`.

src/honey_badger.rs
```diff
@@ -577,37 +563,96 @@ pub enum MessageContent<NodeUid> {
...
+
+    pub fn into_external(self) -> Message<NodeUid> {
+        let InternalMessage { epoch, content } = self;
+        Message {
+            epoch,
+            content: match content {
+                InternalMessageContent::CommonSubset(msg) => MessageContent::CommonSubset(msg),
+                InternalMessageContent::DecryptionShare { proposer_id, share } => {
+                    // FIXME: Handle errors.
+                    let share_copy = Rc::try_unwrap(share).unwrap();
```
afck: I think that _will_ usually fail: Most of the time, the local reference will live longer than the message.

I hadn't realized that `Rc` doesn't automatically implement the serde traits. Maybe instead of the internal-external distinction the message should just always have an `Rc`, but we should implement `Serialize` and `Deserialize` manually? (Or rather with some attribute `#[serde(with = "a_new_rc_serde_module")]`.)

But maybe this is all premature optimization; I'm happy to postpone that to a separate PR.

vkomenda: Issue #77

Issue #47, opened 2018-05-28
Distributed Key Generation


afck:
Implement distributed key generation for the threshold schemes (#38, #40).
See [Distributed Key Generation in the Wild](https://eprint.iacr.org/2012/377.pdf) and [A robust threshold elliptic curve digital signature providing a new verifiable secret sharing scheme](https://www.researchgate.net/profile/Ihab_Ali/publication/4205262_A_robust_threshold_elliptic_curve_digital_signature_providing_a_new_verifiable_secret_sharing_scheme/links/02e7e538f15726323a000000/A-robust-threshold-elliptic-curve-digital-signature-providing-a-new-verifiable-secret-sharing-scheme.pdf?origin=publication_detail).

afck:
As discussed, we might not need the full DKG in the wild, if we plan to do the key generation on the transaction log/blockchain itself. We will probably use a smart contract for this, but I'll describe it just in terms of special transactions. This protocol will have to run once at the end of each instance, i.e. before the set of validators/nodes changes, to generate the new keys. Once they have been distributed, the new Honey Badger instance starts, using the new set of nodes.

Each node (inluding the new one) `k` generates a random symmetric bivariate polynomial `f[k]` of degree `t` over the prime field, i.e. coefficients `coeff[k][i][j]` for `i` and `j` in `0..=t`. It will publish a committment to that polynomial, i.e. all values `commit[k][i][j] = coeff[k][i][j] * g1`, with which all values of the polynomial can be publicly verified:

```rust
f[k](x, y) * g1 ==
    (0..=t).map(|x| (0..=t).map(|y|
        x.pow(i) * y.pow(j) * commit[k][i][j]
    ).sum()).sum()
```

It also encrypts for each node `m` all values `f[k](m, y)` → `row_f[k][m]`. It commits a transaction:

```rust
Propose(commit[k], row_f[k])
```

Each node `m` decrypts `row_f[k][m]`, verifies the values and for each node `l` encrypts `f[k](m, l)` → `value_f[k][m][l]`. It commits a transaction:

```rust
Accept(k, value_f[k][m])
```

Once there are `2 * t + 1` transactions `Accept(k, _)` by different nodes in the log, we call `k`'s proposal "complete": Then at least `t + 1` of the `Accept`s must be honest, so every node `l` can reconstruct the univariate polynomial `f[k](_, l)` from it, and in particular the value `f[k](0, l)`.

[...]

PR #70, opened 2018-06-21
Implement SyncKeyGen.

### LINKED ISSUE #71: Consider making the cryptography tools `Bls12`-only.
It would simplify the code, other modules wouldn't have to be aware of `pairing`, and if https://github.com/ebfull/pairing/pull/87 is implemented, we will remove our general `Engine` serde implementations and replace them with the `Bls12`-specific ones.

afck:
This is a _synchronous_ key generation algorithm. We will use it in
`DynamicHoneyBadger`, on top of `HoneyBadger` to satisfy the synchrony
requirements.

It can also be used independently e.g. on top of a blockchain.

The generic arguments are removed from the cryptography types.
Closes #71

### DIFF
--- a/src/crypto/serde_impl.rs
+++ b/src/crypto/serde_impl.rs
@@ -80,3 +80,105 @@ pub mod projective_vec {
         Ok(wrap_vec.into_iter().map(|CurveWrap(c, _)| c).collect())
     }
 }
+
+/// Serialization and deserialization of vectors of field elements.
+pub mod field_vec {
+    use std::borrow::Borrow;
+    use std::marker::PhantomData;
+
+    use pairing::{PrimeField, PrimeFieldRepr};
+    use serde::de::Error as DeserializeError;
+    use serde::ser::Error as SerializeError;
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+    /// A wrapper type to facilitate serialization and deserialization of field elements.
+    pub struct FieldWrap<F, B>(B, PhantomData<F>);
+
+    impl<F, B> FieldWrap<F, B> {
+        pub fn new(f: B) -> Self {
+            FieldWrap(f, PhantomData)
+        }
+    }
+
+    impl<F> FieldWrap<F, F> {
+        pub fn into_inner(self) -> F {
+            self.0
+        }
+    }
+
+    impl<F: PrimeField, B: Borrow<F>> Serialize for FieldWrap<F, B> {
+        fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+            let mut bytes = Vec::new();
+            self.0
+                .borrow()
+                .into_repr()
+                .write_be(&mut bytes)
+                .map_err(|_| S::Error::custom("failed to write bytes"))?;
+            bytes.serialize(s)
+        }
+    }
+
+    impl<'de, F: PrimeField> Deserialize<'de> for FieldWrap<F, F> {
+        fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+            let bytes: Vec<u8> = Deserialize::deserialize(d)?;
+            let mut repr = F::zero().into_repr();
+            repr.read_be(&bytes[..])
+                .map_err(|_| D::Error::custom("failed to write bytes"))?;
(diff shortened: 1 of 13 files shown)


afck:
@andogro: I also added and modified some module-level documentation. Please let me know if you find any issues.

It's probably still way too short, but I'm not sure how far we can and should go there. I'd like to try and convey the main ideas, but refer to the papers and the code for details.

afck:
(Not closing https://github.com/poanetwork/hbbft/issues/47 with this one yet: I'll close it when `DynamicHoneyBadger` is complete.)

src/sync_key_gen.rs
```diff
@@ -0,0 +1,238 @@
...
+//! If there were a trusted dealer, they would generate a `BivarPoly` of degree `t` and publish
+//! the `BivarCommitment`, with which the polynomial's values can be publicly verified. They'd
+//! then send _row_ `m > 0` to node number `m`. Node `m`, in turn, sends _value_ `s` to node number
+//! `s`. Then if `2 * t + 1` nodes confirm that they received a valid row, and there are at most
+//! `t` faulty nodes, then at least `t + 1` honest nodes sent on an entry of every other node's
+//! column to that node. So we know that every node can now reconstruct its column and the value at
+//! `0` of its column. These values all lie on a univariate polynomial of degree `t`, so they can
+//! be used as secret keys.
+//!
+//! To avoid trusting a single dealer, we make sure that at least `t + 1` nodes use the above
```
andogro: In our **dealerless** environment, at least `t+1` nodes each generate a polynomial using the method above. The sum of the secret keys from each node is used as the master secret key. No single node knows the sum of the secret master key.

src/sync_key_gen.rs
```diff
@@ -0,0 +1,238 @@
...
+//! signatures and encryption. The secret master key is not known by anyone. The protocol succeeds
+//! if up to `threshold` nodes are faulty.
+//!
+//! # How it works
+//!
+//! The algorithm is based on ideas from
+//! [Distributed Key Generation in the Wild](https://eprint.iacr.org/2012/377.pdf) and
+//! [A robust threshold elliptic curve digital signature providing a new verifiable secret sharing scheme](https://www.researchgate.net/profile/Ihab_Ali/publication/4205262_A_robust_threshold_elliptic_curve_digital_signature_providing_a_new_verifiable_secret_sharing_scheme/links/02e7e538f15726323a000000/A-robust-threshold-elliptic-curve-digital-signature-providing-a-new-verifiable-secret-sharing-scheme.pdf?origin=publication_detail).
+//!
+//! If there were a trusted dealer, they would generate a `BivarPoly` of degree `t` and publish
```
andogro: For readability I would recommend breaking down into steps. For example:
In a trusted dealer scenario, the following steps occur:
1. Dealer generates a `BivarPoly` of degree `t` and publishes the `BivarCommitment` which is used to publicly verify the polynomial's values.
2. Dealer sends _row_ `m > 0` to node number `m`.
3. Node `m`, in turn, sends _value_ `s` to node number `s`.
4. If (unsure if we can say 'This process continues until?' or 'Once?') `2 * t + 1` nodes confirm they have received a valid row and there are at most `t` faulty nodes, we know that at least `t + 1` honest nodes sent on an entry of every other node’s column to that node.
5. This means every node can reconstruct its column and the value at `0` of its column. 
6. These values all lie on a univariate polynomial of degree `t` and can be used as secret keys.

afck:
Thank you! I rebased and improved the docs.

src/crypto/serde_impl.rs
```diff
@@ -80,3 +80,105 @@ pub mod projective_vec {
...
+
+#[cfg(test)]
+mod tests {
+    use bincode;
+    use pairing::bls12_381::Bls12;
+    use pairing::Engine;
+    use rand::{self, Rng};
+
+    #[derive(Debug, Serialize, Deserialize)]
+    pub struct Vecs<E: Engine> {
```
vkomenda: Should the tests be kept with the `Engine` parameter? Since you are using `Bls12` anyway, maybe the tests can be simplified too?

afck: Yes, I kept the `serde_impl` module generic for now. If https://github.com/ebfull/pairing/pull/87 gets merged, this module can be removed anyway. Otherwise I'll update it next week in a separate PR.

PR #76, opened 2018-06-22
The schedule 0, 1, get_coin(), ...  

### LINKED ISSUE #69: Optimize Common Coin using a schedule
Computations involving `pairing` are relatively slow. Simulation reveals that with even more messages requiring such computations to be performed by each node, the node run times soar. The main contributor to the run time increase is the common coin algorithm. We should try to reduce the number of encrypted `Coin` messages as suggested in https://github.com/amiller/HoneyBadgerBFT/issues/63. That will also lead to the reduction in the number of `Conf` messages.

vkomenda:
Close #69 

This branch depends on PR #68.


src/agreement/mod.rs
```diff
@@ -201,6 +211,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
                     netinfo,
                     Nonce::new(invocation_id.as_ref(), session_id, proposer_i, 0),
                 ),
+                coin_schedule: CoinSchedule::False,
```
afck: I think there was an argument for starting with `True`: In the best case scenario, where everyone gets their proposals delivered in time, everyone would vote `true` in every instance, and it would be nice to terminate in the first round then.

Also, it is probably generally preferred to err on the `true` side and include more elements in the common subset, and therefore more transactions in the epoch.

src/agreement/mod.rs
```diff
@@ -344,29 +390,47 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
         }
         Ok(())
     }
 
-    // Propagates Common Coin messages to the top level.
+    /// Computes the coin schedule for the current `Agreement` epoch.
+    fn coin_schedule(&self) -> CoinSchedule {
+        match self.epoch % 3 {
+            0 => CoinSchedule::False,
+            1 => CoinSchedule::True,
```
afck: As above, I'd swap `True` and `False` here.

src/agreement/mod.rs
```diff
@@ -466,7 +531,10 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
             *self.netinfo.node_index(&self.proposer_id).unwrap(),
             self.epoch,
         );
+        // TODO: Don't spend time creating a `CommonCoin` instance in epochs where the common coin
+        // is known.
         self.common_coin = CommonCoin::new(self.netinfo.clone(), nonce);
+        self.coin_schedule = self.coin_schedule();
```
afck: Do we need the coin schedule as a member variable at all? Isn't it easier to just call `self.coin_schedule()` where it's needed?

vkomenda: I didn't want to make a call on every `Aux` message.

PR #78, opened 2018-06-22
Reference-counted decryption shares

### LINKED ISSUE #77: Reduce the memory footprint of a decryption share in HoneyBadger
PR #68 introduced a new kind of `HoneyBadger` message that contains a decryption share of the `CommonCoin` for the threshold encryption scheme. The decryption share contains a vector of 48 bytes which is the compressed representation of a group element. #68 took a rather relaxed view on optimization of memory usage and simply cloned the same share from local storage to a message. This can be optimised by wrapping the locally stored share into an `Rc` and cloning that `Rc` to the message instead of the whole vector. However, since serialization is not derivable for `Rc`, that required defining custom `Serialization` and `Deserialization` instances for `HoneyBadger::Message`, or possibly using an attribute `#[serde(with = "a_new_rc_serde_module")]`.

vkomenda:
Close #77 

I ran simulation with and without optimization and I'm not sure if there is any improvement. It seems that the compiler option `--release` might optimize cloning of decryption shares even without reference counting.

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -194,6 +194,16 @@ impl Hash for DecryptionShare {
     }
 }
 
+impl DecryptionShare {
+    pub fn new(elt: G1) -> Self {
+        DecryptionShare(elt)
+    }
+
+    pub fn group_element(&self) -> &G1 {
+        &self.0
+    }
+}
+
 /// A public key and an associated set of public key shares.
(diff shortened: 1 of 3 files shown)


afck:
Sorry for the confusion, that was my fault! (See #77.)
I'd say it's not worth the optimization then.

PR #79, opened 2018-06-25
Make SyncKeyGen NodeUid-aware.

afck:
This allows the caller to address nodes by ID instead of by index.

Also contains a few other minor changes that will be needed for
`DynamicHoneyBadger`.

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -104,7 +104,7 @@ impl Signature {
 }
 
 /// A secret key, or a secret key share.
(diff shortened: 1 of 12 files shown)


afck:
…and make `PublicKeySet` behave more like a map of keys instead of a vec of keys? That would be nice, but `NetworkInfo` contains a `PublicKeySet`, and for this to work, `PublicKeySet` would need to contain something like a `NetworkInfo` itself: It needs to be aware of all existing node IDs. Not sure what's the best solution for this.

PR #86, opened 2018-06-26
Add DynamicHoneyBadger (part 1).

afck:
`DynamicHoneyBadger` is supposed to allow adding and removing nodes. (See https://github.com/poanetwork/hbbft/issues/47#issuecomment-394640406) Internally, it performs Distributed Key Generation and restarts Honey Badger, but all that should be hidden away from the user.

This is still incomplete in several ways, but removing a node works.

### DIFF
--- a/src/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger.rs
@@ -87,14 +87,12 @@ where
     type Error = Error;
 
     fn input(&mut self, input: Self::Input) -> Result<()> {
-        self.honey_badger.input(input.into())?;
+        let tx = self.input_to_tx(input)?;
+        self.honey_badger.input(tx)?;
         self.process_output()
     }
 
     fn handle_message(&mut self, sender_id: &NodeUid, message: Self::Message) -> Result<()> {
-        if !self.netinfo.all_uids().contains(sender_id) {
-            return Err(ErrorKind::UnknownSender.into());
-        }
         match message {
             Message::HoneyBadger(start_epoch, hb_msg) => {
                 self.handle_honey_badger_message(sender_id, start_epoch, hb_msg)
@@ -158,6 +156,9 @@ where
             self.incoming_queue.push(entry);
             return Ok(());
         }
+        if !self.netinfo.all_uids().contains(sender_id) {
+            return Err(ErrorKind::UnknownSender.into());
+        }
         // Handle the message and put the outgoing messages into the queue.
         self.honey_badger.handle_message(sender_id, message)?;
         self.process_output()?;
@@ -176,13 +177,14 @@ where
             let change = self.current_majority().cloned();
             // Add the user transactions to `batch` and handle votes and DKG messages.
             for (id, tx_vec) in hb_batch.transactions {
-                let entry = batch.transactions.entry(id.clone());
+                let entry = batch.transactions.entry(id);
                 let id_txs = entry.or_insert_with(Vec::new);
                 for tx in tx_vec {
                     use self::Transaction::*;
+                    info!("{:?} output {:?}.", self.netinfo.our_uid(), tx);
                     match tx {
                         User(tx) => id_txs.push(tx),
-                        Change(change) => self.insert_vote(id.clone(), change),
+                        Change(s_id, change, sig) => self.handle_vote(s_id, change, &sig)?,
                         Propose(s_id, propose, sig) => self.handle_propose(&s_id, propose, &*sig)?,
                         Accept(s_id, accept, sig) => self.handle_accept(&s_id, accept, &*sig)?,
                     }
@@ -191,6 +193,9 @@ where
             // If DKG completed, apply the change.
             if let Some(change) = change.as_ref() {
                 if let Some((pub_key_set, sk)) = self.get_key_gen_output() {
+                    let sk = sk.unwrap_or_else(|| {
+                        ClearOnDrop::new(Box::new(self.netinfo.secret_key().clone()))
+                    });
                     self.start_epoch = hb_batch.epoch + 1;
                     self.apply_change(change, pub_key_set, sk)?;
                     batch.change = Some(change.clone());
@@ -217,6 +222,18 @@ where
             }
         }
         Ok(())
+    }
+
+    /// Converts the input into a transaction to commit via Honey Badger.
+    fn input_to_tx(&self, input: Input<Tx, NodeUid>) -> Result<Transaction<Tx, NodeUid>> {
+        Ok(match input {
+            Input::User(tx) => Transaction::User(tx),
+            Input::Change(change) => {
+                let our_id = self.our_id().clone();
+                let sig = self.sign(&change)?;
+                Transaction::Change(our_id, change, sig)
+            }
+        })
     }
 
     /// Restarts Honey Badger with a new set of nodes, and resets the Key Generation.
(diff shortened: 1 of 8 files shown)


src/dynamic_honey_badger.rs
```diff
@@ -0,0 +1,449 @@
...
+            Change::Remove(ref id) => all_uids.remove(id),
+            Change::Add(ref id, _) => all_uids.insert(id.clone()),
+        } {
+            debug!("No-op change: {:?}", change);
+        }
+        let netinfo = NetworkInfo::new(self.our_id().clone(), all_uids, sk, pub_key_set);
+        self.netinfo = netinfo.clone();
+        // TODO: Drop the buffer if this node was removed, to become an observer.
+        let buffer = self.honey_badger.drain_buffer();
+        self.honey_badger = HoneyBadger::new(Rc::new(netinfo), self.batch_size, buffer)?;
```
vkomenda: Can you update the batch size here given that the number of nodes has increased?

vkomenda: ...or decreased.

src/dynamic_honey_badger.rs
```diff
@@ -0,0 +1,449 @@
...
+            honey_badger,
+            key_gen: None,
+            incoming_queue: Vec::new(),
+            messages: MessageQueue(VecDeque::new()),
+            output: VecDeque::new(),
+        };
+        Ok(dyn_hb)
+    }
+
+    fn convert_transaction(&self, input: Input<Tx, NodeUid>) -> Transaction<Tx, NodeUid> {
```
DrPeterVanNostrand: I think that a better abstraction here would be to implement a conversion trait instead of using a method (because `&self` isn't being used):

```rust
impl<Tx, NodeUid> From<Input<Tx, NodeUid>> for Transaction<Tx, NodeUid> {
    match input {
        Input::User(tx) => Transaction::User(tx),
        Input::Change(change) => Transaction::Change(change),
    }
}
```

Then converting from `Input -> Transaction` could be written:

```rust
let tx = input.into();
```

or, if the type parameters are necessary:

```rust
let tx: Transaction<Self::Input, Self::NodeUid> = input.into();
```

src/dynamic_honey_badger.rs
afck: It depends on the exact requirements; there is still a [TODO](https://github.com/poanetwork/hbbft/blob/b3b3994ec1dd42d7a4765dc220d1e5470e35c6f5/src/honey_badger.rs#L46) in the code for that. I created issue https://github.com/poanetwork/hbbft/issues/88 for it.

For now, the code does the right thing in the situation of the issue's first bullet point, i.e. where we want a constant block size.

src/dynamic_honey_badger.rs
afck: You're right, I'll change it. (When I originally wrote the method, I still needed `self`.)

src/dynamic_honey_badger.rs
```diff
@@ -0,0 +1,450 @@
...
+        if match change {
+            Change::Remove(id) => pub_keys.remove(&id).is_none(),
+            Change::Add(id, pub_key) => pub_keys.insert(id, pub_key).is_some(),
+        } {
+            debug!("No-op change: {:?}", self.current_majority().unwrap());
+        }
+        // TODO: This needs to be the same as `num_faulty` will be in the _new_
+        // `NetworkInfo` if the change goes through. It would be safer to deduplicate.
+        let threshold = (pub_keys.len() - 1) / 3;
+        let sk = self.netinfo.secret_key().clone();
```
DrPeterVanNostrand: Hmm... this allocates a secret key on stack. I'm not sure what can be done about it though.

I guess we could change `self.netinfo.secret_key()` to return a pointer to the heap allocated secret-key, and then change `SyncKeyGen.sec_key` to be of type `ClearOnDrop<Box<SecretKey>>`.

afck: That's true. I wonder whether the right thing to do would be to turn what's currently `SecretKey` into a private `SecretKeyInner`, and put a `ClearOnDrop<Box<SecretKeyInner>>` inside the new `SecretKey`, so that outside of the `crypto` module there's no way to circumvent it anymore?

afck: I created https://github.com/poanetwork/hbbft/issues/89 for it.

src/dynamic_honey_badger.rs
```diff
@@ -0,0 +1,450 @@
...
+            let decrement = |count: &mut usize| {
+                *count -= 1;
+                *count
+            };
+            if Some(0) == self.vote_counts.get_mut(&old_change).map(decrement) {
+                self.vote_counts.remove(&old_change);
+            }
+        }
+        *self.vote_counts.entry(change).or_insert(0) += 1;
+    }
```
vkomenda: Would the following code be equivalent to `insert_vote` and yet shorter?
```
if self.votes.insert(id, change.clone()).is_none() {
    *self.vote_counts.entry(change).or_insert(0) += 1;
}
```

src/dynamic_honey_badger.rs
```diff
@@ -0,0 +1,450 @@
...
+        self.change.as_ref()
+    }
+}
+
+/// A message sent to or received from another node's Honey Badger instance.
+#[cfg_attr(feature = "serialization-serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone)]
+pub enum Message<NodeUid> {
+    /// A message belonging to the `HoneyBadger` algorithm started in the given epoch.
+    HoneyBadger(u64, honey_badger::Message<NodeUid>),
```
vkomenda: I suppose there are going to be other messages in this `enum`. Otherwise it would have been a newtype.

src/dynamic_honey_badger.rs
afck: Not quite. Sorry for my lousy code comments: The idea is that every node can only have one active vote at a time; voting again revokes the previous vote.
So when we insert a vote, we need to increment the count for that vote, and also decrement the count for the _previous_ vote of that node. Not sure how that can be expressed more concisely.

src/dynamic_honey_badger.rs
afck: Yes, we will need additional messages so that the joining node can take part in the key generation (DKG): It will have to send its `Propose` and `Accept`s as messages to the other nodes.

In fact, I think I'll insert this messaging step for all nodes, to prevent censorship. So every node will send its votes and DKG messages first as messages, and everyone who receives a signed message from a peer will propose it in Honey Badger. (Instead of only proposing their own messages, like they do now.)

afck:
I think something is still wrong with this (or with the test): Shouldn't the to-be-removed node panic [here](https://github.com/poanetwork/hbbft/blob/78fdf6354048c69f4657bac9006ee95f9930bfd5/src/sync_key_gen.rs#L114)?.

src/dynamic_honey_badger.rs
vkomenda: One way would be to avoid duplication by removing `vote_counts` but it does look like a handy optimization. Maybe you could mention this semantics of revoking votes in a comment to `Change` so that it is visible?

afck: No, you're right, removing it is better: the optimization is pointless since it's only computed twice per epoch. Also it required more cloning, and if there are a lot of votes it might even be _slower_, so I removed it.

PR #92, opened 2018-06-27
Avoid redundant key computations.

afck:
This precomputes the keys in `NetworkInfo` instead of doing the calculation on every call to `public_key_share`.


afck:
The variance is very high, so it's hard to benchmark, but I ran the full test suite twice on `master` and twice with these optimizations: User time went down from 3m23,418s + 3m54,841s to 2m58,753s + 3m4,728s.

PR #93, opened 2018-06-27
Add node transaction messages to Dynamic HB.

afck:
This allows adding and removing full peers to/from the pool of observers.

To allow a joining node to participate in the on-chain key generation, messages are signed and broadcast to all nodes, and the receiver inputs them as transactions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,13 @@ Honey Badger is a modular library composed of several independent algorithms.  T
 
 In an optimal networking environment, output includes data sent from each node. In an adverse environment, the output is an agreed upon subset of data. Either way, the resulting output contains a batch of transactions which is guaranteed to be consistent across all nodes.  
 
+In addition to **full nodes**, the algorithms support **observers**: These don't actively participate, and don't need to be trusted, but they receive the output as well, and are able to verify it under the assumption that more than two thirds of the full nodes are correct.
+
 ## Algorithms
 
-- [ ] **[Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/honey_badger.rs):** The top level protocol proceeds in epochs using the protocols below. 
+- [x] **[Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/honey_badger.rs):** Each node inputs transactions. The protocol outputs a sequence of batches of transactions.
+
+- [ ] **[Dynamic Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/dynamic_honey_badger.rs):** This works like Honey Badger itself, but in addition allows dynamically adding and removing nodes to/from the network.
 
 - [x] **[Subset](https://github.com/poanetwork/hbbft/blob/master/src/common_subset.rs):** Each node inputs data. The nodes agree on a subset of suggested data. 
 
--- a/README.md
+++ b/README.md
@@ -21,13 +21,13 @@ Honey Badger is a modular library composed of several independent algorithms.  T
 
 In an optimal networking environment, output includes data sent from each node. In an adverse environment, the output is an agreed upon subset of data. Either way, the resulting output contains a batch of transactions which is guaranteed to be consistent across all nodes.  
 
-In addition to **full nodes**, the algorithms support **observers**: These don't actively participate, and don't need to be trusted, but they receive the output as well, and are able to verify it under the assumption that more than two thirds of the full nodes are correct.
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -21,9 +21,13 @@ Honey Badger is a modular library composed of several independent algorithms.  T
...
 In an optimal networking environment, output includes data sent from each node. In an adverse environment, the output is an agreed upon subset of data. Either way, the resulting output contains a batch of transactions which is guaranteed to be consistent across all nodes.  
 
+In addition to **full nodes**, the algorithms support **observers**: These don't actively participate, and don't need to be trusted, but they receive the output as well, and are able to verify it under the assumption that more than two thirds of the full nodes are correct.
+
 ## Algorithms
 
-- [ ] **[Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/honey_badger.rs):** The top level protocol proceeds in epochs using the protocols below. 
+- [x] **[Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/honey_badger.rs):** Each node inputs transactions. The protocol outputs a sequence of batches of transactions.
+
+- [ ] **[Dynamic Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/dynamic_honey_badger.rs):** This works like Honey Badger itself, but in addition allows dynamically adding and removing nodes to/from the network.
```
andogro: Possible suggestion - unsure if modified does it justice :): 
A modified Honey Badger where nodes can dynamically add and remove other nodes  to/from the network.

README.md
```diff
@@ -21,9 +21,13 @@ Honey Badger is a modular library composed of several independent algorithms.  T
 
 In an optimal networking environment, output includes data sent from each node. In an adverse environment, the output is an agreed upon subset of data. Either way, the resulting output contains a batch of transactions which is guaranteed to be consistent across all nodes.  
 
+In addition to **full nodes**, the algorithms support **observers**: These don't actively participate, and don't need to be trusted, but they receive the output as well, and are able to verify it under the assumption that more than two thirds of the full nodes are correct.
```
andogro: change both instances of full nodes to new terminology once determined.

src/dynamic_honey_badger.rs
```diff
@@ -253,133 +380,175 @@ where
...
+        *entry += 1;
+        if *entry * 2 > netinfo.num_nodes() {
+            return Some(change);
         }
-        None
     }
+    None
 }
 
 /// The transactions for the internal `HoneyBadger` instance: this includes both user-defined
```
vkomenda: Should node addition and removal be removed from the comment after the update?

src/dynamic_honey_badger.rs
```diff
@@ -438,18 +609,41 @@ impl<Tx, NodeUid: Ord> Batch<Tx, NodeUid> {
...
-    pub fn change(&self) -> Option<&Change<NodeUid>> {
-        self.change.as_ref()
+    /// Returns whether any change to the set of participating nodes is in progress or was
+    /// completed in this epoch.
+    pub fn change(&self) -> &ChangeState<NodeUid> {
+        &self.change
     }
 }
 
+/// An internal message that gets committed via Honey Badger to communicate synchronously.
```
vkomenda: I find this sentence confusing. Maybe instead of referring to synchronous communication the comment could summarize the message content in a few words?

src/dynamic_honey_badger.rs
```diff
@@ -1,3 +1,45 @@
...
+//! _batches_ of transactions. The protocol proceeds in _epochs_, starting at number 0, and outputs
+//! one batch in each epoch. It never terminates: It handles a continuous stream of incoming
+//! transactions and keeps producing new batches from them. All correct nodes will output the same
+//! batch for each epoch.
+//!
+//! Unlike Honey Badger, this algorithm allows dynamically adding new full peers from the pool of
+//! observer nodes, and turning full peers back into observers. As a signal to initiate that
+//! process, it defines a special `Change` input variant, which contains either a vote
+//! `Add(node_id, public_key)`, to add a new full peer, or `Remove(node_id)` to remove it. Each
+//! full peer can have at most one active vote, and casting another vote revokes the previous one.
```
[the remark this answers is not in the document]
afck: No, any number of votes is fine, and it's always the last one — i.e. the last one that came _out_ of Honey Badger in a batch — that counts. Every vote therefore revokes the previous one and replaces it.

I think we should allow any number of votes per node. It can always happen that e.g. the validators voted to add a node but then DKG never completes (e.g. because the node went offline). Then they should be able to abort the addition, and vote for something else (e.g. to add a different node).

PR #95, opened 2018-06-28
Move broadcast docs to module-level.

afck:
Also replaces the word "good" with "correct".

### DIFF
--- a/src/broadcast.rs
+++ b/src/broadcast.rs
@@ -1,3 +1,44 @@
+//! # Broadcast
+//!
+//! The Reliable Broadcast Protocol assumes a network of `N` nodes that send signed messages to
+//! each other, with at most `f` of them faulty, where `3 * f < N`. Handling the networking and
+//! signing is the responsibility of this crate's user: only when a message has been verified to be
+//! "from node i", it can be handed to the `Broadcast` instance. One of the nodes is the "proposer"
+//! who sends a value. Under the above conditions, the protocol guarantees that either all or none
+//! of the correct nodes output a value, and that if the proposer is correct, all correct nodes
+//! output the proposed value.
+//!
+//! ## How it works
+//!
+//! * The proposer uses a Reed-Solomon code to split the value into `N` chunks, `f + 1` of which
+//! suffice to reconstruct the value. These chunks are put into a Merkle tree, so that with the
+//! tree's root hash `h`, branch `bi` and chunk `si`, the `i`-th chunk `si` can be verified by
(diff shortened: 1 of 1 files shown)


afck:
I think complicated explanations should in general be module-level docs, not `struct` docs.

/cc @DrPeterVanNostrand, @vkomenda

src/broadcast.rs
```diff
@@ -1,3 +1,44 @@
...
+//! contains the `i`-th share of my value."
+//! * Every (correct) node that receives `Value(pi)` from the proposer sends it on to everyone else
+//! as `Echo(pi)`. An `Echo` translates to: "I have received `pi` directly from the proposer." If
+//! the proposer sends another `Value` message, that is ignored.
+//! * So every node that has received at least `f + 1` `Echo` messages with the same root
+//! hash will be able to decode a value.
+//! * Every node that has received `N - f` `Echo`s with the same root hash from different nodes
+//! knows that at least `f + 1` _correct_ nodes have sent an `Echo` with that hash to everyone, and
+//! therefore everyone will eventually receive at least `f + 1` of them. So upon receiving `N - f`
+//! `Echo`s, they send a `Ready(h)` to everyone to indicate that. `Ready` translates to: "I know
```
andogro: to indicate that -> to indicate this eventuality.

src/broadcast.rs
```diff
@@ -1,3 +1,44 @@
...
+//! that everyone will eventually be able to decode the value." Moreover, since every correct node
+//! only ever sends one kind of `Echo` message, this cannot happen for two different root hashes.
```
andogro: 'this' is a little confusing in this sentence - 'this process' or something like...  Since every correct node can only send one kind of `Echo` message, there is no danger of relaying two different root hashes.

src/broadcast.rs
afck: Not sure; to me "eventuality" sounds like something that's not necessarily supposed to happen. The `Ready` message indicates that I know that the others _will definitely_ receive `f + 1` correct `Echo`s, and that they will be able to decode.

I just used the word "eventually" to highlight that the others haven't necessarily received their `f + 1` `Echo`s _yet_, and (since we are asynchronous and don't make any timing assumptions) I don't know how long it will take until they do. In that context, that word is common, I think.

afck:
Thank you! I updated the text; please take another look.

src/broadcast.rs
andogro: maybe 'this inevitable outcome' or 'eventual outcome'.  Trying to clarify 'that' in this context

andogro:
Reads well to me.

Issue #88, opened 2018-06-26
Enforce batch size, add configuration options.


afck:
There are different ways one may want to specify the batch size in Honey Badger (see the [TODO](https://github.com/poanetwork/hbbft/blob/b3b3994ec1dd42d7a4765dc220d1e5470e35c6f5/src/honey_badger.rs#L46)):

* Currently the batch size itself is a parameter; each node proposes [`batch_size / num_nodes`](https://github.com/poanetwork/hbbft/blob/b3b3994ec1dd42d7a4765dc220d1e5470e35c6f5/src/honey_badger.rs#L182) transactions per epoch. This is desirable e.g. in a blockchain with a fixed maximum block size.
* To achieve its asymptotic complexity, the paper suggests a batch size of `lambda * num_nodes * num_nodes * log(num_nodes)`. Since this comes at the cost of a quickly increasing latency, it's not always the right choice in practice.
* There are other options in between with varying tradeoffs, e.g. `lambda * num_nodes` (constant proposal size per node) or `lambda * num_nodes * log(num_nodes)`, etc.

We could make the batch size specification an enum that allows picking one of these options. This is particularly important for `DynamicHoneyBadger`, where the number of nodes can change.

Also, we should _enforce_ the limit: If another node is sending us a proposal that exceeds the per-node allowance, the transfer needs to be interrupted, the proposal ignored and the node reported/logged.

afck:
With #104, this should only affect the implementation of the default queue, not Honey Badger.

PR #96, opened 2018-06-28
Add a max_future_epochs option to Honey Badger.

### LINKED ISSUE #91: Delay sending messages for future epochs in Honey Badger?
For `DynamicHoneyBadger` it's important to determine which epochs a newly joined observer node can fully see (the new node needs to participate in the "on-chain" key generation, and it needs to see all epochs involved in that process). So if `epoch0` output transactions that represent the decision to include a new node `n`, and if the key generation process starts in `epoch0 + 1`, then we need to make sure that node `n`, still as an observer, receives all messages belonging to `epoch0 + 1`.

This is mostly the responsibility of the user: We will document that once a batch contains the information about a new node, all future `Target::All` messages need to be sent to that node, too.

Currently, however, `HoneyBadger` can run several `CommonSubset` instances in parallel, i.e. it will handle messages belonging to future epochs and even send some `epoch0 + 1` messages (e.g. broadcast `Echo`s) before `epoch0` has output. We could store all later outgoing messages in a cache, and only return them after `epoch0`'s output.

On the other hand, handling them earlier may be a desirable optimization: it allows lagging nodes to still participate in and speed up later epochs before they even arrive there? If that's indeed valuable, we could just introduce a limit and, say, only hold back messages from `epoch0 + 5` and later. Then we could start key generation in `epoch0 + 5` and still know that node `n` will be able to participate.

afck:
This delays handling of messages from epochs that are too far ahead.

Closes #91

### DIFF
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -436,7 +436,7 @@ fn main() {
             sk_set.secret_key_share(id.0 as u64),
             pk_set.clone(),
         ));
-        HoneyBadger::new(netinfo, args.flag_b, txs.clone()).expect("Instantiate honey_badger")
+        HoneyBadger::new(netinfo, args.flag_b, 0, txs.clone()).expect("Instantiate honey_badger")
     };
     let hw_quality = HwQuality {
         latency: Duration::from_millis(args.flag_lag),
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -25,7 +25,7 @@ use serde::Serialize;
 use signifix::{metric, TryFrom};
 
 use hbbft::crypto::SecretKeySet;
-use hbbft::honey_badger::{Batch, HoneyBadger};
+use hbbft::honey_badger::{Batch, HoneyBadger, HoneyBadgerBuilder};
 use hbbft::messaging::{DistAlgorithm, NetworkInfo, Target};
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -436,7 +436,10 @@ fn main() {
             sk_set.secret_key_share(id.0 as u64),
             pk_set.clone(),
         ));
-        HoneyBadger::new(netinfo, args.flag_b, 0, txs.clone()).expect("Instantiate honey_badger")
+        HoneyBadgerBuilder::new(netinfo)
+            .batch_size(args.flag_b)
+            .build_with_transactions(txs.clone())
+            .expect("Instantiate honey_badger")
     };
     let hw_quality = HwQuality {
         latency: Duration::from_millis(args.flag_lag),
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -25,7 +25,7 @@ use serde::Serialize;
 use signifix::{metric, TryFrom};
 
 use hbbft::crypto::SecretKeySet;
-use hbbft::honey_badger::{Batch, HoneyBadger, HoneyBadgerBuilder};
+use hbbft::honey_badger::{Batch, HoneyBadger};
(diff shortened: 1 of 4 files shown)


afck:
The constructor is getting unwieldy, and we'll probably extend it further soon (#88, and maybe make encryption optional). How do you feel about making it a builder pattern?

```rust
let hb = HoneyBadger::builder(netinfo)
    .with_batch_size(batch_size)
    .with_max_future_epochs(max_future_epochs)
    .with_initial_transactions(txs)
    .build()?;
```

/cc @DrPeterVanNostrand @c0gent

c0gent:
Definitely use the builder pattern. No question.

c0gent:
There are two main types of builders if you're not familiar (you probably are): https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

c0gent:
Personally I would add a `HoneyBadger::builder` method also, as you originally had in mind. Makes it nice and simple for the user.

afck:
I actually tried that, but since `HoneyBadger<Tx, NodeUid>` has two type arguments and `HoneyBadgerBuilder<NodeUid>` has only one, it complains that in `HoneyBadger::builder(…)` it can't resolve the `Tx`:

```
error[E0283]: type annotations required: cannot resolve `_: serde::Serialize`
   --> src/dynamic_honey_badger.rs:126:28
    |
126 |         let honey_badger = HoneyBadger::builder(Rc::new(netinfo.clone()))
    |                            ^^^^^^^^^^^^^^^^^^^^
    |
note: required by `<honey_badger::HoneyBadger<Tx, NodeUid>>::builder`
   --> src/honey_badger.rs:201:5
    |
201 |     pub fn builder(netinfo: Rc<NetworkInfo<NodeUid>>) -> Builder<NodeUid> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

So you'd have to write `HoneyBadger::<Tx, _>::builder(…)`. Not sure how to fix this.

c0gent:
I'll make the changes in a gist and you can take a look. Basically you just add an extra pointless lifetime to the builder and add a `Phantom` field to it.

c0gent:
https://github.com/poanetwork/hbbft/pull/97

c0gent:
Can you rebase my commit message please? That's a bit embarrassing, hah. 

To do that just stash any changes you have:
`git stash` 

Then:
`git rebase -i HEAD~2` 

Then when the editor opens the file, on my commit, replace `pick` with `r` or `reword`. Another editor will pop up and you can change lifetime to type. 

Then:
`git stash pop` to open back up any changes you had.

Whoops! Thanks!

afck:
Done. History rewritten. :+1:

c0gent:
My reputation is safe... for now...

afck:
I realized that we don't actually need that for `DynamicHoneyBadger`: It can just restart `HoneyBadger` whenever a new node is voted in _and_ then again when DKG has finished. That effectively sets `max_future_epochs` to `0` only in the epochs where we need it.

And I think we actually shouldn't set it to `0` for the others, since it's an optimization. Restarting `HoneyBadger` is costly, too, but it happens only when the set of validators changes.

I'm still happy with merging this PR: I think we'll need the builder pattern anyway, and thanks to it, the additional configuration options doesn't add much clutter. We should probably add it as an option to `DynamicHoneyBadger`, too.

afck:
I rebased it. Please take another look, @vkomenda. I'm still in favor of merging this; I think we'll want a builder for Honey Badger anyway, no matter how the set of parameters will change.

PR #97, opened 2018-06-28
Add lifetime to `HoneyBadgerBuilder`.

c0gent:


afck:
Thanks, that makes sense!
(It's not a lifetime, though.)

c0gent:
Hah, yeah I meant type.

Issue #72, opened 2018-06-21
Consistently name individual algorithms.


afck:
We started referring to `CommonSubset` as `Subset` in some places, and should make sure the names are the same everywhere. Everything should follow the naming convention described in the Readme.

afck:
Unassigning this for now, until more names have reached consensus.

PR #101, opened 2018-07-02
Add module-level Agreement, Coin and Subset documentation.

### LINKED ISSUE #100: Add more detailed module-level documentation for all algorithms.
Every module should have an extensive description of what it does, and maybe even a short explanation of how it does it. (That explanation would be needed in the code anyway, to make it readable at all, so why not make it part of the module docs themselves.)

Once the API is stable (#66), we should add a usage example to every module, too.

afck:
Also change the signature parity computation, such that every bit is taken into account instead of only the last bit of each byte.

Closes #100

### DIFF
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -45,7 +45,7 @@
 //!   * If both values are candidates, we set `e = s` and proceed to the next epoch.
 //!
 //! In epochs that are 0 modulo 3, the value `s` is `true`. In 1 modulo 3, it is `false`. In the
-//! case 2 modulo 3, we carefully flip a common coin to determine a pseudorandom `s`.
+//! case 2 modulo 3, we flip a common coin to determine a pseudorandom `s`.
 //!
 //! An adversary that knows each coin value, controls a few validators and controls network
 //! scheduling can delay the delivery of `Aux` and `BVal` messages to influence which candidate
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -1,4 +1,67 @@
-//! Binary Byzantine agreement protocol from a common coin protocol.
+//! # Binary Byzantine agreement protocol
+//!
+//! The Binary Agreement protocol allows each node to input one binary (`bool`) value, and will
+//! output a binary value. The output is guaranteed to have been input by at least one correct
+//! node, and all correct nodes will have the same output.
+//!
+//! ## How it works
+//!
+//! The algorithm proceeds in _epochs_, and the number of epochs it takes until it terminates is
+//! unbounded in theory but has a finite expected value. Each node keeps track of an _estimate_
+//! value `e`, which is initialized to the node's own input. Let's call a value `v`
+//! that has been input by at least one correct node and such that `!v` hasn't been _output_ by any
+//! correct node yet, a _viable output_. The estimate will always be a viable output.
+//!
+//! All messages are annotated with the epoch they belong to, but we omit that here for brevity.
+//!
+//! * At the beginning of each epoch, we multicast `BVal(e)`. It translates to: "I know that `e` is
+//! a viable output."
+//!
+//! * Once we receive `BVal(v)` with the same value from _f + 1_ different validators, we know that
+//!   at least one of them must be correct. So we know that `v` is a viable output. If we haven't
+//!   done so already we multicast `BVal(v)`. (Even if we already multicast `BVal(!v)`).
+//!
+//! * Let's say a node _believes in `v`_ if it received `BVal(v)` from _2 f + 1_ validators.
+//! For the _first_ value `v` we believe in, we multicast `Aux(v)`. It translates to:
+//! "I know that all correct nodes will eventually know that `v` is a viable output.
+//! I'm not sure about `!v` yet."
+//!
+//!   * Since every node will receive at least _2 f + 1_ `BVal` messages from correct validators,
+//!   there is at least one value `v`, such that every node receives _f + 1_ `BVal(v)` messages.
+//!   As a consequence, every correct validator will multicast `BVal(v)` itself. Hence we are
+//!   guaranteed to receive _2 f + 1_ `BVal(v)` messages.
+//!   In short: If _any_ correct node believes in `v`, _every_ correct node will.
+//!
+//!   * Every correct node will eventually send exactly one `Aux`, so we will receive at least
+//!   _2 f + 1_ `Aux` messages with values we believe in. At that point, we define the set `vals`
+//!   of _candidate values_: the set of values we believe in _and_ have received in an `Aux`.
+//!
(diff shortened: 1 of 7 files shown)


afck:
The `agreement` documentation glosses over several important details, e.g.:

* It pretends that the epochs happen strictly one after another, but in an asynchronous context, nodes can be at very different stages at the same time. That also means that formulations like "hasn't been output yet" strictly mean "hasn't been output in any epoch until this one".
* All messages are multicast, but by "receive" I always also mean the message that we've sent ourselves. We always handle our own messages implicitly.
* A lot of the edge cases regarding the coin, and in particular the complicated attack that led to the `Conf` messages.

But I guess the docs are not supposed to replace the paper, and they're already complicated enough as they are…

I also replaced "`2 * f + 1`" with "_2 f + 1_"; I'm not sure about it, but that's at least closer to what L<sup>A</sup>T<sub>E</sub>X would do.

src/common_subset.rs
```diff
@@ -1,4 +1,26 @@
-//! Asynchronous Common Subset algorithm.
+//! # Asynchronous Common Subset algorithm.
+//!
+//! The Asynchronous Common Subset protocol assumes a network of _N_ nodes that send signed
+//! messages to each other, with at most _f_ of them malicious, where _3 f < N_. Handling the
+//! networking and signing is the responsibility of the user: only when a message has been
+//! verified to be "from node i", it can be handed to the `CommonSubset` instance.
```
andogro: We discussed dropping Common and renaming to simply Subset. If too vague, can keep as CommonSubset and update readme. Same for Common Coin renamed to Coin.

src/common_coin.rs
```diff
@@ -1,4 +1,26 @@
...
+// TODO: Bold claims! How much can we actually prove about that?
+//!
+//! The nodes input a signal (no data, just `()`), and after _2 f + 1_ nodes have provided input,
+//! everyone receives the output value. In particular, the adversary cannot know the output value
+//! before at least one correct node has provided input.
+//!
+//! ## How it works
+//!
+//! The algorithm uses a threshold signature scheme with the uniqueness property: For each public
+//! key and message, there is exactly one valid signature. This signature can be produced using
```
andogro: This group signature is produced using signature shares from any combination of _2 f + 1_ secret key share holders.

src/common_subset.rs
andogro: Do we need to mention how it is verified to be from node i?

src/common_subset.rs
```diff
@@ -1,4 +1,26 @@
...
+//!
+//! Each participating node proposes an element for inclusion. Under the above conditions, the
+//! protocol guarantees that all of the correct nodes output the same set, consisting of at least
```
andogro: ~~participating~~ .  Not sure if needed because of dynamic HB?

...protocol guarantees that all ~~ of the~~ correct nodes ...

src/common_subset.rs
```diff
@@ -1,4 +1,26 @@
...
+//! _N - f_ of the proposed elements.
+//!
+//! ## How it works
+//!
+//! * `CommonSubset` instantiates one `Broadcast` algorithm for each of the participating nodes.
+//! At least _N - f_ of these - the ones whose proposer is not faulty - will eventually output
+//! the element proposed by that node.
+//! * It also instantiates an `Agreement` instance for each participating node, to decide whether
```
andogro: Binary Agreement?

src/common_subset.rs
afck: Yes, and we also need to rename all the identifiers in the code (https://github.com/poanetwork/hbbft/issues/72). But maybe it's better to postpone this a few more days, otherwise @vkomenda's API redesign will have an even bigger merge conflict.
Let's do the renaming after that, in a separate PR, for all the code and docs at once.

afck:
I rebased it, and all comments should be addressed. Please take another look.

src/common_coin.rs
```diff
@@ -1,4 +1,26 @@
-//! Common coin from a given set of keys based on a `pairing` threshold signature scheme.
+//! # A Cryptographic Common Coin
+//!
+//! The Common Coin produces a pseudorandom binary value that the correct nodes agree on, and that\
```
vkomenda: Do you need the backslash?

src/common_coin.rs
```diff
@@ -1,4 +1,26 @@
...
+//! cannot be known beforehand.
+//!
+//! Every Common Coin instance has a _nonce_ that determines the value, without giving it away: It
+//! is not feasible to compute the output from the nonce alone, and for any "natural" sequence of
+//! nonces (that can be feasibly computed), the output is uniformly distributed.
+// TODO: Bold claims! How much can we actually prove about that?
```
vkomenda: A proof would require determining wether the expected number of zeros or ones in a signature is 50%. In the absence of such proof, can we refer to `tests/common_coin.rs` for that? If you flip the common coin like a 1000 times, you should get pretty much 50% of each of `true` and `false`. There is no test with 1000 flips because of time constraints.

src/agreement/mod.rs
```diff
@@ -1,4 +1,67 @@
...
+//! * Once we have the set of candidate values, we obtain a _coin value_ `s` (see below).
+//!
+//!   * If there is only a single candidate value `b`, we set our estimate `e = b`. If `s == b`,
+//!   we _output_ and send a `Term(b)` message which is interpreted as `BVal(b)` and `Aux(b)` for
+//!   all future epochs. If `s != b`, we just proceed to the next epoch.
+//!
+//!   * If both values are candidates, we set `e = s` and proceed to the next epoch.
+//!
+//! In epochs that are 0 modulo 3, the value `s` is `true`. In 1 modulo 3, it is `false`. In the
+//! case 2 modulo 3, we carefully flip a common coin to determine a pseudorandom `s`.
```
vkomenda: Maybe simply "we flip a common coin"? There is only one way of flipping it. I'm sure though we've done our best to make it happen in the best possible way!

src/common_coin.rs
afck: Yes, it's just that that's not strictly a proof. What I'm kind of claiming here is that there is no way to produce an efficiently computable sequence of nonces whose output isn't uniformly distributed. All we need is just that sequences like `"nonce 1", "nonce 2", "nonce 3", ...` will work fine. Not sure what's the best formulation.

src/agreement/mod.rs
afck: You're right. I was thinking of the `Conf` message, which kind of makes the coin flip more careful: It's careless to give your signature share to the adversary too early. :grin:

src/common_coin.rs
afck: I changed it to a much simpler formulation and don't mention sequences at all. I think the meaning is still pretty clear this way.

Issue #66, opened 2018-06-19
Design a more ergonomic API.


afck:
The current `DistAlgorithm` trait makes it easy for the caller to forget checking `next_message` and `next_output`. It would probably be safer to go back to a design where `handle_message` directly returns a (`#[must_use]`?) result that contains the messages and output.

We should also consider moving `DistAlgorithm` into the tests. Even if we don't, all the algorithms should at least be usable without it, too.

vkomenda:
Here is the way I think we should change the API. Please feel free to comment. I propose to keep the `DistAlgorithm` interface in a modified form for internal purposes, making it more automaton-like. At the moment `DistAlgorithm` already has semantics close to that of an automaton - with `Input` and `Output` but without state. I believe that the interface should only have

1. a step function from inputs (and states, internally) to outputs and states,

2. types of input value, output value and state.

An _input_ is an incoming message and an optional input value.

An _output_ is a list of outgoing messages, a list of log entries and an optional output value. The caller `#[must_use]` the output.

The usual type of state will be the set of four values:

a. not initialized (accepting incoming messages and input values),

b. initialized (accepting incoming messages and discarding input values),

c. terminated (discarding everything),

d. error.

The state value is a field in the algorithm `struct` and is output at each `step`, so there is no need to have an `is_terminated` function to check for termination.

I think that keeping errors separate from state as in `Result<OutputAndState, Error>` would only follow the requirements of external crates such as `error_chain` and is not a necessity.

That was the easy part. So, what does it mean for the user of the library? If you use any of the algorithm `struct`s directly, you must know the types of its input values, output values, messages and states. The top level API can provide an interface call to the `step` function, so that using the `DistAlgorithm` trait is not needed in user code.

afck:
I agree with the plan for the outputs, and I like the idea of replacing multiple booleans (like `terminated`) with a single state `enum`, where possible.

> An _input_ is an incoming message and an optional input value.

I think we would end up only ever calling `input(None, Some(input))` or `input(Some(msg), None)`, so maybe it should be two separate functions.

afck:
Let's also keep in mind the [Rust API guidelines](https://rust-lang-nursery.github.io/api-guidelines/) whenever we make changes to the API.

vkomenda:
I created an issue for that: #108.

PR #106, opened 2018-07-04
Output NetworkInfo from Dynamic Honey Badger

### LINKED ISSUE #102: Dynamic Honey Badger should produce proofs for new public keys.
When the set of validators changes, we create a new set of secret key shares, and a new public key. Users will probably need a proof to trust the new key, i.e. a signature by the old key.

Is that easy to do externally, or should this just be done by default as part of `DynamicHoneyBadger`? After the block in which key generation has finished, should we create signature shares for the new block, commit them via the old `HoneyBadger` instance and only restart with the new validator set once enough of _those_ have been output (i.e. sign the new public key on-chain)? Or should we add an additional message round for exchanging the signature shares, and make the result part of the `dynamic_honey_badger::Batch`?

afck:
This allows the user to use current cryptographic keys and key shares.

An example use case would be if we want to make the sequence of batches
verifiable for third parties, like a blockchain. We would have each
batch signed by the set of validators that produced it, and whenever a
validator changes, sign the new public key using the old key shares.

To a third party that would then look like the owner of the secret key
signed each batch, and occasionally changed their key.

Closes #102

### DIFF
--- a/src/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger.rs
@@ -206,7 +206,7 @@ where
     /// The messages that need to be sent to other nodes.
     messages: MessageQueue<NodeUid>,
     /// The outputs from completed epochs.
-    output: VecDeque<Batch<Tx, NodeUid>>,
+    output: VecDeque<(Batch<Tx, NodeUid>, NetworkInfo<NodeUid>)>,
 }
 
 impl<Tx, NodeUid> DistAlgorithm for DynamicHoneyBadger<Tx, NodeUid>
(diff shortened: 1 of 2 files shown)


c0gent:
It seems like `NetworkInfo` may be pulling double duty here. Perhaps create a separate, `PublicNetworkInfo` type that is similar but contains no secret key?

afck:
That's true, that would make sense anyway, and then we could make the `PublicNetworkInfo` part of the `Batch` itself. But what about the private one? That needs to be passed back to the caller after a validator change, too. I guess we'd still have a tuple then?

```rust
type Output = (Batch<Tx, NodeUid>, SecretKey);
```

c0gent:
You could store it as a struct with three fields then assemble it into a tuple or a two-field struct when needed. I'm not completely clear on the nuances of how this is used though so take my advice with a grain of salt.

afck:
> I'm not completely clear on the nuances of how this is used though
I'm not sure yet either. I'm happy to postpone this until we have a clearer picture of what we need from Parity's side.

afck:
I'll close this for now. Needs some [API design](https://github.com/poanetwork/hbbft/issues/66).

PR #107, opened 2018-07-04
Add recommended metadata to Cargo.toml

afck:
The Rust API guidelines recommend a set of metadata values:
https://rust-lang-nursery.github.io/api-guidelines/documentation.html#c-metadata
(See #108)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 name = "hbbft"
 version = "0.1.0"
-authors = ["Vladimir Komendantskiy <komendantsky@gmail.com>"]
+authors = ["Vladimir Komendantskiy <komendantsky@gmail.com>",
+           "Andreas Fackler <AndreasFackler@gmx.de>",
+           "Peter van Nostrand <jnz@riseup.net>",
+           "Andrew Gross <andogro@gmail.com>",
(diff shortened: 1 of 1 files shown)


afck:
There's also an optional "`homepage`" link; not sure whether it would make sense to link to e.g. https://poa.network/ from the crates.io page, though.

igorbarinov:
I think it's better to link home page to GitHub org page

afck:
OK, then we don't have to add the "`homepage`" field, as it links to the "`repository`" one by default anyway.

Cargo.toml
```diff
@@ -2,6 +2,15 @@
 name = "hbbft"
 version = "0.1.0"
 authors = ["Vladimir Komendantskiy <komendantsky@gmail.com>"]
```
vkomenda: Let's update `authors` too.

afck: Should we add all of us individually, or rather a single POA email address?

vkomenda: You are right, as the team grows, we might need to follow the example of other projects, e.g.,
```
authors = ["Parity Technologies <admin@parity.io>"]
```

afck: @igorbarinov: Is there some email address like that, which we should use as "the author"?

afck: I extended the authors list.

PR #109, opened 2018-07-04
Implement `Rand` for `SecretKey`.

mbr:
I am trying to cut the noise of `let rng = rand::thread_rng(); Foo::random(&mut rng)`, to make tests easier to read and reduce the coupling to a specific RNG, however the API is not well suited to instantiate types that need additional parameters. At least `SecretKey` can (and should) implemented `Rand` here.

Caveats:

* `ThreadRng` in newer versions is marked as cryptographically secure through the `CryptoRng` trait, I do not see any issues in using it. It leverages the OS random generator (which on Linux collects entropy from various sources and adds them to a pool, possibly including real hardware random number generators) and falls back on a generator based on CPU jitter. Said RNG errs on the side of entropy, so it should not be less random, it is simply slower by three to six orders of magnitude.
* The `Rand` trait is deprecated, however the `rand` crate is due another overhaul in the near future it seems. We need to pick one version and implement against that for now and it seems we are stuck on the 0.4 branch, since pairing will not compile with 0.5.
* I briefly considered (and implemented, but left out of the PR) `generate` functions that essentially do the little `thread_rng`-instantiation-dance, but I think they are too much code added for too few gains.

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -16,7 +16,7 @@ use clear_on_drop::ClearOnDrop;
 use init_with::InitWith;
 use pairing::bls12_381::{Bls12, Fr, FrRepr, G1, G1Affine, G2, G2Affine};
 use pairing::{CurveAffine, CurveProjective, Engine, Field, PrimeField};
-use rand::{ChaChaRng, OsRng, Rng, SeedableRng};
+use rand::{ChaChaRng, OsRng, Rand, Rng, SeedableRng};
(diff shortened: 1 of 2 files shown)


vkomenda:
The PR is good. Please correct the formatting.

mbr:
Apparently, stable Rust's rustfmt and nightly Rust's rustfmt have different opinions on how chained function calls should be formatted. CI should pass now, if I got the version right.

PR #111, opened 2018-07-06
Add documentation and screenshot for network simulation

andogro:
Some of the wording may need adjustment in the Heading/Definition section.  I also added info around Issues in the Contributing file and deleted the TODO file.


README.md
```diff
@@ -64,9 +64,54 @@ $ cargo test --release
...
 ```
+$ cargo run --example simulation --release
+```
+
+![Screenshot](screenshot.png)
+
+|  Heading    | Definition                       
+| ----------- | -------------------------------------------------------------------------- | 
+| Epoch       |  Epoch number. In each epoch, transactions are processed in a batch by simulated nodes (default is 10 nodes) on a network.                                                            | 
+| Min Time   | Timestamp on the _first_ Tx (transaction) processed in a batch of transactions.                                         |
```
andogro: Unsure about wording here around min and max time.

igorbarinov: Timestamp in sec/ms?

CONTRIBUTING.md
```diff
@@ -8,18 +8,22 @@ Thank your for contributing to this project! We welcome collaborators and expect
...
 ### General
 
+* We strive to follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/about.html) to maintain consistency and compatibility in our code.     
 * Commits should be one logical change that still allows all tests to pass.  We prefer smaller commits if there could be two levels of logic grouping.  The goal is to provide future contributors (including your future self) the reasoning behind your changes and allow them to cherry-pick, patch or port those changes in isolation to other branches or forks.
 * If during your PR you reveal a pre-existing bug:
   1. Try to isolate the bug and fix it on an independent branch and PR it first.
-  2. Try to fix the bug in a separate commit from other changes:
-     1. Commit the code in the broken state that revealed the bug originally
+  2. Try to fix the bug in a separate commit from other changes.
+     1. Commit the code in the broken state that revealed the bug originally.
```
afck: I think the most common case is really that someone finds a bug and doesn't know how to fix it. Which is fine, as documented below. But maybe we should add "you reveal a pre-existing bug and know how to fix it" to line 17?

And then maybe I'd just make two points in total:

```markdown
1. If you can isolate the bug, fix it in a separate PR.
2. If it depends on your other commits, add the fix in a separate commit to the same PR.

In either case, try to write a regression test that would fail because of the bug, but passes with your fix.
```

CONTRIBUTING.md
```diff
@@ -8,18 +8,22 @@ Thank your for contributing to this project! We welcome collaborators and expect
...
      2. Commit the fix for the bug.
      3. Continue original PR work.
 
+### Issues
+Creating and discussing [Issues](https://github.com/poanetwork/hbbft/issues) provides significant value to the project. If you find a bug but do not want to address it in a PR, you can report it as an Issue.
```
afck: Maybe just "`If you find a bug, you can report it in an Issue.`"? It's usually not because someone doesn't _want_ to fix it themselves.

README.md
afck: Yes, it's in (simulated) milliseconds.

The min time is the time until the first  (correct, i.e. not faulty) _node_ outputs the batch. The max time is the time until the last one outputs the batch. The batch is always output in one piece, with all transactions at once.

README.md
```diff
@@ -64,9 +64,54 @@ $ cargo test --release
...
+
+|  Heading    | Definition                       
+| ----------- | -------------------------------------------------------------------------- | 
+| Epoch       |  Epoch number. In each epoch, transactions are processed in a batch by simulated nodes (default is 10 nodes) on a network.                                                            | 
+| Min Time   | Timestamp on the _first_ Tx (transaction) processed in a batch of transactions.                                         |  
+| Max Time   | Timestamp on the _last_ Tx in a batch.                                                  |  
+| Txs         | Number of transactions processed in the epoch.                                           |  
+| Msgs/Node   | Average number of messages handled by a node. The counter is cumulative and includes the number of messages handled in the current epoch and all previous epochs.                                                               |  
+| Size/Node   | Average message size (in kilobytes) handled by a node. This is cumulative and includes message size for the current epoch and all previous epochs.                                                             |
```
afck: It uses "MB" (or "GB"), too, if the numbers get large enough, so it's not always in kilobytes.

CONTRIBUTING.md
```diff
@@ -14,15 +14,15 @@ Thank your for contributing to this project! We welcome collaborators and expect
...
 * We strive to follow the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/about.html) to maintain consistency and compatibility in our code.     
 * Commits should be one logical change that still allows all tests to pass.  We prefer smaller commits if there could be two levels of logic grouping.  The goal is to provide future contributors (including your future self) the reasoning behind your changes and allow them to cherry-pick, patch or port those changes in isolation to other branches or forks.
-* If during your PR you reveal a pre-existing bug:
-  1. Try to isolate the bug and fix it on an independent branch and PR it first.
-  2. Try to fix the bug in a separate commit from other changes.
-     1. Commit the code in the broken state that revealed the bug originally.
-     2. Commit the fix for the bug.
-     3. Continue original PR work.
+* If during your PR you reveal a pre-existing bug and know how to fix:
+  1. If you can isolate the bug, fix in a separate PR.
```
vkomenda: "to fix" →"to fix it"
"fix in" →"fix it in"

CONTRIBUTING.md
```diff
@@ -14,15 +14,15 @@ Thank your for contributing to this project! We welcome collaborators and expect
...
+  2. If the fix depends on your other commits, add the fix in a separate commit to the same PR.
```
afck: Maybe turn one "`the fix`" into "`it`" to avoid repetition.

PR #117, opened 2018-07-10
Separate queue from Honey Badger.

afck:
This makes Honey Badger a bit more complicated but a lot more flexible:
It is now unaware of transactions and basically just runs one Subset
instance per epoch.

That way, users can use any kind of external queue, control throttling
and prioritization.

### DIFF
--- a/src/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger.rs
@@ -56,7 +56,7 @@ use clear_on_drop::ClearOnDrop;
 use serde::{Deserialize, Serialize};
 
 use crypto::{PublicKey, PublicKeySet, SecretKey, Signature};
-use fault_log::FaultLog;
+use fault_log::{FaultKind, FaultLog};
 use honey_badger::{self, HoneyBadger, Message as HbMessage};
 use messaging::{DistAlgorithm, NetworkInfo, Target, TargetedMessage};
 use sync_key_gen::{Accept, Propose, ProposeOutcome, SyncKeyGen};
@@ -349,6 +349,8 @@ where
                     }
                     if !self.verify_signature(&s_id, &sig, &node_tx)? {
                         info!("Invalid signature from {:?} for: {:?}.", s_id, node_tx);
+                        let fault_kind = FaultKind::InvalidNodeTransactionSignature;
+                        fault_log.append(s_id.clone(), fault_kind);
                         continue;
                     }
                     use self::NodeTransaction::*;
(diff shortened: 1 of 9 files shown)


c0gent:
Yes, that looks very nice. I'll test it out today.

tests/queueing_honey_badger.rs
```diff
@@ -0,0 +1,150 @@
+//! Network tests for Queueing Honey Badger.
+
+extern crate hbbft;
+#[macro_use]
+extern crate log;
+extern crate env_logger;
```
DrPeterVanNostrand: Import order.

tests/queueing_honey_badger.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::cmp;
+use std::collections::BTreeMap;
+use std::iter::once;
+use std::rc::Rc;
+
+use rand::Rng;
+
+use hbbft::messaging::NetworkInfo;
+use hbbft::queueing_honey_badger::{Change, ChangeState, Input, QueueingHoneyBadger};
+
```
DrPeterVanNostrand: Blank line in imports.

src/dynamic_honey_badger.rs
```diff
@@ -338,34 +340,26 @@ where
...
-                            }
-                        }
+            for (id, int_contrib) in hb_batch.contributions {
+                batch.contributions.insert(id, int_contrib.contrib);
+                for SignedTransaction(epoch, s_id, node_tx, sig) in int_contrib.transactions {
+                    if epoch < self.start_epoch {
+                        info!("Obsolete node transaction: {:?}.", node_tx);
+                        continue;
+                    }
+                    if !self.verify_signature(&s_id, &sig, &node_tx)? {
```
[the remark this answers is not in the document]
afck: Good catch! I missed that while resolving the giant merge conflict. :sweat_smile:

src/queueing_honey_badger.rs
```diff
@@ -0,0 +1,200 @@
...
+use std::fmt::Debug;
+use std::hash::Hash;
+use std::marker::PhantomData;
+
+use serde::{Deserialize, Serialize};
+
+use dynamic_honey_badger::{self, Batch as DhbBatch, DynamicHoneyBadger, Message};
+use fault_log::FaultLog;
+use messaging::{DistAlgorithm, NetworkInfo, TargetedMessage};
+
```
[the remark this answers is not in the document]
afck: I moved the `transaction_queue` up, but maybe the `pub use` should be separate?

tests/queueing_honey_badger.rs
afck: I removed line 20 instead: I guess from the point of view of the tests, `hbbft` is an "external crate"?

PR #119, opened 2018-07-10
Add `.rustfmt.toml` and improve formatting.

c0gent:
This gets us much closer to what is fairly commonplace Rust
formatting. I'm not sure why the rustfmt defaults are set the way
they are, this is a huge improvement. I still really dislike how it
indents vertical lists of arguments so far to the right but disabling
that option makes the bracket style far to ugly and hard to read. One
day maybe I'll look into the rustfmt source and see if I can add an
option to reduce maximum indents to +2 or something.

If we decide to use this configuration (which we will because we are
sane and intelligent individuals), authors of PRs can just copy .rustfmt.toml
into their branches and run cargo fmt (to prevent death by merge conflict).


.rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+indent_style = "Visual"
```
afck: I like `"Block"` much better…

.rustfmt.toml
```diff
@@ -0,0 +1,6 @@
...
+brace_style = "PreferSameLine"
+where_single_line = true
+format_strings = true
```
afck: Not sure about this. Then we'd need to explicitly skip the `docopt` strings.

mbr:
My 2 cents: Never changing the defaults on formatters is a Dogma that I can get behind. I care very little how the code looks, it should just look the same everywhere, with minimal effort. `gofmt` gets this right. Or put differently, if `rustfmt` did not have any customization options, we would not even have this conversation =)

Feel free to discard my vote as bikeshedding though.

.rustfmt.toml
DrPeterVanNostrand: I do too. I think that the visual indentation makes the code harder to read more often than it makes it easier.

c0gent:
I agree this is bikeshedding to a degree. I'll say my piece and then happily accept whatever the team decides. 

I've been a go programmer since before release. Gofmt was a great tool, endless discussions and posts like this one were precluded because the defaults had already been decided. This video clip sums it up nicely (30 sec):

[Go Proverbs - Rob Pike - Gopherfest - November 18, 2015](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=8m43s)

Unfortunately, the situation in Rust is different for two reasons. 1.) There was no Rust formatting tool upon release and 2.) The tool that we do have (rustfmt) has bizarre and non-standard defaults that no one uses.

One look at the standard library ([HashMap](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#427)), futures-rs ([1](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-executor/src/spawn.rs#L103), [2](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-executor/src/spawn.rs#L26)), or any of a thousand other official or unofficial libraries and you will see a clear 'conventional' formatting style. Pay close attention to the `where` clauses. I've been programming in Rust for nearly five years and virtually every single project I've ever worked on or seen code for has followed those conventions. It is elegant, efficient (not a ton of wasted lines) and easy to orient yourself when scanning code quickly.

Our current rustfmt settings are unconventional and will make following our code more difficult for other Rust developers. In the end, our decision should be based on what formatting style is better for the future of the project, not our personal opinions.

PR #124, opened 2018-07-11
Extract vote counter from DHB.

### LINKED ISSUE #122: Fix Dynamic Honey Badger messaging logic.
This is currently a bit fuzzy, vulnerable to replaying votes, and doesn't always guarantee that votes eventually get committed. Also, if the joining node only joins after the epoch where the vote for adding it reaches a majority, it won't know the exact vote count and won't be able to detect if it loses majority again.
A solution could look as follows.

We clear the vote counts, restart Honey Badger and set `start_epoch` to the current epoch every time one of the following conditions occurs:
* A change reaches majority: That change remains the "current" one until another change gets a majority of votes that are committed _after_ this epoch. Votes committed in older epochs are dismissed.
* Key generation for the current change completes.

Every key generation message (`Accept` or `Prepare`) is annotated with that `start_epoch` (that's already the case). Also, every change vote message is annotated with the `start_epoch` and in addition with an incrementing `vote_num`ber: That way, nodes can change their votes and old votes can't be replayed. For each validator, the vote that counts (if any) is the vote with the highest `vote_num` that has been committed in an epoch `>= start_epoch`.

For a vote, the message flow is:
* The validator that wants to cast the votes signs the desired change (`Add` or `Remove`) with the `start_epoch` and incremented `vote_num` and sends it to all the other validators.
* When we receive a vote, we put it into `pending_votes: BTreeMap<NodeUid, Vote>` _if_ it has the current `start_epoch`, and has a greater `vote_num` that the existing entry by the same signatory (or if there's no entry yet).
* In each new epoch, we propose all `pending_votes` that are not in `votes` yet.
* After each epoch's output, we put all committed votes into `votes`  _if_ they have the current `start_epoch` and they have a grater `vote_num` than the existing entry by the same signatory (if any). If there is a majority in `votes`, we set `current_change` to that vote, restart Honey Badger and clear `votes` and `pending_votes`. Key generation starts for the new change.

One drawback still remains: the joining node can't verify that it has won the vote and is about to be added. That can't hurt the network, but it's probably still important for the new node to _know_ whether it is actually expected. Also, it should know the current `start_epoch` so that it knows that it has received all votes, and that it will realize if another change gets a majority.
* We could leave it as is: Then it's the user's responsibility to get that information to the new node and verify it.
* We could add the `current_change` and `start_epoch` to each proposal. That is a bit wasteful, but then the new node would automatically receive it.
* We could add an additional `Restart(start_epoch, current_change)` message type, probably for new observers in general. Every time Honey Badger is restarted, that is sent to `Target::All`. The new node needs _f + 1_ `Restart` messages to trust them.

afck:
Closes #122.


src/dynamic_honey_badger/votes.rs
```diff
@@ -0,0 +1,302 @@
+use messaging::NetworkInfo;
```
DrPeterVanNostrand: Import order.

src/dynamic_honey_badger/votes.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn add_pending_vote(
+        &mut self,
+        sender_id: &NodeUid,
+        signed_vote: SignedVote<NodeUid>,
+    ) -> Result<FaultLog<NodeUid>> {
+        if signed_vote.vote.era != self.era {
+            return Ok(FaultLog::new()); // The vote is obsolete.
+        }
+        if !self.validate(&signed_vote)? {
+            return Ok(Fault::new(sender_id.clone(), FaultKind::InvalidVoteSignature).into());
```
DrPeterVanNostrand: You could remove the call to `.into()` by using `Ok(FaultLog::init(sender_id.clone(), FaultKind::InvalidVoteSignature))`.

DrPeterVanNostrand: Same thing goes for everywhere else this is used. Both options do exactly the same thing though.

src/dynamic_honey_badger/votes.rs
```diff
@@ -0,0 +1,302 @@
...
+        // Create the counter instances and the matrix of signed votes.
+        let (mut counters, sv) = setup(node_num, era);
+        // We will only use counter number 0.
+        let ct = &mut counters[0];
+
+        // Node 0 already contains its own vote for `Remove(3)`. Add two more.
+        let faults = ct
+            .add_pending_vote(&1, sv[1][2].clone())
+            .expect("add pending");
+        assert_eq!(faults, FaultLog::new());
```
DrPeterVanNostrand: Maybe we should add an `is_empty()` method onto `FaultLog` so you can do `assert!(fault_log.is_empty());`?
```rust
impl<NodeUid: Clone> FaultLog<NodeUid> {
    pub fn is_empty(&self) -> bool {
        self.0.is_empty()
    }
}
```

src/dynamic_honey_badger/votes.rs
```diff
@@ -0,0 +1,302 @@
...
+            .add_pending_vote(&2, sv[2][1].clone())
+            .expect("add pending");
+        assert_eq!(faults, FaultLog::new());
+        // Include a vote with a wrong signature.
+        let fake_vote = SignedVote {
+            sig: sv[2][1].sig.clone(),
+            ..sv[3][1].clone()
+        };
+        let faults = ct.add_pending_vote(&1, fake_vote).expect("add pending");
+        let expected_fault = Fault::new(1, FaultKind::InvalidVoteSignature);
```
DrPeterVanNostrand: `FaultLog::init()` could be used here to remove the `into()`.

src/dynamic_honey_badger/votes.rs
DrPeterVanNostrand: Same goes for all other `assert!(..., FaultLog::new());`.

src/dynamic_honey_badger/votes.rs
DrPeterVanNostrand: Same goes for the `assert_eq!()` in `test_committed_votes()`.

Issue #89, opened 2018-06-26
Make sure secret keys are always inside a `ClearOnDrop`.


afck:
Overwriting the memory containing a secret key (#57) can still easily be forgotten: e.g. `SecretKey::new` returns a key outside of a `ClearOnDrop`.

Maybe the current `SecretKey` should be turned into a private `SecretKeyInner`, and put a `ClearOnDrop<Box<SecretKeyInner>>` inside the new `SecretKey`, so that outside of the `crypto` module there's no way to circumvent it anymore?

DrPeterVanNostrand:
I don't think that the `secrets` crate takes the same precautions to circumvent the rustc optimization of eliding zeroing out secret data if that memory won't be used again; `clear_on_drop` takes special care in this regard.  However, `secrets` does add a few extra security features like making sure secret values are not included in core dumps and the addition of stack canaries.

Also, the dalek cryptography crate designers chose to use `clear_on_drop` over `secrets`, I'm not sure why exactly, I'll have to read up on it. [Here](https://github.com/dalek-cryptography/curve25519-dalek/issues/11) is an issue where the dalek authors discuss the problem of clearing secrets from memory.

As far a linking to libsodium, I don't know if that is a requirement that I'm willing to put on the user. Personally, I hate it when I'm trying to use a library in a language other than C, but first I have to build and link all of my dependencies. But Idk, what do you (@afck, @mbr) think?

DrPeterVanNostrand:
> `unsafe { sodium::free(self.ptr) }`

After thinking about it, I don't think that the compiler is smart enough to elide a call to a C function. So elision of the memory clearing code may not be a problem with `secrets`.

afck:
Yes, I imagine that won't be optimized away, but I agree that the sodium dependency is annoying. Maybe `secrets` could be modified to depend on one of the sodium Rust crates instead, that vendor the C library?

Anyway, I don't have a strong opinion regarding `clear_on_drop` vs. `secrets` in general.

mbr:
I did not understand yet why we are not using an "established" crate that protects memory, I assume it was due to the libsodium dependency? Or was it just not generic enough? I would think that disabling swap is fairly important for keys.

There's also the tradeoff between implementing `Debug` and otherwise preventing accidental leakage. I did try to address this with another crate a while ago: https://github.com/49nord/sec-rs , it is probably more of a nice to have.

[...]

Issue #90, opened 2018-06-26
SyncKeyGen uses stack allocated SecretKeys


DrPeterVanNostrand:
1. `SyncKeyGen::new()` and `SyncKeyGen::generate()` both create secrets on the stack.
2. `SyncKeyGen` is instantiated using `NetworkInfo.secret_key().clone()` which copies the `SecretKey` from a `ClearOnDrop<Box<SecretKey>>` onto the stack then moves ownership of the `SecretKey` into an instance of `SyncKeyGen` (resulting in two copies).

PR #126, opened 2018-07-12
Replaced clear-on-drop with 'std::ptr::write_volatile'.

DrPeterVanNostrand:
- Removes the need to wrap `SecretKey`s in `ClearOnDrop<Box<...>>`.
- Allows for stack allocated `SecretKey`s.
- Zeros out the memory occupied by each `SecretKey` on drop.
- Ensures that the clearing of `SecretKey`s is not elided.
- Works on stable.

Closes Issues #89 and #90

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ travis-ci = { repository = "poanetwork/hbbft" }
 [dependencies]
 bincode = "1.0.0"
 byteorder = "1.2.3"
-clear_on_drop = "0.2.3"
 derive_deref = "1.0.1"
 env_logger = "0.5.10"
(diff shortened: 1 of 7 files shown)


c0gent:
This looks good and zeros properly in my tests. It may be better in the future, however, to use a more thorough memory security library such as [`memsec`](https://github.com/quininer/memsec)/[`seckey`](https://github.com/quininer/seckey). They will use a special allocator (like sodium) and do a few other things to make sure secure memory isn't swapped to disk, etc.

For now though this is a definite usability improvement over `ClearOnDrop`. Please merge or give me the ok to do so :)

PR #127, opened 2018-07-12
Minor fixes and simplifications.

afck:
* Clear outdated key gen messages from the buffer.
* Process output after proposing, to make `HoneyBadger` work with a
  single validator.
* Print an error if threshold decryption fails.
* Verify decryption shares with the correct ciphertext.
* Insert all ciphertexts from an epoch at once; otherwise contributions
  can be omitted from a batch.
* Remove `BoolWithFaultLog`: It's easier to return a tuple, and it's
  used only in one place now.
* Avoid redundant signature verification in `VoteCounter`.
* Fix the tests for `QueueingHoneyBadger`.
* Use fewer network sizes to speed up tests a bit.

### DIFF
--- a/src/queueing_honey_badger.rs
+++ b/src/queueing_honey_badger.rs
@@ -2,6 +2,7 @@
 //!
 //! This works exactly like Dynamic Honey Badger, but it has a transaction queue built in.
 
+use std::cmp;
 use std::collections::VecDeque;
 use std::fmt::Debug;
 use std::hash::Hash;
@@ -154,9 +155,7 @@ where
             self.queue.remove_all(batch.iter());
             self.output.push_back(batch);
         }
-        if !self.dyn_hb.has_input() {
-            self.propose()?.merge_into(&mut fault_log);
-        }
+        self.propose()?.merge_into(&mut fault_log);
         Ok(fault_log)
     }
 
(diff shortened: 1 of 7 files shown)


src/queueing_honey_badger.rs
```diff
@@ -154,9 +155,7 @@ where
             self.queue.remove_all(batch.iter());
             self.output.push_back(batch);
         }
-        if !self.dyn_hb.has_input() {
-            self.propose()?.merge_into(&mut fault_log);
-        }
+        self.propose()?.merge_into(&mut fault_log);
```
DrPeterVanNostrand: I think we are getting rid of `merge_into()` in exchange for `fault_log.extend(self.propose()?);`

afck: Yes, there are still plenty of `merge_into` calls. I'd prefer to replace them in a separate PR.
(I didn't change this line; only removed indentation, that's why I missed it).

DrPeterVanNostrand: Alright, sounds good.

vkomenda:
The branch looks OK. There are a few duplications with #130 which are likely to mean more conflicts when merging #130. I can merge #127  and then again rebase #130. Actually, this time I might squash the #130 branch. It's difficult to keep it up to date with people making concurrents PRs that work with the old API.

Issue #129, opened 2018-07-12
Tell the new node it's invited.


afck:
Currently a new observer can only join a `DynamicHoneyBadger` network after an epoch in which a `ChangeState::Complete` has been output, i.e. a node has just joined or left: That's when the internal Honey Badger restarts, key generation is terminated, the vote counts are reset and the "current change" is set to `None`. Joining at any other point would mean that its vote count is out of sync, it doesn't know what new validator set key generation is in progress for, it misses key generation messages, and it could even have missed Honey Badger messages (unless `max_future_epochs` is set to `0`).

However, a natural point for a new node `n` to join as an observer is after the epoch where `ChangeState::InProgress(Change::Add(n, _))` was output, i.e. where key generation for the new validator set including `n` is beginning. After `InProgress(c)`, Honey Badger restarts, and the vote counts are also cleared. However, the current change is set to `Some(c)`. What's still missing is a parameter `change`, so that the new node can initialize its `DynamicHoneyBadger` accordingly, and know what the current key generation process is about. In particular, it will know if it is the current candidate for joining as a validator, and that it's supposed to take part in key generation.

So the process of joining the network would be something like this:

The new node with ID `n` generates a random secret key `sk`, and the corresponding public key `pk`, and tells the network something like `IWannaJoin(n, pk)`, and the network decides whether to accept it. That part is application-level logic.
Once accepted, the validators input `Input::Change(Add(n, pk))` into `DynamicHoneyBadger`. Soon after that, an output batch in some epoch `e` will contain `ChangeState::InProgress(Add(n, pk))`. Starting with the messages from the method call that produced this output, all future `Target::All` messages must be sent to the new node. (Or cached and sent later, if not connected yet.)
Now the nodes send some kind of `Welcome(e + 1, pub_key_set, node_ids)` (application-level) message to the new node to inform it about the epoch in which it is joining as an observer, and about the existing node IDs (excluding `n`) and public keys (excluding `pk`) in the network.
The new node now creates a new Dynamic Honey Badger instance and sends its initial key generation messages:

```rust
let netinfo = NetworkInfo::new(n, node_ids, sk, pub_key_set);
let mut dhb = DynamicHoneyBadger::builder(netinfo)
    .start_epoch(e + 1)
    .change(Change::Add(n, pk))
    .build();
for msg in dhb.message_iter() {
    send_over_the_network(msg);
}
```

We need to:
* add the new parameter and, if present, initialize key generation on startup,
* document this,
[...]

Issue #120, opened 2018-07-11
Test fault logs.


afck:
The existing tests should verify that none of the correct nodes are ever reported as faulty.

In the long term, the other tests should verify that the faulty ones _are_ reported, whenever they do one of the detectable misdeeds.

afck:
Let's also `#[cfg(test)] panic!()` if a `Step` with a nonempty fault log is dropped, to make sure our code never loses a fault entry. The user, however, should probably not be forced to process them.

vkomenda:
#130 adds compiler attributes in each main test module to throw an error if the `Step` result is not used. This is however only a half-measure. `Step::fault_log` should be used from that result.

afck:
Some checks were added in #279: in `network`-based tests (but not the new `net`-based ones yet), we now verify that no correct node is reported as faulty. (But not whether faulty nodes are correctly reported.)

vkomenda:
I [added](https://github.com/poanetwork/hbbft/commit/15bb2488eb6d1e504ddd120cd052af3c7a4b72f0) the fault check in the `net` module. The reordering attack started failing on duplicate `BVal` messages. I'll investigate what's going on there.

afck:
I think the MITM test just needs to be fixed so that the network knows that that node belong to the adversary.

vkomenda:
It was my copy-paste mistake! The MITM test is fine.

PR #130, opened 2018-07-12
Output API refactored

### LINKED ISSUE #66: Design a more ergonomic API.

vkomenda:
This is a part of the refactoring undertaken to fix #66. It includes:

* the new algorithm output value API,

* integration with the earlier `FaultLog` functionality.

It does not include a new outgoing message API which will be included in a different PR considering the number of files involved in the refactoring and possible work on fixing tests.

### DIFF
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -74,8 +74,8 @@ use itertools::Itertools;
 
 use agreement::bin_values::BinValues;
 use common_coin;
-use fault_log::FaultLog;
 use common_coin::{CommonCoin, CommonCoinMessage, CommonCoinStep};
+use fault_log::FaultLog;
 use messaging::{DistAlgorithm, NetworkInfo, Step, Target, TargetedMessage};
 
 error_chain!{
@@ -197,8 +197,8 @@ impl<NodeUid: Clone + Debug + Ord> DistAlgorithm for Agreement<NodeUid> {
     type Error = Error;
 
     fn input(&mut self, input: Self::Input) -> AgreementResult<AgreementStep<NodeUid>> {
-        let fault_log = self.set_input(input);
-        self.step().with_fault_log(fault_log)
+        let fault_log = self.set_input(input)?;
+        self.step(fault_log)
     }
 
     /// Receive input from a remote node.
@@ -207,23 +207,23 @@ impl<NodeUid: Clone + Debug + Ord> DistAlgorithm for Agreement<NodeUid> {
         sender_id: &Self::NodeUid,
         message: Self::Message,
     ) -> AgreementResult<AgreementStep<NodeUid>> {
-        if self.terminated || message.epoch < self.epoch {
+        let fault_log = if self.terminated || message.epoch < self.epoch {
             // Message is obsolete: We are already in a later epoch or terminated.
-            return self.step();
-        }
-        if message.epoch > self.epoch {
+            FaultLog::default()
+        } else if message.epoch > self.epoch {
             // Message is for a later epoch. We can't handle that yet.
             self.incoming_queue.push((sender_id.clone(), message));
-            return self.step();
-        }
-        let fault_log = match message.content {
-            AgreementContent::BVal(b) => self.handle_bval(sender_id, b)?,
-            AgreementContent::Aux(b) => self.handle_aux(sender_id, b)?,
-            AgreementContent::Conf(v) => self.handle_conf(sender_id, v)?,
-            AgreementContent::Term(v) => self.handle_term(sender_id, v)?.map(|()| FaultLog::new()),
-            AgreementContent::Coin(msg) => self.handle_coin(sender_id, *msg)?,
+            FaultLog::default()
+        } else {
+            match message.content {
+                AgreementContent::BVal(b) => self.handle_bval(sender_id, b)?,
+                AgreementContent::Aux(b) => self.handle_aux(sender_id, b)?,
+                AgreementContent::Conf(v) => self.handle_conf(sender_id, v)?,
+                AgreementContent::Term(v) => self.handle_term(sender_id, v)?,
+                AgreementContent::Coin(msg) => self.handle_coin(sender_id, *msg)?,
+            }
         };
-        self.step().with_fault_log(fault_log)
+        self.step(fault_log)
     }
 
     /// Take the next Agreement message for multicast to all other nodes.
@@ -280,8 +280,11 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         }
     }
 
-    fn step(&mut self) -> AgreementResult<AgreementStep<NodeUid>> {
-        Ok(Step::new(self.output.take()))
+    fn step(&mut self, fault_log: FaultLog<NodeUid>) -> AgreementResult<AgreementStep<NodeUid>> {
+        Ok(Step::new(
+            self.output.take().into_iter().collect(),
+            fault_log,
+        ))
     }
 
     /// Sets the input value for agreement.
@@ -298,7 +301,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
             // Set the initial estimated value to the input value.
             self.estimated = Some(input);
             // Record the input value as sent.
-            self.send_bval(input);
+            self.send_bval(input)
         }
     }
 
@@ -450,7 +453,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
     /// Receives a `Term(v)` message. If we haven't yet decided on a value and there are more than
     /// `num_faulty` such messages with the same value from different nodes, performs expedite
     /// termination: decides on `v`, broadcasts `Term(v)` and terminates the instance.
-    fn handle_term(&mut self, sender_id: &NodeUid, b: bool) -> AgreementResult<()> {
+    fn handle_term(&mut self, sender_id: &NodeUid, b: bool) -> AgreementResult<FaultLog<NodeUid>> {
         self.received_term.insert(sender_id.clone(), b);
         // Check for the expedite termination condition.
         if self.decision.is_none()
@@ -459,7 +462,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         {
             self.decide(b);
         }
-        Ok(())
+        Ok(FaultLog::new())
     }
 
     /// Handles a Common Coin message. If there is output from Common Coin, starts the next
@@ -474,9 +477,13 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         self.on_coin_step(coin_step)
     }
 
-    fn on_coin_step(&mut self, coin_step: CommonCoinStep<NodeUid>) -> AgreementResult<FaultLog<NodeUid>> {
+    fn on_coin_step(
+        &mut self,
+        coin_step: CommonCoinStep<NodeUid>,
+    ) -> AgreementResult<FaultLog<NodeUid>> {
         let mut fault_log = FaultLog::new();
-        if let Some(coin) = coin_step.output {
+        fault_log.extend(coin_step.fault_log);
+        if let Some(coin) = coin_step.output.into_iter().next() {
             let def_bin_value = self.count_conf().1.definite();
             fault_log.extend(self.on_coin(coin, def_bin_value)?);
         }
@@ -515,7 +522,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
             let step = self.handle_message(&sender_id, msg)?;
             fault_log.extend(step.fault_log);
             // Save the output of the internal call.
-            self.output = step.output;
+            self.output = step.output.into_iter().next();
             if self.terminated {
                 break;
             }
@@ -567,20 +574,17 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
     }
 
     fn try_finish_conf_round(&mut self) -> AgreementResult<FaultLog<NodeUid>> {
-        let mut fault_log = FaultLog::new();
-        if self.conf_round {
-            let (count_vals, _) = self.count_conf();
-            if count_vals < self.netinfo.num_nodes() - self.netinfo.num_faulty() {
-                // Continue waiting for (N - f) `Conf` messages
-                return Ok(fault_log);
-            }
+        if self.conf_round
+            && self.count_conf().0 >= self.netinfo.num_nodes() - self.netinfo.num_faulty()
+        {
             // Invoke the comon coin.
             let coin_step = self.common_coin.input(())?;
-            fault_log.extend(coin_step.fault_log);
             self.extend_common_coin();
-            self.on_coin_step(coin_step)?;
+            self.on_coin_step(coin_step)
+        } else {
+            // Continue waiting for (N - f) `Conf` messages
+            Ok(FaultLog::default())
         }
-        Ok(fault_log)
     }
 
     fn send_aux(&mut self, b: bool) -> AgreementResult<FaultLog<NodeUid>> {
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -74,8 +74,8 @@ use itertools::Itertools;
 
 use agreement::bin_values::BinValues;
 use common_coin;
-use common_coin::{CommonCoin, CommonCoinMessage};
 use fault_log::FaultLog;
+use common_coin::{CommonCoin, CommonCoinMessage, CommonCoinStep};
 use messaging::{DistAlgorithm, NetworkInfo, Step, Target, TargetedMessage};
 
 error_chain!{
@@ -173,11 +173,9 @@ pub struct Agreement<NodeUid> {
     /// A cache for messages for future epochs that cannot be handled yet.
     // TODO: Find a better solution for this; defend against spam.
     incoming_queue: Vec<(NodeUid, AgreementMessage)>,
-    /// Termination flag. The Agreement instance doesn't terminate immediately
-    /// upon deciding on the agreed value. This is done in order to help other
-    /// nodes decide despite asynchrony of communication. Once the instance
-    /// determines that all the remote nodes have reached agreement, it sets the
-    /// `terminated` flag and accepts no more incoming messages.
+    /// Termination flag. Once the instance determines that all the remote nodes have reached
+    /// agreement or have the necessary information to reach agreement, it sets the `terminated`
+    /// flag and accepts no more incoming messages.
     terminated: bool,
     /// The outgoing message queue.
     messages: VecDeque<AgreementMessage>,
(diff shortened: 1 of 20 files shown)


vkomenda:
@c0gent - please review.

c0gent:
I've been working on bootstrapping a network with multiple remote nodes (using #129) and I haven't had a chance to look at this yet. I decided not to use my merged (#129 + #130) branch just yet in order to better isolate the issues. My simplified tests (using just #130), which aren't much different from the crate tests, work fine though.

Hopefully I can give this a more thorough review tomorrow. I'll just have to clear some of these huge piles of code off my plate...

src/agreement/mod.rs
```diff
@@ -210,39 +208,43 @@ pub struct Agreement<NodeUid> {
...
-    ) -> AgreementResult<FaultLog<NodeUid>> {
-        if self.terminated || message.epoch < self.epoch {
-            // Message is obsolete; we're already in a later epoch or terminated.
-            return Ok(FaultLog::new());
-        }
-        if message.epoch > self.epoch {
+    ) -> AgreementResult<AgreementStep<NodeUid>> {
+        let fault_log = if self.terminated || message.epoch < self.epoch {
+            // Message is obsolete: We are already in a later epoch or terminated.
+            FaultLog::default()
```
DrPeterVanNostrand: You could use `FaultLog::new()` here, which just calls `default()`. Same goes for the `else if` branch.

src/agreement/mod.rs
```diff
@@ -489,15 +493,21 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
+        self.on_coin_step(coin_step)
+    }
 
-        if let Some(coin) = self.common_coin.next_output() {
+    fn on_coin_step(
+        &mut self,
+        coin_step: CommonCoinStep<NodeUid>,
+    ) -> AgreementResult<FaultLog<NodeUid>> {
+        let mut fault_log = FaultLog::new();
+        fault_log.extend(coin_step.fault_log);
```
DrPeterVanNostrand: You could clone out the `FaultLog` from `coin_step` and save yourself a line:

```rust
let mut fault_log = coin_step.fault_log.clone();
```

vkomenda:
I attempted adding `Step::messages` in a straightforward manner and got a CI test failure in `tests/queueing_honey_badger.rs`. Other tests are passing.

If this PR still satisfies CI checks after the semi-automated [merge](https://github.com/poanetwork/hbbft/pull/130/commits/d7a2808774d75042bff72704a3dd4c0714f4a95e),it would help if we start working on finalising it still without the refactored message API. In the meantime, I would finish rebasing my WIP on messages and address the failing tests.

src/agreement/mod.rs
```diff
@@ -558,31 +579,33 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
+        if self.conf_round
+            && self.count_conf().0 >= self.netinfo.num_nodes() - self.netinfo.num_faulty()
+        {
+            // Invoke the comon coin.
+            let coin_step = self.common_coin.input(())?;
             self.extend_common_coin();
+            self.on_coin_step(coin_step)
+        } else {
+            // Continue waiting for (N - f) `Conf` messages
+            Ok(FaultLog::default())
```
DrPeterVanNostrand: You could use `FaultLog::new()` here, but both do the same thing.

src/common_coin.rs
```diff
@@ -90,26 +92,29 @@ where
...
-            return Ok(FaultLog::new());
-        }
-        let CommonCoinMessage(share) = message;
-        self.handle_share(sender_id, share)
+    ) -> Result<CommonCoinStep<NodeUid>> {
+        let fault_log = if !self.terminated {
+            let CommonCoinMessage(share) = message;
+            self.handle_share(sender_id, share)?
+        } else {
+            FaultLog::default()
```
DrPeterVanNostrand: `FaultLog::new()`

src/common_subset.rs
```diff
@@ -102,41 +102,41 @@ pub struct CommonSubset<NodeUid: Rand> {
     decided: bool,
 }
 
+pub type CommonSubsetStep<NodeUid> = Step<NodeUid, BTreeMap<NodeUid, ProposedValue>>;
```
DrPeterVanNostrand: In the above files, you were using `<N>` (not `<NodeUid>`) as the type parameter. Sticking to one naming convention might be beneficial.

src/common_subset.rs
```diff
@@ -244,9 +255,10 @@ impl<NodeUid: Clone + Debug + Ord + Rand> CommonSubset<NodeUid> {
         self.broadcast_results.insert(proposer_id.clone(), value);
         let set_agreement_input = |agreement: &mut Agreement<NodeUid>| {
             if agreement.accepts_input() {
-                agreement.set_input(true)
+                // FIXME: Use the result.
+                agreement.input(true)
             } else {
-                Ok(FaultLog::new())
+                Ok(Default::default())
```
DrPeterVanNostrand: `FaultLog::new()` vs `default()`, but it looks like you specifically chose to use `default` over `new` here. If that was you're intention, then this is fine with me.

src/dynamic_honey_badger/mod.rs
```diff
@@ -112,6 +112,8 @@ where
     output: VecDeque<Batch<C, NodeUid>>,
 }
 
+pub type DynamicHoneyBadgerStep<C, NodeUid> = Step<NodeUid, Batch<C, NodeUid>>;
```
DrPeterVanNostrand: Same, type parameter `<N>` vs `<NodeUid>`.

src/dynamic_honey_badger/mod.rs
```diff
@@ -181,6 +183,10 @@ where
     C: Eq + Serialize + for<'r> Deserialize<'r> + Debug + Hash,
     NodeUid: Eq + Ord + Clone + Debug + Serialize + for<'r> Deserialize<'r> + Hash + Rand,
 {
+    fn step(&mut self, fault_log: FaultLog<NodeUid>) -> Result<DynamicHoneyBadgerStep<C, NodeUid>> {
+        Ok(Step::new(self.output.drain(0..).collect(), fault_log))
```
DrPeterVanNostrand: You could use `.drain(..)` here.

src/dynamic_honey_badger/mod.rs
```diff
@@ -246,14 +250,20 @@ where
         self.verify_signature(sender_id, &sig, &kg_msg)?;
         let tx = SignedKeyGenMsg(self.start_epoch, sender_id.clone(), kg_msg, sig);
         self.key_gen_msg_buffer.push(tx);
-        self.process_output()
+        // FIXME: Remove the call to `process_output`. There wasn't any output from HB in this
+        // function.
+        self.process_output(Default::default())
```
DrPeterVanNostrand: Same, `FaultLog::new()` vs. `Default::default()` vs. `FaultLog::default()`.

src/honey_badger.rs
```diff
@@ -178,6 +175,13 @@ where
         HoneyBadgerBuilder::new(netinfo)
     }
 
+    fn step(
+        &mut self,
+        fault_log: FaultLog<NodeUid>,
+    ) -> HoneyBadgerResult<HoneyBadgerStep<C, NodeUid>> {
+        Ok(Step::new(self.output.drain(0..).collect(), fault_log))
```
DrPeterVanNostrand: Same, `drain(0..)` could be changed to `drain(..)`.

src/honey_badger.rs
```diff
@@ -545,23 +550,19 @@ where
...
-    fn process_output(&mut self) -> HoneyBadgerResult<FaultLog<NodeUid>> {
+    fn process_output(
+        &mut self,
+        step: CommonSubsetStep<NodeUid>,
+    ) -> HoneyBadgerResult<FaultLog<NodeUid>> {
         let mut fault_log = FaultLog::new();
-        if let Some(cs_output) = self.take_current_output() {
-            self.send_decryption_shares(cs_output)?
-                .merge_into(&mut fault_log);
+        fault_log.extend(step.fault_log);
```
DrPeterVanNostrand: Same, the fault log could be cloned out to save yourself a line:

```rust
let mut fault_log = step.fault_log.clone();
```

src/queueing_honey_badger.rs
```diff
@@ -122,6 +122,8 @@ where
     output: VecDeque<Batch<Tx, NodeUid>>,
 }
 
+pub type QueueingHoneyBadgerStep<Tx, NodeUid> = Step<NodeUid, Batch<Tx, NodeUid>>;
```
DrPeterVanNostrand: `<N>` vs. `<NodeUid>`

src/queueing_honey_badger.rs
```diff
@@ -133,42 +135,45 @@ where
...
         &mut self,
         sender_id: &NodeUid,
         message: Self::Message,
-    ) -> Result<FaultLog<NodeUid>> {
-        let mut fault_log = self.dyn_hb.handle_message(sender_id, message)?;
-        while let Some(batch) = self.dyn_hb.next_output() {
+    ) -> Result<QueueingHoneyBadgerStep<Tx, NodeUid>> {
+        let step = self.dyn_hb.handle_message(sender_id, message)?;
+        let mut fault_log = FaultLog::new();
+        fault_log.extend(step.fault_log);
```
DrPeterVanNostrand:
```rust
let mut fault_log = step.fault_log.clone()`;
```

src/queueing_honey_badger.rs
```diff
@@ -189,11 +194,20 @@ where
         QueueingHoneyBadgerBuilder::new(netinfo)
     }
 
+    fn step(
+        &mut self,
+        fault_log: FaultLog<NodeUid>,
+    ) -> Result<QueueingHoneyBadgerStep<Tx, NodeUid>> {
+        Ok(Step::new(self.output.drain(0..).collect(), fault_log))
```
DrPeterVanNostrand: `drain(..)`

src/agreement/mod.rs
vkomenda: Maybe we could `drain` the log instead of cloning it? Might go in a separate issue.

src/broadcast.rs
```diff
@@ -256,34 +258,32 @@ impl<NodeUid: Debug + Clone + Ord> DistAlgorithm for Broadcast<NodeUid> {
         // from this tree and send them, each to its own node.
         let proof = self.send_shards(input)?;
         let our_uid = &self.netinfo.our_uid().clone();
-        self.handle_value(our_uid, proof)
+        let fault_log = self.handle_value(our_uid, proof)?;
```
afck: Good catch!
How did our broadcast tests even pass without that?

src/messaging.rs
```diff
@@ -51,6 +51,37 @@ impl<M, N> TargetedMessage<M, N> {
     }
 }
 
+/// Result of one step of the local state machine of a distributed algorithm. Such a result should
+/// be used and never discarded by the client of the algorithm.
+pub struct Step<N, O>
```
afck: Should this be annotated with `#[must_use]`?
(Edit: Ah, I just saw you annotated the methods. But I think you can also annotate a `struct`; not sure exactly what difference that makes.)

Maybe it should also print an error when dropped (or even panic — at least if `#[cfg(test)]`) if the output hasn't been consumed.

src/agreement/mod.rs
```diff
@@ -489,15 +493,20 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
         self.extend_common_coin();
+        self.on_coin_step(coin_step)
+    }
 
-        if let Some(coin) = self.common_coin.next_output() {
+    fn on_coin_step(
+        &mut self,
+        coin_step: CommonCoinStep<NodeUid>,
+    ) -> AgreementResult<FaultLog<NodeUid>> {
+        let mut fault_log = coin_step.fault_log.clone();
```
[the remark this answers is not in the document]
DrPeterVanNostrand: Maybe destructure then? We still the `output` value.

```rust
let CommonCoinStep { fault_log, output } = coin_step;
```

src/queueing_honey_badger.rs
c0gent: I plan to reorganize the traits, adding supertraits...
```
    /// A transaction, user message, etc.
    pub trait Contribution: Eq + Debug + Hash {}
    impl<C> Contribution for C where C: Eq + Debug + Hash {}

    /// A peer node's unique identifier.
    pub trait NodeUid: Eq + Ord + Clone + Debug + Hash {}
    impl<N> NodeUid for N where N: Eq + Ord + Clone + Debug + Hash  {}
````
and shortening the trait parameters to single letters sometime when the churn dies down a bit.

File: src/messaging.rs (hunk 16 above)
Comment (c0gent) [reply src/messaging.rs]: Annotating the struct with `#[must_use]` is preferable to annotating the methods. Mostly because you don't have to remember to annotate every method that returns it.

File: src/common_subset.rs [hunk 18]
```diff
@@ -244,9 +255,10 @@ impl<NodeUid: Clone + Debug + Ord + Rand> CommonSubset<NodeUid> {
         self.broadcast_results.insert(proposer_id.clone(), value);
         let set_agreement_input = |agreement: &mut Agreement<NodeUid>| {
             if agreement.accepts_input() {
-                agreement.set_input(true)
+                // FIXME: Use the result.
```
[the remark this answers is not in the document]
vkomenda: Right, the comment seems obsolete.

src/queueing_honey_badger.rs
vkomenda: @c0gent, I used long names but please feel free to promote single-letter type variables if you have to.

src/broadcast.rs
```diff
@@ -43,7 +43,7 @@
 //!
 //! ## Example usage
 //!
-//! ```rust
```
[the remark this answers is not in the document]
vkomenda: I'm adding a note because I prefer this test to be fixed after the messaging API updates are merged as well.

src/agreement/mod.rs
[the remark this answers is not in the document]
vkomenda: @afck, what if the user wants to clone `FaultLog`? Wouldn't we violate the common traits convention if we remove `Clone` from a public type?

src/messaging.rs
vkomenda: I tried to add both the warning and the panic in the test configuration:
```
#[must_use = "The algorithm step result must be used."]
#[cfg_attr(test, forbid(unused_must_use))]
```
The first line works but second seems to have no effect. It still prints a warning in tests and doesn't stop comilation.

afck: Do you need to say `allow` instead of `forbid`?

Anyway, once #120 is done, this line should become unnecessary anyway.

vkomenda: It's either `forbid` or `deny`: https://doc.rust-lang.org/reference/attributes.html#lint-check-attributes

So, this works if you annotate tests rather than the type: `#[deny(unused_must_use)]`. It throws the required error.

vkomenda: I tried other combinations and none of them work:
* `#![cfg_attr(test, deny(unused_must_use))] in crate root,
* same in the `messaging.rs` morule,
* `deny` in `tests/network/mod.rs`.

The only attribution that works would be module-level in each test module. If `Cargo.toml` allowed to define dev attributes, that would probably suffice.

afck: I think if you annotate an item with something like `allow` or `forbid`, that applies to only the code in that item. So it would be an error to not use another `must_use` item inside of `Step`'s own code.

Annotating the tests accordingly would be correct, I think. (Once the tests are updated.)

src/agreement/mod.rs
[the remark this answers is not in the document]
afck: Maybe, yes. Depends on whether it's more helpful to prevent accidental cloning or to allow cloning.
Not sure; I was just thinking about our own code, where we'd probably want the former.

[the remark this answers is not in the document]
vkomenda: OK, I'll try to remove `Clone` from `FaultLog` then.

src/agreement/mod.rs
```diff
@@ -558,31 +581,33 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
-            let (count_vals, _) = self.count_conf();
-            if count_vals < self.netinfo.num_nodes() - self.netinfo.num_faulty() {
-                // Continue waiting for (N - f) `Conf` messages
-                return Ok(fault_log);
-            }
-            self.common_coin.input(())?.merge_into(&mut fault_log);
+        if self.conf_round
+            && self.count_conf().0 >= self.netinfo.num_nodes() - self.netinfo.num_faulty()
+        {
+            // Invoke the comon coin.
```
afck: Typo: `common`

src/dynamic_honey_badger/mod.rs
```diff
@@ -245,14 +249,20 @@ where
         self.verify_signature(sender_id, &sig, &kg_msg)?;
         let tx = SignedKeyGenMsg(self.start_epoch, sender_id.clone(), kg_msg, sig);
         self.key_gen_msg_buffer.push(tx);
-        self.process_output()
+        // FIXME: Remove the call to `process_output`. There wasn't any output from HB in this
+        // function.
+        self.process_output(Step::default())
```
afck: Feel free to do that right away. Or does something fail if you remove it?

src/honey_badger.rs
```diff
@@ -97,7 +97,7 @@ pub struct HoneyBadger<C, NodeUid: Rand> {
     /// The messages that need to be sent to other nodes.
     messages: MessageQueue<NodeUid>,
     /// The outputs from completed epochs.
-    output: VecDeque<Batch<C, NodeUid>>,
+    pub(crate) output: Vec<Batch<C, NodeUid>>,
```
afck: Why does that have to be `pub(crate)`? Shouldn't everything outside this module just use the `Step`?

src/dynamic_honey_badger/mod.rs
vkomenda: It did fail when I tried but that was some time ago. I can double-check.

src/honey_badger.rs
vkomenda: I think `output` gets accessed directly at the time when `HoneyBadger` is restarted.

src/queueing_honey_badger.rs
mbr: I think the usual style is single-letter? At least that is what I am used to from the stdlib and other crates.

c0gent:
Looks good to me :)

src/queueing_honey_badger.rs
c0gent: Yeah I reorganize/rename type params. per: https://github.com/poanetwork/hbbft/issues/143 probably next week when I have time.

PR #134, opened 2018-07-13
Add QueueingHoneyBadger::dyn_hb.

c0gent:
Needed to retrieve netinfo.

I'm not sure about the name or if we want it to be a permanent method at all but please add this so I don't have to keep cherry-picking it onto my updated branch every day :)


c0gent:
Agreed. That would be better.

Issue #139, opened 2018-07-16
Also zero the memory of `SecretKeySet`s.


afck:
#89 clears the memory of `SecretKey`s on drop, but not yet `SecretKeySet`s. Those have a `Poly` as their only field. Maybe we should just clear `Poly` on drop, too?

PR #140, opened 2018-07-16
Clear 'SecretKeySet' on drop.

DrPeterVanNostrand:
- Added `impl Drop for Poly` which fills the `coeff` vector with `Fr::zero()`.
- Added `impl Drop for BivarPoly` which fills the `coeff` vector with `Fr::zero()` and resets the `degree`.

Closes Issue #139

### DIFF
--- a/src/crypto/poly.rs
+++ b/src/crypto/poly.rs
@@ -18,6 +18,7 @@
 
 use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
+use std::ptr::write_volatile;
 use std::{cmp, iter, ops};
 
 use pairing::bls12_381::{Fr, FrRepr, G1, G1Affine};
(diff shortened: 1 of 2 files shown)


afck:
Sorry about that! That should have been an `ln -sf` [in the Travis file](https://github.com/poanetwork/hbbft/pull/138/files#diff-354f30a63fb0907d4ad57269548329e3R17).

DrPeterVanNostrand:
Np, I'll add `Drop` to `BivarPoly`. Should I change the travis file with this PR?

afck:
Sure, we can make the same change in every PR. Whatever gets merged first fixes it.

src/crypto/poly.rs
```diff
@@ -327,6 +340,20 @@ pub struct BivarPoly {
     coeff: Vec<Fr>,
 }
 
+impl Drop for BivarPoly {
+    fn drop(&mut self) {
+        let degree_ptr = self.degree as *mut usize;
+        let coeff_ptr = self.coeff.as_mut_ptr();
+        unsafe {
+            write_volatile(degree_ptr, 0);
```
afck: That's unnecessary; the degree is in the commitment, too, so it can't be a secret anyway.

Issue #45, opened 2018-05-24
Create a type of test adversary that can perform Man-in-the-Middle attacks


vkomenda:
Currently the test adversary can perform only eavesdropping attacks. To the attack in Issue #37, there has to be a more capable adversary that can also

- reorder messages,

- delay messages for any amount of time.

While man-in-the-middle attacks can be more powerful than the #37 attack, that attack requires the attacker to control the message flow rather than just listen to it. This level of control should be provided by the man-in-the-middle adversary type.

d33a94975ba60d59:
The net framework provides this functionality, as demonstrated by #270. I believe this issue can be closed.

mbr:
Yeah, this is outdated =).

Issue #37, opened 2018-05-22
Fix Common Coin use in `Agreement`.


afck:
The algorithm needs a third round of messages (`Conf`), otherwise an adversary that controls the network scheduler and one node can potentially stall agreement: https://github.com/amiller/HoneyBadgerBFT/issues/59

afck:
Apart from the tests, this is done. But the `Conf` messages might allow [further optimizations](https://github.com/amiller/HoneyBadgerBFT/issues/59#issuecomment-408664323).

afck:
Closing this, since we have #228 for the missing test.

Issue #84, opened 2018-06-26
Random adversary.


afck:
While it's impossible to cover every conceivable sophisticated attack strategy, we should send at least one fake message of each variant in some test, i.e. for each algorithm have a random adversary that creates all kinds of nonsense messages.

Where possible, we could make it a bit smarter (or have a separate, less random adversary) and e.g. generate fake messages with the correct epoch (or ±1), since that is more likely to expose bugs.

mbr:
Assign me, please? =)

afck:
I can't, possibly because you don't have push access?
I don't have admin permissions for the repository. I'll assign myself as a placeholder for now.

mbr:
A first draft is in PR #112, for your consideration ;)

Some comments beyond:

> While it's impossible to cover every conceivable sophisticated attack strategy, we should send at least one fake message of each variant in some test, i.e. for each algorithm have a random adversary that creates all kinds of nonsense messages.

* [x] Nonsense
* [x] Replay attacks

> Where possible, we could make it a bit smarter (or have a separate, less random adversary) and e.g. generate fake messages with the correct epoch (or ±1), since that is more likely to expose bugs.

Right now the epoch is random, mainly because this was a way longer hanging fruit. It is not impossible to add though, see the PR #112 for details.

> Another way to generate more "realistic" fake messages is to simulate a complete instance of the algorithm in parallel and use it as inspiration, i.e. modify its messages a bit and send them to the honest nodes.

As mentioned in PR #112, the `Adversary` API could use some changes. One addition not proposed in the PR would be to restructure the whole construct and have faulty as well as correct nodes take part in the network the same way. Instead of forcing the `Adversary` to handle messages for their nodes through the `push_message()` function, the adversary should be a message filter in front of each node instead.

This would allow an adversary implementation to easily play along and only occasionally change a message.

## Further ideas

I would like to suggest two other things: It seems that a few bits of random input could be tested more efficiently, e.g. property-based testing could be used to generate these inputs. E.g., when testing the broadcast algorithm, a crate like [proptest](https://docs.rs/proptest) or [quickcheck](https://docs.rs/quickcheck/) would generate multiple inputs to be tested. Proptest has the nice feature that it will save the random seed used in a file that can be checked into version control, causing all tests that failed in the past to be run against first; this efficiently keeps track of regressions.

[...]

PR #141, opened 2018-07-17
Allowed adversary to control message delivery

vkomenda:
Part of #45. Also required by #37.

The adversary fully controls message delivery.

### DIFF
--- a/tests/honey_badger.rs
+++ b/tests/honey_badger.rs
@@ -21,7 +21,7 @@ use std::sync::Arc;
 use itertools::Itertools;
 use rand::Rng;
 
-use hbbft::honey_badger::{self, Batch, HoneyBadger, MessageContent};
+use hbbft::honey_badger::{Batch, HoneyBadger, MessageContent};
 use hbbft::messaging::{NetworkInfo, Target, TargetedMessage};
 use hbbft::transaction_queue::TransactionQueue;
 
@@ -66,21 +66,22 @@ impl Adversary<UsizeHoneyBadger> for FaultyShareAdversary {
 
     fn push_message(
         &mut self,
-        sender_id: NodeUid,
-        msg: TargetedMessage<honey_badger::Message<NodeUid>, NodeUid>,
-    ) {
-        let NodeUid(sender_id) = sender_id;
+        msg: MessageWithSender<UsizeHoneyBadger>,
+    ) -> Vec<MessageWithSender<UsizeHoneyBadger>> {
+        let NodeUid(sender_id) = msg.sender;
         if sender_id < self.num_good {
             if let TargetedMessage {
                 target: Target::All,
                 message,
-            } = msg
+            } = msg.tm.clone()
             {
                 let epoch = message.epoch();
                 // Set the trigger to simulate decryption share messages.
                 self.share_triggers.entry(epoch).or_insert(true);
             }
         }
+        // Output the message without changes.
+        vec![MessageWithSender::new(NodeUid(sender_id), msg.tm)]
     }
 
     fn step(&mut self) -> Vec<MessageWithSender<UsizeHoneyBadger>> {
@@ -137,7 +138,7 @@ where
         }
         if node.outputs().last().unwrap().is_empty() {
             let last = node.outputs().last().unwrap().epoch;
-            node.queue.retain(|(_, ref msg)| msg.epoch() < last);
+            node.queue.retain(|(_, msg)| msg.epoch() < last);
         }
         false
     };
--- a/tests/honey_badger.rs
+++ b/tests/honey_badger.rs
@@ -72,8 +72,8 @@ impl Adversary<UsizeHoneyBadger> for FaultyShareAdversary {
         if sender_id < self.num_good {
             if let TargetedMessage {
                 target: Target::All,
-                message,
-            } = msg.tm.clone()
+                ref message,
+            } = msg.tm
             {
                 let epoch = message.epoch();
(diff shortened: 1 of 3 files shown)


tests/honey_badger.rs
```diff
@@ -66,21 +66,22 @@ impl Adversary<UsizeHoneyBadger> for FaultyShareAdversary {
...
-        let NodeUid(sender_id) = sender_id;
+        msg: MessageWithSender<UsizeHoneyBadger>,
+    ) -> Vec<MessageWithSender<UsizeHoneyBadger>> {
+        let NodeUid(sender_id) = msg.sender;
         if sender_id < self.num_good {
             if let TargetedMessage {
                 target: Target::All,
                 message,
-            } = msg
+            } = msg.tm.clone()
```
afck: You should be able to avoid the cloning with:

```rust
if let TargetedMessage {
    target: Target::All,
    ref message,
} = msg.tm
```

tests/network/mod.rs
```diff
@@ -439,28 +464,39 @@ where
...
-            match msg.target {
-                Target::All => {
-                    for node in self.nodes.values_mut() {
-                        if node.id != sender_id {
-                            node.queue.push_back((sender_id, msg.message.clone()))
+            // Transform the queued message by means of the adversary.
+            let adv_msgs = self.adversary.push_message(MessageWithSender {
+                sender: sender_id,
+                tm: msg.clone(),
+            });
```
afck: I think we should at least make sure that the original message, if it was from a correct to a correct node, is unchanged. E.g. we could just give a reference to `push_message`, and dispatch its outputs _in addition_ to the original one.

Also, we should probably assert that all the `adv_msgs` have as sender one of the adversary's nodes: The adversary is not able to forge, modify or indefinitely block messages from a different sender.

tests/network/mod.rs
```diff
@@ -439,28 +464,39 @@ where
...
-                    } else if let Some(node) = self.nodes.get_mut(&to_id) {
-                        node.queue.push_back((sender_id, msg.message));
-                    } else {
-                        warn!(
-                            "Unknown recipient {:?} for message: {:?}",
-                            to_id, msg.message
-                        );
+                    Target::Node(to_id) => {
+                        if self.adv_nodes.contains_key(&to_id) {
+                            self.adversary.push_message(MessageWithSender {
```
afck: Or should that even be a different method? This one is for messages sent to the adversary, while the other invocation is kind of eavesdropping on any communication.

tests/network/mod.rs
vkomenda: For #37, `push_message` can withhold the message within the adversary and return an empty vector. In that case nothing should be dispatched. In the current tests this is still not implemented. That's why all `push_message`s return the same message. But it is not the case in general.

tests/network/mod.rs
vkomenda: Or we could rename `push_message`. It is more of a `transform_message` now.

tests/network/mod.rs
vkomenda: What you propose for `adv_msgs` would not allow implementing #37. `adv_msgs` is the result of the adversarial message transformation such as reordering of messages from correct nodes.

afck: But maybe reordering should be handled differently: by allowing adversaries to flip entries in the nodes' incoming queues? That, in combination with picking the next node to handle a message, should allow the adversary to do everything they are allowed to do.
(Alternatively, I think @mbr suggested to have just one queue for all the nodes, where entries have both sender and recipient IDs. Reordering that queue would replace both of the above: It allows the adversary to choose message order and the next node.)

At least I think that functionality should be separate from the method that informs the adversary about a message arriving in one of their own nodes. (Which is all that `push_message` currently does, I think. Not a great name, admittedly. :grimacing: )

afck: > This would require modifying the current scheduling mechanism which is at the moment based on picking nodes and delivering all queued incoming messages to it.

Exactly; it would replace `pick_node` with something like `shuffle_queue`, and we could keep `push_message` in its current form. (Or later replace it with something else, too — I think one idea was to actually instantiate correct nodes for the adversary, and just allow the adversary to modify _their_ incoming and outgoing messages?)

vkomenda: If we keep `push_message` in its current form, where do you think the messages should be pushed onto the single queue? `shuffle_queue` is a great idea. I don't quite see where the queue itself comes from in your proposal though.

afck: I was thinking that the `TestNetwork` itself would own the queue, and only allow the adversary to move messages inside it. (Not sure how the `Adversary` interface would have to look for that… maybe a method that gets a non-mutable reference to the queue and outputs a list of positions, i.e. a permutation?) That would make it impossible to write an adversary that cheats by removing or modifying entries.

But I guess it's also okay to do something simpler, and just make sure all our adversaries behave.

vkomenda: I'm maybe OK with the permutation interface. However, if we need another kind of attack in the future, we would probably rewrite the queue to have mutable access rather than add another interface to it.

@mbr, what do you think, does the single queue have to be owned by `Adversary`, in which case we have to ensure adversaries don't perform unexpected transformations on the queue, or should the queue be owned by `TestNetwork` with specialized interfaces for particular kinds of adversarial transformations?

afck:
I actually think it would be better to postpone the whole issue until we completed the API change, and the tests have been rewritten.

vkomenda:
Closing in expectation of the ongoing work on #84.

PR #146, opened 2018-07-18
Separate keys and key shares.

### LINKED ISSUE #131: Use a separate key set for threshold cryptography.
Currently for up to three nodes the threshold is 0, so they will all have _identical_ key shares. If this is used for anything other than threshold cryptography, we should probably have a second, separate key set for which everyone just generated their key independently.

In particular, it would be safer if that second key set would be used for DKG.

afck:
This adds another map of public keys to `NetworkInfo`: It is unsafe to use key _shares_ as keys.

Since creating a new `NetworkInfo` is even more complicated now, `NetworkInfo::generate_map` creates a set of matching instances for testing.

(I also noticed that Clippy wasn't applied to the example, so I added it and fixed the lints.)

Closes #131

### DIFF
--- a/src/messaging.rs
+++ b/src/messaging.rs
@@ -1,7 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::fmt::Debug;
 
-use crypto::{PublicKey, PublicKeySet, SecretKey};
+use crypto::{PublicKey, PublicKeySet, PublicKeyShare, SecretKey, SecretKeyShare};
 use fault_log::FaultLog;
 
 /// Message sent by a given source.
@@ -139,16 +139,18 @@ impl<'a, D: DistAlgorithm + 'a> Iterator for MessageIter<'a, D> {
     }
 }
 
-/// Common data shared between algorithms.
+/// Common data shared between algorithms: the nodes' IDs and key shares.
 #[derive(Debug, Clone)]
 pub struct NetworkInfo<NodeUid> {
     our_uid: NodeUid,
-    all_uids: BTreeSet<NodeUid>,
     num_nodes: usize,
     num_faulty: usize,
     is_validator: bool,
+    // TODO: Should this be an option? It only makes sense for validators.
+    secret_key_share: SecretKeyShare,
     secret_key: SecretKey,
     public_key_set: PublicKeySet,
+    public_key_shares: BTreeMap<NodeUid, PublicKeyShare>,
     public_keys: BTreeMap<NodeUid, PublicKey>,
     node_indices: BTreeMap<NodeUid, usize>,
 }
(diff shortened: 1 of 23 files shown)


c0gent:
Looks good so far. I'll plug it in and try it out today.

src/messaging.rs
```diff
@@ -243,4 +264,54 @@ impl<NodeUid: Clone + Ord> NetworkInfo<NodeUid> {
...
         self.is_validator
     }
+
+    /// Returns `true` if the given node takes part in the consensus itself. If not, it is only an
+    /// observer.
+    pub fn is_node_validator(&self, uid: &NodeUid) -> bool {
+        self.public_keys.contains_key(uid)
+    }
+
+    /// Generates a map of matching `NetworkInfo`s for testing.
```
DrPeterVanNostrand: Maybe add `#[cfg(test)]`?

DrPeterVanNostrand: Everything else looks good though!

afck: Then the build fails for me. I think `#[cfg(test)]` is only included for the `--lib`'s own unit tests, but not available to the integration tests or the `simulation` example.

src/sync_key_gen.rs
```diff
@@ -10,7 +10,89 @@
 //! signatures and encryption. The secret master key is not known by anyone. The protocol succeeds
 //! if up to `threshold` nodes are faulty.
 //!
-//! # How it works
+//! ## Usage example
+//!
+//! ```
```
c0gent: Doc code should be prefixed with ` ```rust,ignore ` or ` ` ```rust,no_run ` (to compile only). Ideally though it's better to let it run it that way it stays up to date.

afck: Yes, in this case I wrote the example in a way that it runs and passes. (`` ``` `` is equivalent to `` ```rust ``.)

I hope that doesn't go too much at the expense of helpfulness and readability. An alternative would be to write an example that's more like what you would write in an actual application, instead of simulating several nodes.

c0gent:
Might `Proposal` be a better name for `Propose` since it's a noun? :)

src/sync_key_gen.rs
c0gent: Oh, ` ``` ` is equivalent now? I guess I missed that change to rustdoc.

c0gent: Yeah nevermind I'm just mixing it up with regular markdown. Ignore me.

src/sync_key_gen.rs
```diff
@@ -10,7 +10,89 @@
...
+//! }
+//!
+//! // All nodes now handle the proposals and send the resulting `Accept` messages.
+//! let mut accepts = Vec::new();
+//! for (sender_id, proposal) in proposals {
+//!     for (&id, node) in &mut nodes {
+//!         match node.handle_propose(&sender_id, proposal.clone()) {
+//!             Some(ProposeOutcome::Valid(accept)) => accepts.push((id, accept)),
+//!             Some(ProposeOutcome::Invalid(faults)) => panic!("Invalid proposal: {:?}", faults),
+//!             None => panic!("We are not an observer, so we should send Accept."),
```
c0gent: The wording here is a tiny bit confusing (or at least it was to me at first glance). More detailed documentation on `::handle_propose` (and other methods) explaining why/when `None` is returned would probably be helpful.

afck:
> Might `Proposal` be a better name for `Propose` since it's a noun? :)

I have an ask for you native speakers. Take this as an invite to have your say:
Can't I just noun `Propose`? ([Relevant xkcd](https://xkcd.com/1443/))

I'll make another PR with further extensions to the documentation. Maybe I'll rename the message types to `Proposal` and `Acceptance`, although the latter is a bit lengthy. How about `Part` and `Ack`? (The resulting key sets will literally be the sum of the `Part`s.)

PR #150, opened 2018-07-18
Outgoing message API refactoring

### LINKED ISSUE #149: Refactor the outgoing message API
As a continuation of #66, following up on #130 we need to add outgoing messages to the algorithm `Step`.

vkomenda:
Close #149 

This branch is currently WIP since the tests of `DynamicHoneyBadger` and `QueueingHoneyBadger` don't pass.

### DIFF
--- a/examples/network/node.rs
+++ b/examples/network/node.rs
@@ -144,8 +144,8 @@ impl<T: Clone + Debug + AsRef<[u8]> + PartialEq + Send + Sync + From<Vec<u8>> + 
 
                 if let Some(v) = value {
                     // FIXME: Use the output.
-                    let _ = broadcast.input(v.clone().into()).expect("propose value");
-                    for msg in broadcast.message_iter() {
+                    let step = broadcast.input(v.clone().into()).expect("propose value");
+                    for msg in step.messages {
                         tx_from_algo.send(msg).expect("send from algo");
                     }
                 }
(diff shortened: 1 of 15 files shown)


DrPeterVanNostrand:
> error: very complex type used

vkomenda:
In order to find the issue that causes test failures of `DynamicHoneyBadger` and `QueueingHoneyBadger` I tried producing a first step of those algorithms just they are created, e.g.,
```
QueueingHoneyBadger::build(&self) -> (QueueingHoneyBadger<Tx, NodeUid>, Step<Tx, NodeUid>)
```
That didn't change the outcome of the tests a bit and moreover it became clear to me that, at least in the tests, the outgoing message queues of both of these algorithms are empty, which rendered the attempt useless as far as fixing the tests is concerned. The simulation however does start with a non-empty vector of messages and does fail as well because the simulated network infrastructure is very close to that of the integration tests, although it constructs the `QueueingHoneyBadger` using a slightly different builder function.

I'm not going to pursue the line of making a step at the time of algorithm creation any further. It complicates types of results and requires extra handling of initial outputs and messages in `TestNetwork::new`, both in the integration tests and in the simulation. Instead I'm going to make these algorithms consistent with the rest and have all possible inputs provided to those via the `input` methods. These methods produce a step in a uniform way.

For now I'll just rebase this branch.

PR #153, opened 2018-07-18
Small correction.

c0gent:
Presumably (I haven't looked) `::handle_propose` checks whether or not the `sender_id` is in the list of nodes passed in during creation (or == `our_id`). Shall I add documentation to the method explaining this and about the `Some`/`None` return values?


c0gent:
Whoops, Nevermind! (Docs still need some beefing -- but I'm sure we'll get to it).

afck:
Yes, the documentation is still severely lacking. Sorry about that! I'll try to add some today.

PR #154, opened 2018-07-19
Extend DKG docs, rename messages, add into_network_info.

afck:
This adds a "Usage" section to the `sync_key_gen` module and extends some item documentation, too.

The new `SyncKeyGen::into_network_info` method avoids cloning the (non-threshold) keys and helps simplify `DynamicHoneyBadger` a bit.

Finally, `Propose` and `Accept` are renamed to `Part` and `Ack`:
* Structs ideally shouldn't have verbs as their name.
* "`Proposal`"/"`Propose`" suggests that it's a proposal for the actual final keys, but in fact it's just a summand for them. Every node is supposed to contribute a `Part`, and in the ideal case, every node's `Part` will be used.
* "`Ack`" is much shorter than "`Acceptance`" or "`Acknowledgement`".

### DIFF
--- a/src/sync_key_gen.rs
+++ b/src/sync_key_gen.rs
@@ -8,9 +8,46 @@
 //!
 //! When the protocol completes, every node receives a secret key share suitable for threshold
 //! signatures and encryption. The secret master key is not known by anyone. The protocol succeeds
-//! if up to `threshold` nodes are faulty.
+//! if up to _t_ nodes are faulty, where _t_ is the `threshold` parameter. The number of nodes must
+//! be at least _2 t + 1_.
 //!
-//! ## Usage example
+//! ## Usage
+//!
+//! Before beginning the threshold key generation process, each validator needs to generate a
+//! regular (non-threshold) key pair and multicast its public key. `SyncKeyGen::new` returns the
+//! instance itself and a `Propose` message, containing a contribution to the new threshold keys.
+//! It needs to be sent to all nodes. `SyncKeyGen::handle_propose` in turn produces an `Accept`
+//! message, which is also multicast.
+//!
+//! All nodes must handle the exact same set of `Propose` and `Accept` messages. In this sense the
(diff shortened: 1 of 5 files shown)


src/sync_key_gen.rs
```diff
@@ -279,76 +335,86 @@ impl<NodeUid: Ord + Clone + Debug> SyncKeyGen<NodeUid> {
...
         }
         let opt_sk = opt_sk_val.map(SecretKeyShare::from_value);
         (pk_commit.into(), opt_sk)
     }
 
-    /// Handles an `Accept` message or returns an error string.
-    fn handle_accept_or_err(
+    /// Consumes the instance, generates the key set and returns a new `NetworkInfo` with the new
+    /// keys.
+    pub fn into_network_info(self) -> NetworkInfo<NodeUid> {
```
DrPeterVanNostrand: We could `impl<NodeUid> From<SyncKeyGen> for NetworkInfo<NodeUid>` here.  I like using conversion traits, but it may be clearer as currently written.

afck: The reason why I was hesitating is that maybe `From` suggests that it's a "simple" and cheap conversion, where in reality it does quite a bit of computation to generate the keys, and just happens to consume the `SyncKeyGen` to avoid some cloning.

But on the one hand I'm not sure that `into_network_info` doesn't suggest the same thing; and on the other hand I don't know whether it's only me who would read it that way.

Another option would be renaming it to something like `generate_network_info`.
Not sure what's best?

afck: Also, I'd probably assume that if it's `Into`/`From`, it's _always_ fine to do the conversion; but it only produces a safe result after it ate enough `Part` and `Ack` messages.

DrPeterVanNostrand: I definitely agree with your point that `Into/From` would infer that the conversion is always possible, which rules the conversion trait out to me.

PR #155, opened 2018-07-20
Clarify synchrony requirements in SyncKeyGen.

afck:


src/sync_key_gen.rs
```diff
@@ -297,6 +297,9 @@ impl<NodeUid: Ord + Clone + Debug> SyncKeyGen<NodeUid> {
     /// Handles a `Part` message. If it is valid, returns an `Ack` message to be broadcast.
     ///
     /// If we are only an observer, `None` is returned instead and no messages need to be sent.
+    ///
+    /// All participating nodes must handle the exact same sequence of messages.
+    /// Note that `handle_part` also needs to explicitely be called with this instance's own `Part`.
```
c0gent: explicitly

afck: Thanks! I force-pushed; didn't want to pollute the commit history because of one byte.

c0gent: I use the force often...

PR #157, opened 2018-07-21
Handle initial QHB messages, fix DHB.

afck:


src/dynamic_honey_badger/mod.rs
```diff
@@ -266,22 +236,22 @@ where
...
-        step: honey_badger::Step<InternalContrib<C, NodeUid>, NodeUid>,
-    ) -> Result<FaultLog<NodeUid>> {
-        let mut fault_log = FaultLog::new();
-        fault_log.extend(step.fault_log);
+        hb_step: honey_badger::Step<InternalContrib<C, NodeUid>, NodeUid>,
+    ) -> Result<Step<C, NodeUid>> {
+        let mut step: Step<C, NodeUid> = Step::default();
         let start_epoch = self.start_epoch;
-        for hb_batch in step.output {
+        let output = step.extend_with(hb_step, |hb_msg| Message::HoneyBadger(start_epoch, hb_msg));
```
vkomenda: Well spotted!

afck:
Thanks!
I agree, this use of the builder pattern is not ideal. But having an input instead would also be weird. I couldn't come up with a better solution yet.

PR #159, opened 2018-07-23
Accept more types in threshold crypto API.

afck:
This removes some unnecessary allocation and conversion by accepting
more primitive types and references as the index in threshold decryption
and signing, and as the argument to a polynomial.

It also adds scalar multiplication, addition and subtraction to `Poly` for convenience.

### DIFF
--- a/src/crypto/into_fr.rs
+++ b/src/crypto/into_fr.rs
@@ -1,0 +1,55 @@
+use pairing::bls12_381::Fr;
+use pairing::{Field, PrimeField};
+
+/// A conversion into an element of the field `Fr`.
+pub trait IntoFr: Copy {
+    fn into_fr(self) -> Fr;
+}
+
+impl IntoFr for Fr {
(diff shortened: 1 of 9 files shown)


src/crypto/into_fr.rs
```diff
@@ -0,0 +1,55 @@
+use pairing::bls12_381::Fr;
+use pairing::{Field, PrimeField};
+
+/// A conversion into an element of the field `Fr`.
+pub trait IntoFr: Copy {
```
vkomenda: Is there an advantage of this specialized type over the more idiomatic `Into<Fr>`? I think `Fr::into` would look better.

afck: I'm not allowed to add `impl X for Y` in a crate to which neither `X` nor `Y` belongs. (With generic arguments it's a bit more complicated, I think, but I can't e.g. `impl Into<Fr> for usize`.)

Issue #160, opened 2018-07-23
Remove the output and message queue fields.


afck:
With #149 done, the outgoing message queue and output queue, which are still present in some of the algorithms, should now always be empty between message calls. Consider removing them, where feasible.

PR #161, opened 2018-07-23
Remove output and message queue from HoneyBadger.

afck:
(See #160.)

### DIFF
--- a/src/honey_badger.rs
+++ b/src/honey_badger.rs
@@ -24,10 +24,11 @@
 
 use rand::Rand;
 use std::collections::btree_map::Entry;
-use std::collections::{BTreeMap, BTreeSet, VecDeque};
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Debug;
 use std::hash::Hash;
 use std::marker::PhantomData;
+use std::mem;
 use std::sync::Arc;
 
 use bincode;
@@ -36,8 +37,8 @@ use serde::{Deserialize, Serialize};
 
 use common_subset::{self, CommonSubset};
 use crypto::{Ciphertext, DecryptionShare};
-use fault_log::{FaultKind, FaultLog};
-use messaging::{self, DistAlgorithm, NetworkInfo, Target, TargetedMessage};
+use fault_log::{Fault, FaultKind, FaultLog};
+use messaging::{self, DistAlgorithm, NetworkInfo, Target};
 
 error_chain!{
     links {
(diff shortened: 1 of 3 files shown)


src/honey_badger.rs
```diff
@@ -358,64 +350,44 @@ where
...
-        self.try_output_batch()
-    }
-
-    /// Returns true if and only if contributions have been decrypted for all selected proposers in
-    /// this epoch.
-    fn all_contributions_decrypted(&mut self) -> bool {
-        match self.ciphertexts.get(&self.epoch) {
-            None => false, // No ciphertexts yet.
-            Some(ciphertexts) => ciphertexts.keys().eq(self.decrypted_contributions.keys()),
+    fn try_decrypt_and_output_batch(&mut self) -> Result<Step<C, NodeUid>> {
```
vkomenda: This function has changed quite a bit. Since `try_output_batch` is doing decryption now, it may seem strange that only this function has `decrypt` in its name. Because it is applying `try_output_batch` iteratively, maybe you could rename it to something with `iterate` or `complete` in the name instead of `decrypt`?

src/honey_badger.rs
```diff
@@ -567,23 +547,18 @@ where
...
-            }
+        cs_step: common_subset::Step<NodeUid>,
+        epoch: u64,
+    ) -> Result<Step<C, NodeUid>> {
+        let mut step = Step::default();
+        let cs_outputs = step.extend_with(cs_step, |cs_msg| {
+            MessageContent::CommonSubset(cs_msg).with_epoch(epoch)
+        });
+        for cs_output in cs_outputs {
+            // There is at most one output.
```
vkomenda: I'm fine to keep this as a comment but maybe the code could check that statement or ensure multiple outputs don't get `extend`ed so easily? What would be the best solution here? The function might return a `Fault` if there is a second output.

afck: I was thinking we might change `Step` and `DistAlgorithm` a bit, and essentially replace what's currently `Output` with either `Option<Output>` or `Vec<Output>`, depending on the algorithm. Not sure whether that's a good idea…

src/honey_badger.rs
afck: Or just `try_output_batches`? `try_output_all_batches`?

src/honey_badger.rs
afck: I wouldn't return a fault: it can't be attributed to a particular node, and this could actually only happen due to a bug in the code. So it should be either an error or a panic, I guess.

afck: I made it an error log for now.

PR #163, opened 2018-07-24
Remove queue fields from Agreement, CommonCoin, Broadcast and CommonSubset.

afck:
See #160.

### DIFF
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -145,6 +145,7 @@ impl rand::Rand for AgreementContent {
 
 /// Possible values of the common coin schedule defining the method to derive the common coin in a
 /// given epoch: as a constant value or a distributed computation.
+#[derive(Debug)]
 enum CoinSchedule {
     False,
     True,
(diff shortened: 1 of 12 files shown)


src/agreement/mod.rs
```diff
@@ -323,7 +303,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         self.epoch == 0 && self.estimated.is_none()
     }
 
-    fn handle_bval(&mut self, sender_id: &NodeUid, b: bool) -> Result<FaultLog<NodeUid>> {
+    fn handle_bval(&mut self, sender_id: &NodeUid, b: bool) -> Result<Step<NodeUid>> {
```
vkomenda: Ideally we could distinguish beween the `Step` that the `DistAlgorithm` returns to the caller of a public function call and an intermediate data structure that is used to construct such a step.

afck: Possibly, yes. Or at least make some of the currently public `Step` methods `pub(crate)`.

PR #165, opened 2018-07-25
Added mlocking for secret types.

DrPeterVanNostrand:
- Added trait `ContainsSecret`.
- Added `mlock` and `munlock` for secret types.
- Removed stack copying of secret prime field elements.
- Changed `SecretKey(Fr)` to `SecretKey(Box<Fr>)`.
- Added [`memsec`](https://crates.io/crates/memsec) dependency.
- Redacted `Debug` implementations of secrets.

Closes Issue #89

# Request for Comments

@vkomenda @afck @c0gent @mbr

Regarding the recent error handling debates, I'm not sure how to handle errors from syscalls. You will see in places where `ContainsSecret` is implemented that I just print out when an `mlock` or `munlock` call fails. 2/3rds of the `ContainsSecret` trait methods are only called from the `drop()` method, which can't return an error. Is this a case for `panic()`?

This PR is **NOT** ready to be merged until this is discussed because `println!("You're crypto is broke")` is probably not the ideal way to handle memory errors.

### DIFF
--- a/src/crypto/error.rs
+++ b/src/crypto/error.rs
@@ -6,5 +6,13 @@ error_chain! {
         DuplicateEntry {
             description("signature shares contain a duplicated index")
         }
+        MlockFailed(desc: String) {
+            description("failed mlock a region of memory")
+            display("{}", desc)
+        }
+        MunlockFailed(desc: String) {
+            description("failed munlock a region of memory")
+            display("{}", desc)
+        }
     }
 }
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -15,6 +15,7 @@ use std::mem::size_of_val;
 use std::ptr::{copy_nonoverlapping, write_volatile};
 
 use byteorder::{BigEndian, ByteOrder};
+use errno::errno;
 use init_with::InitWith;
 use memsec::{memzero, mlock, munlock};
 use pairing::bls12_381::{Bls12, Fr, G1, G1Affine, G2, G2Affine};
@@ -230,7 +231,7 @@ impl ContainsSecret for SecretKey {
         if mlock_succeeded {
             Ok(())
         } else {
-            let msg = "failed to mlock `SecretKey` memory".to_string();
+            let msg = format!("failed to mlock `SecretKey` memory ({:?})", errno());
             Err(ErrorKind::MlockFailed(msg).into())
         }
     }
@@ -242,7 +243,7 @@ impl ContainsSecret for SecretKey {
         if munlock_succeeded {
             Ok(())
         } else {
-            let msg = "failed to munlock `SecretKey` memory".to_string();
+            let msg = format!("failed to munlock `SecretKey` memory ({:?})", errno());
             Err(ErrorKind::MunlockFailed(msg).into())
         }
     }
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -226,9 +226,7 @@ impl ContainsSecret for SecretKey {
     fn mlock_secret_memory(&self) -> Result<()> {
         let ptr = &*self.0 as *const Fr as *mut u8;
         let n_bytes = size_of_val(&*self.0);
-        let mlock_succeeded = unsafe {
-            mlock(ptr, n_bytes)
-        };
+        let mlock_succeeded = unsafe { mlock(ptr, n_bytes) };
         if mlock_succeeded {
             Ok(())
         } else {
@@ -240,9 +238,7 @@ impl ContainsSecret for SecretKey {
     fn munlock_secret_memory(&self) -> Result<()> {
         let ptr = &*self.0 as *const Fr as *mut u8;
         let n_bytes = size_of_val(&*self.0);
-        let munlock_succeeded = unsafe {
-            munlock(ptr, n_bytes)
-        };
+        let munlock_succeeded = unsafe { munlock(ptr, n_bytes) };
         if munlock_succeeded {
             Ok(())
         } else {
@@ -644,9 +640,18 @@ mod tests {
         assert_ne!(pk_set.public_key(), pk_set.public_key_share(2).0);
 
         // Make sure we don't hand out the main secret key to anyone.
-        assert_ne!(sk_set.secret_key().unwrap(), sk_set.secret_key_share(0).unwrap().0);
-        assert_ne!(sk_set.secret_key().unwrap(), sk_set.secret_key_share(1).unwrap().0);
-        assert_ne!(sk_set.secret_key().unwrap(), sk_set.secret_key_share(2).unwrap().0);
+        assert_ne!(
+            sk_set.secret_key().unwrap(),
+            sk_set.secret_key_share(0).unwrap().0
+        );
+        assert_ne!(
+            sk_set.secret_key().unwrap(),
+            sk_set.secret_key_share(1).unwrap().0
+        );
+        assert_ne!(
+            sk_set.secret_key().unwrap(),
+            sk_set.secret_key_share(2).unwrap().0
+        );
 
         let msg = "Totally real news";
 
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -11,13 +11,15 @@ pub mod serde_impl;
 
 use std::fmt;
 use std::hash::{Hash, Hasher};
-use std::ptr::write_volatile;
+use std::mem::size_of_val;
+use std::ptr::{copy_nonoverlapping, write_volatile};
 
 use byteorder::{BigEndian, ByteOrder};
 use init_with::InitWith;
+use memsec::{memzero, mlock, munlock};
 use pairing::bls12_381::{Bls12, Fr, G1, G1Affine, G2, G2Affine};
 use pairing::{CurveAffine, CurveProjective, Engine, Field};
-use rand::{ChaChaRng, OsRng, Rng, SeedableRng};
+use rand::{ChaChaRng, OsRng, Rand, Rng, SeedableRng};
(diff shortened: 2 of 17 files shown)


src/crypto/mod.rs
```diff
@@ -28,6 +30,18 @@ const CHACHA_RNG_SEED_SIZE: usize = 8;
 
 const ERR_OS_RNG: &str = "could not initialize the OS random number generator";
 
+/// Marks a type as containing one or more secret prime field elements.
+pub(crate) trait ContainsSecret {
+    /// Calls the `mlock` system call on the region of memory allocated for
+    /// the secret prime field element or elements. This tesults in that
```
DrPeterVanNostrand: Spelling

src/crypto/mod.rs
```diff
@@ -10,13 +10,15 @@ pub mod serde_impl;
 
 use std::fmt;
 use std::hash::{Hash, Hasher};
-use std::ptr::write_volatile;
+use std::mem::size_of_val;
+use std::ptr::{copy_nonoverlapping, write_volatile};
 
 use byteorder::{BigEndian, ByteOrder};
 use init_with::InitWith;
+use libc::{c_void, memset, mlock, munlock};
```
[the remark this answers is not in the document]
vkomenda: This is a good point since `m(un)lock` are in the `unix` module of `libc` and that module is not included on Windows.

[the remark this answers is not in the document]
DrPeterVanNostrand: I was using the [`memsec`](https://docs.rs/memsec/0.5.4/memsec/) crate for `mlock`, `munlock`, and `memzero` functions, but I was getting `SIGSEGV`segfaults from their `munlock` on EC2 Ubuntu. The `memsec` crate is cross platform. I'll keep trying to get it to work.

src/crypto/mod.rs
```diff
@@ -156,47 +170,107 @@ impl fmt::Debug for SignatureShare {
...
+    }
+}
+
+impl ContainsSecret for SecretKey {
+    fn mlock_secret_memory(&self) {
+        let ptr = &*self.0 as *const Fr as *mut u8;
+        let n_bytes = size_of_val(&*self.0);
+        unsafe {
+            if !mlock(ptr, n_bytes) {
+                println!("SK MLOCK FAILED!");
```
vkomenda: Could this be a `panic!`? Or an `error!`? Or a `Result<(), errno::Errno>`? Arguably a `panic!` would be too strong. 

A `Result` would be better, in my opinion. Then all callers would need to handle the `Result`. Currently there is no special handling for situations such as when `Poly::from` is unable to `mlock` secret memory. Returning and handling the `Result` would allow this handling to take place.

However it would be perfectly fine to just log using `error!` if all you eventually do to handle the result would just be logging.

In Unix, `mlock` has many possible error codes. `memsec::mlock` returns a simple `bool` but the error code might still be recoverable using https://docs.rs/errno/0.2.4/errno/. If so, the panic or error message can include the description of the error, or at least the numeric error code.

src/crypto/mod.rs
```diff
@@ -156,47 +170,107 @@ impl fmt::Debug for SignatureShare {
...
+        }
+    }
+
+    fn munlock_secret_memory(&self) {
+        let ptr = &*self.0 as *const Fr as *mut u8;
+        let n_bytes = size_of_val(&*self.0);
         unsafe {
-            write_volatile(ptr, SecretKey::default());
+            if !munlock(ptr, n_bytes) {
+                println!("SK MUNLOCK FAILED!");
```
vkomenda: This could be a `panic!` or an `error!` or an `Err`.

src/crypto/poly.rs
```diff
@@ -192,27 +217,51 @@ impl<'a> ops::Mul<u64> for Poly {
...
+    }
+}
+
+impl ContainsSecret for Poly {
+    fn mlock_secret_memory(&self) {
+        let ptr = self.coeff.as_ptr() as *mut u8;
+        let n_bytes = size_of_val(self.coeff.as_slice());
+        unsafe {
+            if !mlock(ptr, n_bytes) {
+                println!("POLY MLOCK FAILED!");
```
vkomenda: This could be a `panic!` or an `error!` or an `Err`.

src/crypto/poly.rs
```diff
@@ -192,27 +217,51 @@ impl<'a> ops::Mul<u64> for Poly {
...
+
+    fn munlock_secret_memory(&self) {
+        let ptr = self.coeff.as_ptr() as *mut u8;
+        let n_bytes = size_of_val(self.coeff.as_slice());
         unsafe {
-            for i in 0..self.coeff.len() {
-                let ptr = start.offset(i as isize);
-                write_volatile(ptr, Fr::zero());
+            if !munlock(ptr, n_bytes) {
+                println!("POLY MUNLOCK FAILED!");
```
vkomenda: This could be a `panic!` or an `error!` or an `Err`.

src/crypto/poly.rs
```diff
@@ -284,7 +333,20 @@ impl Poly {
...
+    }
+
+    // Removes the `mlock` for zero prime field elements that have been
+    // truncated from the `coeff` vector.
+    fn munlock_truncated(&self, len: usize) {
+        let n_bytes_truncated = len * size_of::<Fr>();
+        unsafe {
+            let ptr = self.coeff.as_ptr().offset(self.coeff.len() as isize) as *mut u8;
+            if !munlock(ptr, n_bytes_truncated) {
+                println!("POLY TRUNCATE MUNLOCK FAILED!");
```
vkomenda: This could be a `panic!` or an `error!` or an `Err`.

src/crypto/poly.rs
```diff
@@ -404,25 +472,71 @@ pub struct BivarPoly {
...
+impl ContainsSecret for BivarPoly {
+    fn mlock_secret_memory(&self) {
+        let ptr = self.coeff.as_ptr() as *mut u8;
+        let n_bytes = size_of_val(self.coeff.as_slice());
         unsafe {
-            for i in 0..self.coeff.len() {
-                let ptr = start.offset(i as isize);
-                write_volatile(ptr, Fr::zero());
+            if !mlock(ptr, n_bytes) {
+                println!("BivarPoly MLOCK FAILED!");
```
vkomenda: This could be a `panic!` or an `error!` or an `Err`.

src/crypto/poly.rs
```diff
@@ -404,25 +472,71 @@ pub struct BivarPoly {
...
             }
         }
     }
+
+    fn munlock_secret_memory(&self) {
+        let ptr = self.coeff.as_ptr() as *mut u8;
+        let n_bytes = size_of_val(self.coeff.as_slice());
+        unsafe {
+            if !munlock(ptr, n_bytes) {
+                println!("BivarPoly MUNLOCK FAILED!");
```
vkomenda: This could be a `panic!` or an `error!` or an `Err`.

c0gent:
I would create new `Crypto` error variants for each kind of error. Return them in `Result`s where possible and panic within `::drop` calls. Add some comments somewhere in the crypto error module for your new error variants describing exactly what they should display when printed and I will add them to my new `failure` errors after we merge this. That way, when you panic with `panic!("{}", err)`, the error will automatically print a friendly message.

Panicking within a drop is the appropriate thing to do, especially for something like this.

c0gent:
So, most or all of those unwraps really need to be error return values. If panicking is necessary for some reason (for example within `drop`) there **must** be very clear and thorough [documentation describing exactly what situations can cause panic](https://doc.rust-lang.org/book/first-edition/documentation.html#special-sections). Error return values should also be documented and described but it's not quite as big of a deal.

Also, in those situations you absolutely must panic, prefer `expect` over `unwrap`.

DrPeterVanNostrand:
@c0gent do you mean that I shouldn't call unwrap in tests? I `panic`ed in `drop`, `clone`, and the various `impl ops::*` that don't have return values.

c0gent:
For tests I'd at least add a simple expect message describing what you were doing. 

For (non-doc) examples I'd heavily lean towards using results with a full error type just as you would in production code (I know the current examples don't do this but they should). [For documentation examples, use `?` (or `.map_err(...)?`) instead of `unwrap`](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark) (note that the example listed there was pre-result main).

For the foreign `impl`s, prefer expect just to give a little bit more context to the user. Also I'd add very descriptive doc comments to the method  within a `# Panic` section (as described in the link from my prev. comment).

src/crypto/error.rs
```diff
@@ -6,5 +6,13 @@ error_chain! {
         DuplicateEntry {
             description("signature shares contain a duplicated index")
         }
+        MlockFailed(desc: String) {
+            description("failed mlock a region of memory")
```
afck: "`failed to`…"? (Also below.)

src/crypto/mod.rs
```diff
@@ -156,47 +171,129 @@ impl fmt::Debug for SignatureShare {
...
+        }
     }
 }
 
 impl Drop for SecretKey {
     fn drop(&mut self) {
-        let ptr = self as *mut Self;
+        self.zero_secret_memory();
+        if let Err(e) = self.munlock_secret_memory() {
+            panic!("{}", e);
```
afck: Would it make sense to only `print!` if `thread::panicking`? Otherwise we'd always get a "panic while panicking" if munlock fails on two or more instances.

afck:
Sorry, I commented on the wong PR. This can be closed, right?

PR #166, opened 2018-07-25
Replace `chain_error` with `failure`

### LINKED ISSUE #142: TODO: Convert error types from `error_chain` to `failure`
Will do this next week.

c0gent:
### Changes

* Remove `error_chain` and convert errors to `failure` types.
* Add variants for each possible error generation point.
    * Allows the user to programatically determine the precise origin of each error.
    * Obviates the need for backtraces. 

### Other Possibilities

* The verbose Error + ErrorKind (E + EK) failure pattern has been used for 'top-level' error-returning modules (`*honey_badger`) as a convenience to the user (it includes backtraces automatically). These are not strictly necessary and the simpler (Error only) pattern could be used instead. The E + EK pattern, though more cumbersome, is also more future-proof and versatile.

### Unfinished

- [ ] Come up with more meaningful / permanent names and display messages for the following error variants (also see [this comment](https://github.com/poanetwork/hbbft/pull/166#issuecomment-409027594)):
  * `agreement::Error::HandleCoinCommonCoin`, `TryFinishConfRoundCommonCoin`
  * `broadcast::Error::CodingNewReedSolomon`, `CodingEncodeReedSolomon`, 
`CodingReconstructShardsReedSolomon`, `CodingReconstructShardsTrivialReedSolomon`
  * `common_coin::Error::CombineAndVerifySigCrypto`
  * `common_subset::Error::NewAgreement`, `ProcessAgreementAgreement0`, `ProcessAgreementAgreement1`, `NewBroadcast`, `ProcessBroadcastBroadcast`
  * `honey_badger::Error::ProposeBincode`, `ProposeCommonSubset0`, `ProposeCommonSubset1`, `HandleCommonMessageCommonSubset0`, `HandleCommonMessageCommonSubset1`
  * `dynamic_honey_badger::Error::SendTransactionBincode`, `VerifySignatureBincode`, `SignVoteForBincode`, `ValidateBincode`, `ProposeHoneyBadger`, `HandleHoneyBadgerMessageHoneyBadger`
- [x] Simple documentation on each type and variant (pending renaming of variants, etc.).
- [ ] Decide whether or not to apply the E + EK (+ Context) pattern to more error types. The advantage to this pattern is mainly automatic creation of backtraces. (Conversions to E + EK errors is another benefit but I argue that we should stray from the norm and not use on non-contextual conversions.)
- [x] Squash commits.

Closes #142

### DIFF
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -77,20 +77,21 @@ use common_coin::{self, CommonCoin, CommonCoinMessage};
 use fault_log::{Fault, FaultKind};
 use messaging::{self, DistAlgorithm, NetworkInfo, Target};
 
-error_chain!{
-    links {
-        CommonCoin(common_coin::Error, common_coin::ErrorKind);
-    }
-
-    errors {
-        UnknownProposer {
-            description("unknown proposer")
-        }
-        InputNotAccepted {
-            description("input not accepted")
-        }
-    }
+/// An agreement error.
+#[derive(Clone, Eq, PartialEq, Debug, Fail)]
+pub enum Error {
+    #[fail(display = "HandleCoinCommonCoin error: {}", _0)]
+    HandleCoinCommonCoin(common_coin::Error),
+    #[fail(display = "TryFinishConfRoundCommonCoin error: {}", _0)]
+    TryFinishConfRoundCommonCoin(common_coin::Error),
+    #[fail(display = "Unknown proposer")]
+    UnknownProposer,
+    #[fail(display = "Input not accepted")]
+    InputNotAccepted,
 }
+
+/// An agreement result.
+pub type Result<T> = ::std::result::Result<T, Error>;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub enum AgreementContent {
@@ -267,7 +268,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         proposer_id: NodeUid,
     ) -> Result<Self> {
         if !netinfo.is_node_validator(&proposer_id) {
-            return Err(ErrorKind::UnknownProposer.into());
+            return Err(Error::UnknownProposer);
         }
         Ok(Agreement {
             netinfo,
@@ -291,7 +292,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
     /// Sets the input value for agreement.
     fn set_input(&mut self, input: bool) -> Result<Step<NodeUid>> {
         if self.epoch != 0 || self.estimated.is_some() {
-            return Err(ErrorKind::InputNotAccepted.into());
+            return Err(Error::InputNotAccepted);
         }
         // Set the initial estimated value to the input value.
         self.estimated = Some(input);
@@ -411,9 +412,9 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
     ) -> Result<Step<NodeUid>> {
         let coin_step = match self.coin_state {
             CoinState::Decided(_) => return Ok(Step::default()), // Coin value is already decided.
-            CoinState::InProgress(ref mut common_coin) => {
-                common_coin.handle_message(sender_id, msg)?
-            }
+            CoinState::InProgress(ref mut common_coin) => common_coin
+                .handle_message(sender_id, msg)
+                .map_err(Error::HandleCoinCommonCoin)?,
         };
         self.on_coin_step(coin_step)
     }
@@ -571,7 +572,9 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         // Invoke the common coin.
         let coin_step = match self.coin_state {
             CoinState::Decided(_) => return Ok(Step::default()), // Coin has already decided.
-            CoinState::InProgress(ref mut common_coin) => common_coin.input(())?,
+            CoinState::InProgress(ref mut common_coin) => common_coin
+                .input(())
+                .map_err(Error::TryFinishConfRoundCommonCoin)?,
         };
         let mut step = self.on_coin_step(coin_step)?;
         step.extend(self.try_update_epoch()?);
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -20,7 +20,7 @@ use pairing::{CurveAffine, CurveProjective, Engine, Field};
 use rand::{ChaChaRng, OsRng, Rng, SeedableRng};
 use ring::digest;
 
-use self::error::{ErrorKind, Result};
+use self::error::{Error, Result};
 use self::into_fr::IntoFr;
 use self::poly::{Commitment, Poly};
 use fmt::HexBytes;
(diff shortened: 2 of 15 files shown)


src/crypto/mod.rs
```diff
@@ -457,13 +457,13 @@ where
         .map(|(i, sample)| (from_repr_plus_1::<C::Scalar>(i.clone().into()), sample))
         .collect();
     if samples.len() < t {
-        return Err(ErrorKind::NotEnoughShares.into());
+        return Err(Error::NotEnoughShares.into());
```
afck: This `into()` is probably redundant now?

src/crypto/mod.rs
```diff
@@ -457,13 +457,13 @@ where
...
     }
     let mut result = C::zero();
     let mut indexes = Vec::new();
     for (x, sample) in samples.iter().take(t) {
         if indexes.contains(x) {
-            return Err(ErrorKind::DuplicateEntry.into());
+            return Err(Error::DuplicateEntry.into());
```
afck: …and this one, too.

src/agreement/mod.rs
```diff
@@ -481,7 +482,7 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
         sender_id: &NodeUid,
         msg: CommonCoinMessage,
     ) -> Result<FaultLog<NodeUid>> {
-        let coin_step = self.common_coin.handle_message(sender_id, msg)?;
+        let coin_step = self.common_coin.handle_message(sender_id, msg).map_err(Error::HandleCoinCommonCoin)?;
```
vkomenda: `.map_err(Error::HandleCoinCommonCoin)`
could be
`.map_err(|e| Error::CommonCoin(e, line!()))`.

src/dynamic_honey_badger/error.rs
```diff
@@ -1,17 +1,63 @@
...
+    fn from(kind: ErrorKind) -> Error {
+        Error { inner: Context::new(kind) }
     }
+}
 
-    foreign_links {
-        Bincode(Box<bincode::ErrorKind>);
+impl From<Context<ErrorKind>> for Error {
+    fn from(inner: Context<ErrorKind>) -> Error {
+        Error { inner: inner }
```
vkomenda: See the clippy lint.

src/honey_badger.rs
```diff
@@ -40,20 +41,64 @@ use crypto::{Ciphertext, DecryptionShare};
...
+    fn from(kind: ErrorKind) -> Error {
+        Error { inner: Context::new(kind) }
     }
+}
 
-    foreign_links {
-        Bincode(Box<bincode::ErrorKind>);
+impl From<Context<ErrorKind>> for Error {
+    fn from(inner: Context<ErrorKind>) -> Error {
+        Error { inner: inner }
```
vkomenda: See the clippy lint.

src/queueing_honey_badger.rs
```diff
@@ -34,12 +36,57 @@ use transaction_queue::TransactionQueue;
...
+
+impl From<ErrorKind> for Error {
+    fn from(kind: ErrorKind) -> Error {
+        Error { inner: Context::new(kind) }
+    }
+}
+
+impl From<Context<ErrorKind>> for Error {
+    fn from(inner: Context<ErrorKind>) -> Error {
+        Error { inner: inner }
```
vkomenda: See the clippy lint.

mbr:
I think this is a good step in the right direction, if I remember correctly, Rust consensus moved from handwritten `Error`s -> `quickerror!` -> `error_chain!` to `failure`.

While I happily skipped the whole `error_chain!` phase, I am not as firm with `failure` as I'd like to be. However, there's two things I'd want to mention:

* I am always loath to introduce `failure` at all, if hand-writing errors is still an option, especially in small libraries. I believe this concern is addressed by failure itself though and should not be an issue. And `hbbft` is anything but a small library, bordering on an application, so I am guessing we are good here? My judgement is also a bit colored by doing embedded and bare metal programming, where allocating or passing extra info is often not desirable due to memory constraints.
* At the moment, I am not convinced that a construction like `ErrorKind::ProposeCommonSubset0` and `ErrorKind::ProposeCommonSubset1` is the way to go, as it binds the interface (more precisely, the type(s) of errors returned) very tightly to the implementation (unless it's just a naming issue). To distinguish between errors of the same kind at different occasions, should we not use things like `failure::ResultExt::with_context` or some other mechanism?

An analogous problem: If given the task of writing a CSV-table naive with no extra-libraries into a text file, what kind of error should we expect? Would it be `Io(io::Error)` (all `io::Error`s in a single variant), `IoHeader(io::Error), IoBody(io::Error)` (this distinguishes between the two call sites, same information content as a backtrace) or `IoHeader(io::Error), IoRow(usize, io::Error)` (names the exact row which failed, different information than backtrace).

I am arguing here that for simplicity, I would prefer the first approach. This also enables the implementor to write everything to a string first and then to a file, causing only a single call to `write_all`, resulting in a single site for an `IoError`. It still is paramount to have the contextual information about the row the error occured in, which might need to transported through another channel (such as a context).

I do not have a perfect answer for this yet. 

I might change my mind after further study; but would it be possible to hold off on this for 1-2 extra days, to give me a chance to sort this out?

src/agreement/mod.rs
```diff
@@ -78,21 +78,22 @@ use common_coin::{self, CommonCoin, CommonCoinMessage};
...
-        UnknownProposer {
-            description("unknown proposer")
-        }
-        InputNotAccepted {
-            description("input not accepted")
-        }
-    }
+#[derive(Clone, Eq, PartialEq, Debug, Fail)]
+pub enum Error {
+    #[fail(display = "HandleCoinCommonCoin error: {}", _0)]
```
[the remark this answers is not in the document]
c0gent: Well those names on `...CommonCoin`, `...ReedSolomon` errors are meant to be temporary and really need to have some sort of meaningful explaination of why the error occurred.

As for reducing the boilerplate using macros, This is easily possible but as soon as you modify an errors API, to add a new conversion or other method, the error impl's diverge and those macros quickly become useless.

src/crypto/mod.rs
c0gent: Yup, missed those.

src/queueing_honey_badger.rs
c0gent: It's probably going to say don't `inner: inner`. I don't use clippy :P

src/agreement/mod.rs
c0gent: The inclusion of the line information is redundant though. The error variants themselves include all of the information needed to trace the precise point of origin both programatically and via display/debug messages. Also, as Mark alluded to, including this kind of information comes at a cost.

c0gent:
Marc, great comments. The reason I have designed the errors this way is so that the origin of each error can be uniquely identified without relying on backtraces, line macros, etc. It makes it possible to be as thorough as possible and to have as much information as possible when handling errors.

Having a single variant for all errors of a certain type, for example having an `Error::Io(std::io::Error)` variant for all IO errors and providing a `From` impl, I now consider inadequate and even somewhat lazy. I've come to realize the many advantages of more precise error reporting when it comes time to handling errors and debugging. If the extra level of precision is unnecessary, a user can simply mash error variants together when `match`ing (with `|`).

This design is ideal for bare metal / embedded because no backtraces or other allocated data need be included with errors. Also, all errors are `Send + Sync + 'static` so they can quickly and easily be passed between threads / tasks.

As to the naming issue, the names of the numbered error variants are meant to be temporary. They need to more meaningful and still each be unique. Even if they do end up including the name of the origin method: if that method is renamed it's an API breakage anyway so it doesn't really matter if the error variant is renamed too.

c0gent:
Rather than try to come up with meaningful names for things right now I propose we leave the ugly error variant names in place and change them (& combine them) as/when we have better ideas or they become a big enough annoyance. With the API in such a state of flux (as it should be), it's unlikely that any naming we settle on now will last very long.

PR #169, opened 2018-07-26
Avoids cloning of Agreement::received_term

vkomenda:
I wonder what you think about this small change.

### DIFF
--- a/src/agreement/mod.rs
+++ b/src/agreement/mod.rs
@@ -505,8 +505,13 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
 
         self.estimated = Some(b);
         let mut step = self.send_bval(b)?;
-        for (sender_id, b) in self.received_term.clone() {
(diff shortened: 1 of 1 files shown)


src/agreement/mod.rs
```diff
@@ -505,8 +505,13 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
         self.estimated = Some(b);
         let mut step = self.send_bval(b)?;
-        for (sender_id, b) in self.received_term.clone() {
+        // Create a temporary map of received TERM messages to avoid a second mutable access to
+        // `self`.
+        let received_term = replace(&mut self.received_term, BTreeMap::new());
+        for (sender_id, b) in received_term {
             step.extend(self.handle_term(&sender_id, b)?);
+            // Place the TERM message back.
+            self.received_term.insert(sender_id, b);
```
afck: That line can probably be removed; it happens in `handle_term` anyway.

PR #171, opened 2018-07-30
Fix agreement bugs, and refactor a few fields.

afck:
* Don't drop `Term` messages from previous epochs. They are still
  relevant for all future epochs.
* Restructure some fields to avoid unnecessary iteration and counting.
* Simplify the fields related to the common coin.
* Reorder the methods, so that the message handlers are all in one
  place.
* Handle the case where the coin value arrives before the required
  number of `Conf` messages.

### DIFF
--- a/src/agreement/bin_values.rs
+++ b/src/agreement/bin_values.rs
@@ -1,5 +1,6 @@
 use std::iter::FromIterator;
 use std::mem::replace;
+use std::vec;
 
 /// A lattice-valued description of the state of `bin_values`, essentially the same as the set of
 /// subsets of `bool`.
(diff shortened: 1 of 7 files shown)


src/agreement/bin_values.rs
```diff
@@ -109,3 +110,17 @@ impl FromIterator<BinValues> for BinValues {
         v
     }
 }
+
+impl IntoIterator for BinValues {
+    type Item = bool;
+    type IntoIter = vec::IntoIter<bool>;
```
vkomenda: I propose not to go via `vec` but use statically allocated slices instead:
https://play.rust-lang.org/?gist=17bf2b390e687054a58840d9604acd09&version=stable&mode=debug&edition=2015

src/agreement/mod.rs
```diff
@@ -467,75 +409,138 @@ impl<NodeUid: Clone + Debug + Ord> Agreement<NodeUid> {
...
+            }
+            CoinState::InProgress(_) => self.send_conf(aux_vals), // Start the `Conf` message round.
+        }
+    }
+
+    /// Multicasts a `BVal(b)` message, and handles it.
+    fn send_bval(&mut self, b: bool) -> Result<Step<NodeUid>> {
+        // Record the value `b` as sent. If it was already there, don't send it again.
+        if !self.sent_bval.insert(b) {
+            return Ok(Step::default());
```
vkomenda: This can happen only in case of programming error? Maybe insert a critical error here?

afck: No, it happens e.g. when we input `true`, and then receive the _f + 1_ st `BVal(true)`. (Or vice versa, I think.)

PR #172, opened 2018-07-30
crypto module moved into its own crate

### LINKED ISSUE #164: Move the crypto module into its own crate
The crypto module
* has a different commit dynamic to the rest of the `hbbft` crate,
* is related to the main code through an interface,
* has it's own CI tests that add up to the CI tests of `hbbft` and don't have to be run on every commit to the rest of `hbbft`.

Separating it into its own crate can also facilitate the research on other types of cryptography (#54) in `hbbft` because other crates could be designed to use the same interface.

Crypto-related issues such as #89 would move with the module as well.

vkomenda:
Close #164 

Also please check the new crypto crate and send issues there at https://github.com/poanetwork/threshold_crypto/issues


vkomenda:
Updated to `failure`, removed `protobuf` support, and made the matching changes in https://github.com/poanetwork/threshold_crypto.

PR #174, opened 2018-07-30
Represent BinValues as a u8; rename to BoolSet.

afck:


afck:
Maybe we should also rename it to `BoolSet`?

PR #177, opened 2018-07-31
Remove protobuf support.

### LINKED ISSUE #176: Remove protobuf support.
It is currently only partly implemented, and not required in addition to serde.

afck:
Close #176.


src/crypto/mod.rs
```diff
@@ -5,8 +5,6 @@
 pub mod error;
 mod into_fr;
 pub mod poly;
-#[cfg(feature = "serialization-protobuf")]
```
vkomenda: Do you think `serialization-protobuf` might stay in the `threshold_crypto` crate?

afck: I'd vote for removing it there, too, but if we make sure the implementation is complete and functional, I'm happy with keeping it, too.

(In this PR, I guess, I can remove it, since the whole module will be moved out either way?)

vkomenda: Please keep the change here. I'll take care of in in the other crate. We have to remove `protobuf_impl.rs` as well.

afck: Right, good catch! I totally missed that!

Issue #121, opened 2018-07-11
Split `honey_badger` into smaller modules.


afck:
The decryption part involves many lines of code in `honey_badger.rs`. I wonder whether this could be split out into a private submodule. We could write a kind of collection in which you can put the ciphertext and decryption shares and which would automatically return the decrypted item as soon as it has enough information. If both steps (Subset and Decryption) are separate modules, `honey_badger.rs` should be very simple and easy to read.

afck:
In addition to separating the Subset and Decryption step, we could also factor out a single epoch. So:

`HoneyBadger` would simply dispatch and filter messages by epoch, and run a `HoneyBadgerEpoch` instance for each of them. A `HoneyBadgerEpoch` would first run one `CommonSubset`, and then one `Decryption` instance.

c0gent:
I think this is a great idea. Splitting things up creates a bit more work but is helpful for keeping concerns separate and making drawing clearer boundaries. I sometimes separate things based on shared import grouping. This occasionally highlights structural patterns that I hadn't otherwise considered. Obviously there are limits to all of this and following the principles of laziness tends to be a good guide!

PR #178, opened 2018-08-01
Split honey_badger into submodules.

afck:
See #121.
This doesn't change any code yet, it only splits up the module (and makes a few fields `pub(super)`).


afck:
Not sure about `src/honey_badger/honey_badger_impl.rs`… I think I'd actually prefer `src/honey_badger/impl.rs` — what do you think?

afck:
I changed it to `honey_badger.rs`.

PR #179, opened 2018-08-01
Change the QHB criterion for moving on to the next epoch.

### LINKED ISSUE #147: What should `QueueingHoneyBadger` do with a single validator?
Currently `QueueingHoneyBadger` always immediately initiates the next epoch as soon as the previous one is output. That doesn't work for a single validator, since it always outputs immediately in every epoch, so it would just output empty batches in an infinite loop. We need to decide on the semantics here:

1. Don't allow single-validator networks. Refuse to construct an instance with a single validator, and terminate if one of two nodes leaves the network.
  _That makes it a bit cumbersome to start up a network. Starting with a single validator and adding more step-by-step would be convenient._

2. Make the user explicitly start the next epoch. Then they can determine a minimal batch time, and delay the epoch, or wait for another node or a transaction.
  _It might feel a bit unnatural if you explicitly have to trigger that every time. Also, it makes it easy to make the same mistake and just always trigger after output → infinite loop with a single validator._

3. Automatically start the new epoch, but only if there are either transactions in the queue or key generation messages pending.
  _That would block a network, though, if e.g. half the nodes have plenty of pending transactions and half of them don't have any._

4. Only automatically start the next epoch if there are either transactions in the queue, or key generation messages pending, or at least _k_ proposals from other nodes… for _k = f + 1_? Or _k = 1_?
  _That would still block if only k - 1 nodes have pending transactions. And if k ≤ f, it would allow an adversary to generate lots of empty blocks. But maybe that's not too bad? And k could be a configuration option._

afck:
`QueueingHoneyBadger` now waits after an output, and only makes its
proposal for the next epoch when:

* there are pending transactions in the queue,
* there are pending key generation or vote messages, or
* _f + 1_ other validators have already made their proposal.

This rule should work well for small networks: With 1 - 3 nodes, it will
produce a new batch whenever at least one of them has transactions to
contribute. In larger networks, it prevents an adversary controlling _f_
nodes from producing lots of empty epochs.

An exception is made for a currently joining validator: We will commit
up to _(N + 1)² + 1_ key generation messages for them, which is the
maximum number a correct node will send.

Closes #147.

### DIFF
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -307,14 +307,14 @@ where
         }
     }
 
-    /// Handles a queued message in one of the nodes with the earliest timestamp.
-    pub fn step(&mut self) -> NodeUid {
+    /// Handles a queued message in one of the nodes with the earliest timestamp, if any. Returns
+    /// the recipient's ID.
+    pub fn step(&mut self) -> Option<NodeUid> {
         let min_time = self
             .nodes
             .values()
             .filter_map(TestNode::next_event_time)
-            .min()
-            .expect("no more messages in queue");
+            .min()?;
         let min_ids: Vec<NodeUid> = self
             .nodes
             .iter()
@@ -328,7 +328,7 @@ where
             node.out_queue.drain(..).collect()
         };
         self.dispatch_messages(msgs);
-        next_id
+        Some(next_id)
     }
 
     /// Returns the number of messages that have been handled so far.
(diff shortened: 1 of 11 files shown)


c0gent:
I like how this sounds. I'll try it out in a bit :)

c0gent:
Just to clarify, this PR is ready to be merged now imo.

PR #181, opened 2018-08-01
Update links

andogro:
Updated HB and DynamicHB subdirectory links and added QueueingHB to algorithm list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,11 @@ In addition to **validators**, the algorithms support **observers**: These don't
 
 ## Algorithms
 
-- [x] **[Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/honey_badger.rs):** Each node inputs transactions. The protocol outputs a sequence of batches of transactions.
+- [x] **[Honey Badger](https://github.com/poanetwork/hbbft/blob/master/src/honey_badger/mod.rs):** Each node inputs transactions. The protocol outputs a sequence of batches of transactions.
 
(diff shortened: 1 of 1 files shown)


andogro:
Wondering about a few things:
1) Should I mention and link to the new crypto crate?
2) Any additions to current TODOs- i.e. should I include Parity integration?
3) I am planning to move protocol modifications to the github wiki when I add the error handling overview.

c0gent:
> 1) Should I mention and link to the new crypto crate?

I'd say a mention would be good. Perhaps a bit about what role it plays overall.

> 3) I am planning to move protocol modifications to the github wiki when I add the error handling overview.

Good idea.

As an aside, speaking of the Wiki, I really think we should have a proper, dedicated, informational website for users. There are a number of limitations with github wikis, for example searching and navigation. I'm sure this is probably something you've already considered. I'd be interested to hear more about our strategy/plans for 'user facing' information. Could we could carry on this conversation elsewhere? Please send me some links if this is already being discussed elsewhere :)

c0gent:
Ah... and as for Parity integration, I'll have some written material about that for you next week.

afck:
Sorry for the broken links, that was my bad!
I wonder whether there's an easy way to check these in CI.

c0gent:
@andogro, can you please rebase this to remove the merge commit? Let me know if that makes sense or if you need any help :)

andogro:
@c0gent should be good to go - let me know if not. Thanks!

PR #187, opened 2018-08-02
Add common supertraits and rename related type parameters.

c0gent:
* Add the `Contribution`, `NodeUidT`, and `Message` supertraits.
* Rename type parameters:
  * `Tx` -> `T` or `C`
  * `NodeUid` -> `N`

### Open Issues

* I'm not happy with the naming of `NodeUidT`. It technically *could* just be called `NodeUid`, and i'm learning towards just renaming it back to that, but it does share a name with `DistAlgorithm::NodeUid`. This is allowed but seems like it could be slightly confusing. Maybe not.
* Also the `traits` module name isn't wonderful.

As usual, It takes as more effort to name things as it does to code them...

@mbr, @afck, @drPeterVanNostrand, @vkomenda: Please let me know what you guys think.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,3 +135,21 @@ pub mod messaging;
 pub mod queueing_honey_badger;
 pub mod sync_key_gen;
 pub mod transaction_queue;
+
+/// Common supertraits.
+pub mod traits {
+    use std::fmt::Debug;
+    use std::hash::Hash;
+
+    /// A transaction, user message, etc.
+    pub trait Contribution: Eq + Debug + Hash + Send + Sync {}
+    impl<C> Contribution for C where C: Eq + Debug + Hash + Send + Sync {}
+
+    /// A peer node's unique identifier.
+    pub trait NodeUidT: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
+    impl<N> NodeUidT for N where N: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
+
+    /// Messages.
+    pub trait Message: Debug + Send + Sync {}
+    impl<M> Message for M where M: Debug + Send + Sync {}
+}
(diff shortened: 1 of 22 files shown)


c0gent:
Also, @mbr, please try out these changes against your test code. It might help with the trait issues you were having.

src/lib.rs
```diff
@@ -135,3 +135,21 @@ pub mod messaging;
 pub mod queueing_honey_badger;
 pub mod sync_key_gen;
 pub mod transaction_queue;
+
+/// Common supertraits.
+pub mod traits {
+    use std::fmt::Debug;
+    use std::hash::Hash;
+
+    /// A transaction, user message, etc.
```
mbr: I would like to see a longer comment here, documenting why the trait bounds were chosen as they are, since this is such an integral piece.

c0gent: Hmm, I usually don't document why I chose the specific bounds of supertraits. I hear you 100% about more thorough documentation but I'm not sure what I would say other than "We needed all this stuff at some point, somewhere".

I'm at a loss. What do you think I should write?

mbr: I assume there was a specific reason for each of these to be added at some point? I will admit that the current set seems fairly innocuous though.

mbr:
> Also, @mbr, please try out these changes against your test code. It might help with the trait issues you were having.

Works well. The only thing I keep adding is `D::Message: Clone`, but I am assuming that's probably not a great thing to have as a requirement on all messages?

c0gent:
@mbr, I considered that and I can't think of a reason not to add a bound for it. Generally speaking, one reason not to require `Clone` is if a `Message` or `Contribution` ever needed to be a shared (mutable) reference.

afck:
If you're okay with it, I'd prefer merging this as soon as possible: Renaming `NodeUidT` or adding comments won't create nearly as many merge conflicts and can be done in a separate PR.

PR #189, opened 2018-08-03
Broken PR.

mbr:
DO NOT MERGE.

This PR is intentionally broken to see if merge-only-if-CI-succeeds is enabled.


mbr:
I have merge button I can click, so it seems it's not enabled yet.

c0gent:
Indeed.

PR #194, opened 2018-08-06
Refinements to SyncKeyGen error and fault handling

c0gent:
* Add a new `AckMessageFault` type.
* Add new `sync_key_gen::Error` variants and remove `From` impl.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ reed-solomon-erasure = "3.1.0"
 ring = "^0.12"
 serde = "1.0.55"
 serde_derive = "1.0.55"
-threshold_crypto = { git = "https://github.com/poanetwork/threshold_crypto" }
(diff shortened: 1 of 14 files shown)


c0gent:
If you like these changes you can approve and then click `Rebase and merge`. The commit will automatically be added to your branch/PR.

DrPeterVanNostrand:
Ugg, idk why your tests are failing because of `mlock`. I'll look into it.

c0gent:
Ok and I changed two of the `error!` logging macros back to `debug!` since the duplicate messages are normal.

c0gent:
By the way, the tests pass for me. Because they fail on Travis it means tests are not correct and need to be fixed (in addition to whatever fixes you make to the `mlock` code). I'm sure you're probably already on top of it though :)

PR #197, opened 2018-08-07
Deploy key

natlg:


c0gent:
Thanks :)

If anyone is curious, this key will be used for generating rustdoc (when I get around to it).

I'll merge it now and we can make sure everything is working.

PR #200, opened 2018-08-08
Replace merkle.rs with custom implementation.

afck:
This also contains #199 and splits up the broadcast module.


c0gent:
Very nice. I'll test it out a little later. Do you think it would be worth putting this into its own crate?

afck:
Possibly. But it would require much more work to add more functionality and documentation and make it more generally useful. And there are already several Merkle tree crates.

mbr:
While [XKCD 927](https://xkcd.com/927/) applies, we did try though, right?

PR #207, opened 2018-08-10
`SIGILL` on travis

mbr:
~~Do not pull this yet and be very, very quiet; I'm trying to tease out a `SIGILL`.~~
I cut it short and created one, to ensure it works: https://travis-ci.org/poanetwork/hbbft/builds/415424600

The `after_failure` block will print the last 250 lines of the the syslog, which should contain information about the `SIGILL`.

Unfortunately it also adds a `sudo` dependency; travis analyses CI configurations beforehand and runs those that require `sudo` on VMs instead of containers. As a result there's a ~40 second startup time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,12 @@ before_install:
   - rustup component add --toolchain=${RUST_NEXT} rustfmt-preview clippy-preview
   # Some symlinking is still necessary for clippy to function properly.
   - ln -sf ${HOME}/.rustup/toolchains/${RUST_NEXT}-x86_64-unknown-linux-gnu/bin/clippy-driver ${HOME}/.rustup/toolchains/${RUST_NEXT}-x86_64-unknown-linux-gnu/bin/cargo-clippy $HOME/.cargo/bin/
+# after_failure:
+#   # Outputs the syslog after a failed build, e.g. to debug `SIGILL` occurrences.
+#   # Unfortunately this is likely to disable container-based travis images,
+#   # causing a CI slowdown, so this option is commented out by default. It can
+#   # be enabled per-branch to debug issues.
+#   - sudo tail -n 250 /var/log/syslog
 env:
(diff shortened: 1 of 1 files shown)


mbr:
We should probably remove the `after_failure` block once we are sure to have fixed all `SIGILL`s.

vkomenda:
I just had an idea to remove the `sudo` dependency. Is `journalctl` available? The last hour's errors in reverse order might be retrieved like so:
```
journalctl -r -S -1h
```

.travis.yml
```diff
@@ -18,6 +18,11 @@ before_install:
   - rustup component add --toolchain=${RUST_NEXT} rustfmt-preview clippy-preview
   # Some symlinking is still necessary for clippy to function properly.
   - ln -sf ${HOME}/.rustup/toolchains/${RUST_NEXT}-x86_64-unknown-linux-gnu/bin/clippy-driver ${HOME}/.rustup/toolchains/${RUST_NEXT}-x86_64-unknown-linux-gnu/bin/cargo-clippy $HOME/.cargo/bin/
+after_failure:
+  # Outputs the syslog after a failed build, e.g. to debug `SIGILL` occurences.
```
afck: "`occurrences`"

mbr:
> I just had an idea to remove the sudo dependency. Is journalctl available?

That's a pretty good idea, unfortunately, travis is on Ubuntu 14(!).04. `systemd` was introduced for 15.04, I think.

You can see for yourself here: https://travis-ci.org/poanetwork/hbbft/builds/419541247.

We need to chose now: Omit this functionality or go back to sudo builds? Or wait until 2020, when travis will have to update their build environment? =/

mbr:
After discussion, this should be the final form; a commented out section that we can use for debugging if need be. Anything beyond is blocked by travis being the way it is.

PR #211, opened 2018-08-23
Use latest `pairing` crate and accordingly updated `threshold_crypto`.

mbr:
The tests will fail until https://github.com/poanetwork/threshold_crypto/pull/12 is merged.

As with `threshold_crypto`, the changes are fairly small. We should lobby for an updated `pairing` release regardless.


afck:
As discussed, let's stick to 0.14.2 until 0.15.0 is released.

PR #216, opened 2018-08-27
New test network implementation

mbr:
This is part of a four-merge-effort, the first being the new test networking code, to be followed by proper proptests, model based testing and mocktography (not necessarily in that order).

When merging, this should be probably be squashed. I don't expect a lot of conflicts, since only a single line was changed in the original code, everything else is implemented in new files.

I'm still very much open to feedback, suggestions, improvements, etc.

For reading, I suggest starting with the `README.md`, then `net_dynamic_hb.rs` and finally the `tests/net/mod.rs` module.


tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+```rust
+net.broadcast_input(input).expect("algorithm failed");
+```
+
+### Cranking the network
+
+The network advances through the `crank()` function, on every call
+
+1. the adversary is given a chance to re-order the message queue,
```
vkomenda: Actually it seems that `Adversary::pre_crank` allows to mutate any part of the `VirtualNet`. So I think this sentence is not complete.

If you wanted to restrict the amount of state the adversary is allowed to mutate, I think it's better to do that by passing that exact amount of data to `pre_crank`. But having an all-powerful adversary has its positive sides as well. I'm fine with just updating this list item.

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+1. the next message in the queue is delivered to its destination node (if the node is non-faulty) or the adversary (if the node is faulty),
+1. all messages from the resulting step are queued again,
```
vkomenda: Is "again" necessary?

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+```rust
+// Create a network of 10 nodes, out of which 3 are faulty.
+let mut net = NetBuilder::new(0..10)
+    .num_faulty(3)
+    .using(move |node| { DynamicHoneyBadger::builder().build(node.netinfo) })
+    .build()
+    .expect("could not construct test network");
+```
+
+Algorithms that return a `Step` upon construction should use `using_step` instead.
```
andogro: Can you clarify the algorithms that require using_step and provide a 1 line example?

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+}
+```
+
+### Inspecting the network
+
+In addition to the returned `Step`s, the network and nodes can be queried through various methods: `VirtualNet::{nodes, faulty_nodes, correct_nodes, get, get_mut}`.
+
+### Adversaries
+
+Adversaries can be introduced through the `.adversary` method on the constructor and are expected to implement the `net::adversary::Adversay` trait. Generic adversaries are available in the same module, while algorithm-specific ones should live next to each test case.
```
andogro: net::adversary::Adversay - assuming this should be Adversary?

tests/net/err.rs
```diff
@@ -0,0 +1,111 @@
+//! Test network errors
+//!
+//! The error implementation for the test networking code is slightly involved and has been factored
```
andogro: Would this generally live elsewhere? If not, I don't think you need this explanation.

tests/net/err.rs
```diff
@@ -0,0 +1,111 @@
...
+use super::NetMessage;
+
+/// Network crank error.
+///
+/// Errors resulting from processing a single message ("cranking").
+pub enum CrankError<D>
+where
+    D: DistAlgorithm,
+{
+    /// The algorithm ran by the node produced a `DistAlgorithm::Error` while processing input.
```
andogro: ~~ran~~ run

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
+//! A test network.
+//!
+//! Test networks simulate a real networking that includes an adversary as well as the plumbing to
```
andogro: networking **environment** that

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+        &mut self.algorithm
+    }
+
+    /// Check whether or not node is marked faulty.
+    #[inline]
+    pub fn is_faulty(&self) -> bool {
+        self.is_faulty
+    }
+
+    /// Get nodes ID.
```
andogro: Get **node's** ID

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+    ///
+    /// Any output made by a node is captured by the node for easy comparison.
+    #[inline]
+    pub fn outputs(&self) -> &[D::Output] {
+        self.outputs.as_slice()
+    }
+}
+
+/// A network message on the virtual network.
+// Note: We do not use `messaging::TargetedMessage` and `messaging::SourceMessage` here, since we
```
andogro: since ~~we~~

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+/// Mapping from node IDs to actual node instances.
+pub type NodeMap<D> = collections::BTreeMap<<D as DistAlgorithm>::NodeUid, Node<D>>;
+
+/// A virtual network message tied to a distributed algorithm.
+pub type NetMessage<D> =
+    NetworkMessage<<D as DistAlgorithm>::Message, <D as DistAlgorithm>::NodeUid>;
+
+/// Process a step.
+///
+/// Expands every message in the step by turning all broadcast-messages into peer-to-peer messages,
```
andogro: broadcast messages

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+        .expect("Trying to process a step with non-existing node ID")
+        .outputs
+        .extend(step.output.iter().cloned());
+
+    message_count
+}
+
+/// New network node construction information.
+///
+/// Helper structure passed to node constructors when constructing virtual networks.
```
andogro: when ~~constructing~~ building.... just a suggestion to avoid too many constructions!

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+For convenience, an iterator interface is also available:
+
+```rust
+for res in net {
+    let (node_id, step) = res.expect("algorithm error");
+    // ...
+}
+```
+
+This has the drawback that access to the network is not available between cranks, since it will be borrowed inside the for-loop. A common workaround is using a while loop instead:
```
andogro: ~~since it will be~~  as it is borrowed

tests/net/adversary.rs
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// The default implementation does not alter the passed network in any way.
+    #[inline]
+    fn pre_crank(&mut self, _net: &mut VirtualNet<D>) {
+        // The default implementation does not alter anything.
+    }
+
+    /// Tamper with a faulty nodes operation.
+    ///
+    /// Faulty nodes can (but are not required to) be run like regular nodes. A core difference is
```
andogro: You can (but are not required to) run faulty nodes like regular nodes.

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Adversaries can be introduced through the `.adversary` method on the constructor and are expected to implement the `net::adversary::Adversay` trait. Generic adversaries are available in the same module, while algorithm-specific ones should live next to each test case.
+
+```rust
+// Missing example.
+```
+
+### Tracing
+
+By default, all network tests write traces of every network message into logfiles, named `net-trace_*.txt`. Each log stores one message per line, in the format of `[SENDER] -> [RECEIVER]: MSG`.
```
DrPeterVanNostrand: Could you put the path/relative-path to where the log files are written?

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+This behavior can be controlled using the `HBBFT_TEST_TRACE` environment variable; if set and equal to `0` or `false`, this functionality is disabled. Tracing is enabled by default.
+
+The `NetBuilder` allows hard-coding the trace setting, any value passed will override environment environment settings:
```
DrPeterVanNostrand: Double "environment"s.

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+pub type NetMessage<D> =
+    NetworkMessage<<D as DistAlgorithm>::Message, <D as DistAlgorithm>::NodeUid>;
+
+/// Process a step.
+///
+/// Expands every message in the step by turning all broadcast-messages into peer-to-peer messages,
+/// and appends them to the network queue. Additionally, saves a copy of each output to the output
+/// buffer of the `sender` node.
+///
+/// At the end, the number of additional messages created by non-faulty nodes will be returned.
```
andogro: ~~will be~~ is

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+            adversary: None,
+            trace: None,
+            crank_limit: None,
+            message_limit: None,
+        }
+    }
+
+    /// Set an adversary.
+    ///
+    /// If not set, the virtual network will be constructed with a `NullAdversary`.
```
andogro: ~~will be~~ is

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+    /// If set, overrides the environment setting of whether or not tracing should be enabled.
+    #[inline]
+    pub fn trace(mut self, trace: bool) -> Self {
+        self.trace = Some(trace);
+        self
+    }
+
+    /// Constructor function (with step).
+    ///
+    /// The constructor function is used to construct each node in the network. Any step returned
```
andogro: ~~will be~~ is

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+### Inspecting the network
+
+In addition to the returned `Step`s, the network and nodes can be queried through various methods: `VirtualNet::{nodes, faulty_nodes, correct_nodes, get, get_mut}`.
+
+### Adversaries
+
+Adversaries can be introduced through the `.adversary` method on the constructor and are expected to implement the `net::adversary::Adversay` trait. Generic adversaries are available in the same module, while algorithm-specific ones should live next to each test case.
+
+```rust
+// Missing example.
```
DrPeterVanNostrand: Did you want to keep this missing for now?

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+    D::Message: Clone,
+    D::Input: Clone,
+    D::Output: Clone,
+{
+    /// Send input to all nodes.
+    ///
+    /// Equivalent to sending the same input to all nodes in order. Returns a vector of the
+    /// resulting `Step`s, which have had their messages queued already.
+    ///
+    /// If an error occurs, the first error will be returned and broadcasting aborted.
```
andogro: ~~will be~~ is

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+The old testing code can be found inside the `network` module and `.rs` files in the `tests` subdirectory that are not prefixed with `net_`. The newer networking code is contained inside the `net` module and the remaining `.rs` files.
+
+## VirtualNet
+
+Core of most tests is the `net::VirtualNet` struct, which simulates a network of nodes all running an instance of a distributed algorithm. Messages sent by these nodes are queued by the network. Every time the network is "cranked", a buffered message is delivered to its destination node and processed.
+
+Virtual networks can also host an adversary that can affect faulty nodes (which are tracked automatically) or reorder queued messages.
+
+To create a new network, the `NetBuilder` should be used:
```
andogro: Use the `NetBuilder` to create a new network.

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+    .using(move |node| { DynamicHoneyBadger::builder().build(node.netinfo) })
+    .build()
+    .expect("could not construct test network");
+```
+
+Algorithms that return a `Step` upon construction should use `using_step` instead.
+
+### Sending input
+
+`Input` can be sent to any node of the `VirtualNet` using the `send_input` method:
```
andogro: Send `Input` to any `VirtualNet` node using the `send_input` method:

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+  .trace(false)   // Never log network messages.
+  // ...
+```
+
+### Checking outputs
+
+As a convenience, all nodes capture any generated output during operation for inspection. The following code fragment demonstrates how to use this to verify the end result:
+
+```rust
+let first = net.correct_nodes().nth(0).unwrap().outputs();
```
DrPeterVanNostrand: To me, the sentence "demonstrates how to [...] verify the end result" sounds a bit vague. Maybe something like following would be more explicit:

```
The following code fragment demonstrates how to verify that all non-faulty nodes have output the same thing.
```

tests/README.md
DrPeterVanNostrand: I think that it is the responsibility of [the test implementer](https://github.com/poanetwork/hbbft/pull/216/files#diff-b5477a5a9bdae3f3f27bc239c3cab996R10) not to mutate the network in any way not allowed by our current model, maybe just note that "you may notice that it's possible for an adversary to mutate any part of `VirtualNet` (i.e. to change things beyond the scope of our adversary model), our current model only allows for an adversary to do: x, y, and z."

tests/README.md
```diff
@@ -0,0 +1,120 @@
+# Honey-badger tests
+
+The `hbbft` crate comes with a toolkit for testing its various algorithms in simulated network environments.
+
+## Old vs new
+
+The old testing code can be found inside the `network` module and `.rs` files in the `tests` subdirectory that are not prefixed with `net_`. The newer networking code is contained inside the `net` module and the remaining `.rs` files.
```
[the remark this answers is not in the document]
mbr: We will migrate step-by-step, I will try not to send people running towards the hills with another four-digit pull request! I'm adding a section to the `README.md`.

tests/README.md
mbr: > Actually it seems that Adversary::pre_crank allows to mutate any part of the VirtualNet. So I think this sentence is not complete.

Yes, that is the unfortunate reality. High on my list is adding a trait that restricts manipulation by the adversary to those operations they are actually allowed to perform, but I did not want to hold up review for this and added a note instead.

@DrPeterVanNostrand is right here. The `README.md` makes the mistake of assuming that one has already read all of the source code when reading the intro, which is silly. I will add more information to that specific section.

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## VirtualNet
+
+Core of most tests is the `net::VirtualNet` struct, which simulates a network of nodes all running an instance of a distributed algorithm. Messages sent by these nodes are queued by the network. Every time the network is "cranked", a buffered message is delivered to its destination node and processed.
```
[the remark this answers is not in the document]
mbr: See below.

tests/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+While the resulting step is returned, it needn't be processed to keep the network going, as its messages are automatically added to the queue.
+
+Instead of targeting a node in particular, the same input can be sent to all nodes:
+
+```rust
+net.broadcast_input(input).expect("algorithm failed");
+```
+
+### Cranking the network
```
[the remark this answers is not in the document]
mbr: I added those because I felt the word *crank* was not commonly used. I felt like we need an extra iteration-type of term, because we are using so many others in different areas already (iteration, step). The distinction really helps when talking about it.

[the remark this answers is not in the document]
mbr: I will remove the quotes and try to reword the section a bit.

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+                    id: id.clone(),
+                    netinfo,
+                    faulty: is_faulty,
+                });
+                steps.insert(id.clone(), step);
+                (id, Node::new(algorithm, is_faulty))
+            }).collect();
+
+        let mut message_count: usize = 0;
+        // For every recorded step, apply it.
```
afck: "Apply all recorded steps."?

tests/net/mod.rs
mbr: It's even more accurate, since we are using a builder. Awesome!

tests/README.md
mbr: If I only knew that myself ;). They are opened with no path prefix, so they end up in `cargo test`'s CWD. I'll add an explanation.

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+            }
+        }
+
+        // Step 0: We give the Adversary a chance to affect the network.
+
+        // Swap the adversary out with a dummy, to get around ownership restrictions.
+        let mut adv = mem::replace(&mut self.adversary, None);
+        if let Some(ref mut adversary) = adv {
+            // If an adversary was set, we let it affect the network now.
+            adversary.pre_crank(self)
```
afck: Should this have a semicolon, to make it clearer that it's `()`?

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+        if let Some(limit) = self.message_limit {
+            if self.message_count >= limit {
+                return Some(Err(CrankError::MessageLimitExceeded(limit)));
+            }
+        }
+
+        // Step 0: We give the Adversary a chance to affect the network.
+
+        // Swap the adversary out with a dummy, to get around ownership restrictions.
+        let mut adv = mem::replace(&mut self.adversary, None);
```
afck:
```rust
let mut adv = self.adversary.take();
// ...
self.adversary = adv;
```

would be more concise, I think. No need for `mem::replace`.
(Also below.)

tests/README.md
mbr: Yes, because I will add the adversarial traits next and will have to write a new example then. I will make it a priority though.

tests/net/mod.rs
mbr: I should set up an auto-correct -.-.

tests/README.md
mbr: Will add one, albeit later.

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+        steps.iter().for_each(|(id, step)| {
+            self.message_count = self.message_count.saturating_add(process_step(
+                &mut self.nodes,
+                id.clone(),
+                step,
+                &mut self.messages,
+            ));
+        });
+
+        Ok(steps)
```
afck: Would it be easier to call `send_input` in a loop?

tests/net/util.rs
```diff
@@ -0,0 +1,76 @@
+//! Utility functions
+//!
+//! A collection miscellaneous functions that are used in the tests, but are generic enough to be
```
afck: "collection of"

tests/net/util.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl<T> SubSlice for [T] {
+    #[inline]
+    fn subslice(
+        &self,
+        mut range: ops::Range<usize>,
+    ) -> &<Self as ops::Index<ops::Range<usize>>>::Output {
+        if range.end > self.len() {
+            range.end = self.len();
+        }
```
afck: Should the case `range.start > self.len()` also be taken into account?

tests/net_dynamic_hb.rs
```diff
@@ -0,0 +1,201 @@
...
+        .expect("broadcasting failed");
+
+    // We are tracking (correct) nodes' state through the process by ticking them off individually.
+    let mut awaiting_removal: collections::BTreeSet<_> =
+        net.correct_nodes().map(|n| n.id().clone()).collect();
+    let mut awaiting_addition: collections::BTreeSet<_> =
+        net.correct_nodes().map(|n| n.id().clone()).collect();
+    let mut expected_outputs: collections::BTreeMap<_, collections::BTreeSet<_>> = net
+        .correct_nodes()
+        .map(|n| (n.id().clone(), (0..10).into_iter().collect()))
```
afck: Should this be `(0..total_txs).into_iter().collect()`? And is the `into_iter()` necessary?

tests/net_dynamic_hb.rs
```diff
@@ -0,0 +1,201 @@
...
+            && awaiting_addition.is_empty()
+            && awaiting_removal.is_empty()
+        {
+            // All outputs are empty and all nodes have removed and added the single pivot node.
+            break;
+        }
+
+        // If not done, check if we still want to propose something.
+        if has_output {
+            if !queue.is_empty() {
```
afck: Doesn't the first block already handle the case `queue.len() == 0`?

tests/net_dynamic_hb.rs
```diff
@@ -0,0 +1,201 @@
...
+                // Out of the remaining transactions, select a suitable amount.
+                let proposal = rand::seq::sample_slice(
+                    &mut rng,
+                    // FIXME: Use better numbers and proptest.
+                    queue.as_slice().subslice(0..10),
+                    10.min(3.min(queue.len())),
```
afck: Isn't that always equal to `3.min(queue.len())`? Or did you mean to write `total_txs` instead of `10`?
You could also flip the order to reduce the nesting:

```rust
queue.len().min(total_txs).min(3)
```
```

**Comment by mbr** [comment]
> Also, Clippy and rustfmt complained in CI.

I was trying to move us towards stable by rustfmt'ing with beta, but clippy seems to be broken there. I'll revert my environment back to nightly-2018-07-13.

File: tests/README.md [hunk 36]
```diff
@@ -0,0 +1,124 @@
...
+
+The network advances through the `crank()` function, on every call
+
+1. the adversary is given a chance to re-order<sup>1</sup> the message queue,
```
DrPeterVanNostrand: Should these be numbered 1, 2, 3, and 4, instead of 1, 1, 1, and 1?

tests/net/adversary.rs
```diff
@@ -0,0 +1,77 @@
+//! Adversaries for test networks
+//!
+//! Adversaries can alter message ordering, inject messages and control the behaviour of any faulty
```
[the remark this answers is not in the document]
mbr: Make orthography great again!

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+    D::Message: Clone,
+    D::Output: Clone,
+    I: IntoIterator<Item = D::NodeUid>,
+{
+    /// Construct a new network builder.
+    ///
+    /// `node_ids` must be an iterator of the node ids to use for the new node. In many test cases,
+    /// a `Range(usize)` is a convenient value:
+    ///
+    /// ```rust,ignore
```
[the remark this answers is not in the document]
mbr: It won't typecheck, because there's no chance for `D` and `I` to be inferred without extra info. Adding those adds a lot of noise.

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+    /// let builder = NetBuilder::new(0..10) // ...
+    /// ```
+    #[inline]
```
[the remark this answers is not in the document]
mbr: The compiler automatically inlines within a crate, so they are technically not necessary. I add them on short functions habitually, in case they are ever called cross-crate.

Also, `#[inline]` is just allowing functions to be inline (as opposed to `#[inline(always)]` or `#[inline(never)]`. Putting them there has, in this case, no effect if not called from the outside, if I am not mistaken.

I am really hoping the default changes at some point...

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+    /// Iterator over *all* nodes in the network.
+    #[inline]
+    pub fn nodes(&self) -> impl Iterator<Item = &Node<D>> {
+        self.nodes.values()
+    }
+
+    /// Iterator over all faulty nodes in the network.
+    #[inline]
+    pub fn faulty_nodes(&self) -> impl Iterator<Item = &Node<D>> {
+        self.nodes().filter(|n| n.is_faulty())
```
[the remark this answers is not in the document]
mbr:
```
error[E0631]: type mismatch in function arguments
   --> tests/net/mod.rs:456:22
    |
97  |     pub fn is_faulty(&self) -> bool {
    |     ------------------------------- found signature of `for<'r> fn(&'r net::Node<_>) -> _`
...
456 |         self.nodes().filter(Node::is_faulty)
    |                      ^^^^^^ expected signature of `for<'r> fn(&'r &net::Node<D>) -> _`
```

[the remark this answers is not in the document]
mbr: Also breaks symetry with `correct_nodes`!

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+where
+    D: DistAlgorithm,
+    D::Message: Clone,
+    D::Output: Clone,
+{
+    /// Create new virtual network with step constructor.
+    ///
+    /// Creates a new network from `node_ids`, with the first `faulty` nodes marked faulty. To
+    /// construct nodes, the `cons` function is passed the ID and the generated `NetworkInfo` and
+    /// expected to return a (`DistAlgorithm`, `Step`) tuple.
```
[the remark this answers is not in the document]
mbr: Well, it's returning `(D, Step<D>)`, I thought the above was easier to understand?

tests/net/mod.rs
```diff
@@ -0,0 +1,782 @@
...
+        F: Fn(NewNodeInfo<D>) -> D + 'static,
+    {
+        self.using_step(move |node| (cons_simple(node), Default::default()))
+    }
+
+    /// Create the network.
+    ///
+    /// Finalizes the builder and creates the network.
+    ///
+    /// # Panic
```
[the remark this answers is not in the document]
mbr: `# Panics` is correct, I'll fix it.

tests/net/adversary.rs
```diff
@@ -0,0 +1,80 @@
...
+    D: DistAlgorithm,
+    D::Message: Clone,
+    D::Output: Clone,
+{
+    /// Pre-crank hook.
+    ///
+    /// Executed before each crank, the `pre_crank` function allows the adversary to manipulate the
+    /// order of network messages by manipulating the `net` parameter.
+    ///
+    /// The default implementation does not alter the passed network in any way.
```
DrPeterVanNostrand: Duplicate comment here and within the body of the method.

tests/net/mod.rs
mbr: It would, but iterating over all node IDs borrows `self` by borrowing either `self.nodes` or `self.nodes()`. At that point, it's not possible to call `self.send_input` anymore.

The implementation here avoids having to copy the list of IDs and as a bonus sends all inputs "at once", that is steps are only processed once all nodes have sent theirs.

tests/net/adversary.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// The default implementation does not perform any tampering, but instead calls
+    /// `VirtualNet::dispatch_message`, which results in the message being processed as if the node
+    /// was not faulty.
+    #[inline]
+    fn tamper(
+        &mut self,
+        net: &mut VirtualNet<D>,
+        msg: NetMessage<D>,
+    ) -> Result<Step<D>, CrankError<D>> {
+        // By default, no tampering is done.
```
DrPeterVanNostrand: Duplicate comment here and within the method docs.

tests/README.md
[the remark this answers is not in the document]
DrPeterVanNostrand: I think that defining "crank" as the network handling a single message (I think... right? I may be wrong) would be beneficial here. As "crank" is a commonly used term in the code, but may not be totally understood without an explicit definition.

tests/net/util.rs
mbr: Fixed, also added a test.

tests/net/err.rs
```diff
@@ -0,0 +1,108 @@
...
+        /// Network message that triggered the error.
+        msg: NetMessage<D>,
+        // Note: Currently, neither [Error](std::error::Error) nor [Fail](failure::Fail) are
+        //       implemented for [`D::Error`].
+        //       Either would be preferable, and would enable a [`failure::Fail::cause`]
+        //       implementation.
+        /// Error produced by `D`.
+        err: D::Error,
+    },
+    /// A node unexpectly disappeared from the list of notes. Note that this is likely a bug in
```
DrPeterVanNostrand: "notes" -> "nodes"

tests/README.md
```diff
@@ -0,0 +1,124 @@
...
+For convenience, an iterator interface is also available:
+
+```rust
+for res in net {
+    let (node_id, step) = res.expect("algorithm error");
+    // ...
+}
+```
+
+This has the drawback that access to the network is not available between cranks, as it is borrowed inside the for-loop. A common workaround is using a while loop instead:
```
vkomenda: "Cranking" is an application of a crank. "Between cranks" doesn't sound right. Better said "between applications of `VirtualNet::crank`".

tests/net/mod.rs
```diff
@@ -0,0 +1,796 @@
...
+
+                dest.push_back(NetworkMessage::new(
+                    sender.clone(),
+                    tmsg.message.clone(),
+                    to.clone(),
+                ));
+            }
+            // Broadcast messages get expanded into multiple direct messages.
+            messaging::Target::All => for to in nodes.keys() {
+                if *to == sender {
```
vkomenda: Maybe add `filter` instead to the loop iterator for readibility?

tests/net/mod.rs
```diff
@@ -0,0 +1,796 @@
...
+        let msg_copy = msg.clone();
+        let step = node
+            .algorithm
+            .handle_message(&msg.from, msg.payload)
+            .map_err(move |err| CrankError::AlgorithmError { msg: msg_copy, err })?;
+
+        Ok(step)
+    }
+
+    /// Send input to a specific node
```
DrPeterVanNostrand: "node" -> "node."

tests/README.md
mbr: Clarified.

tests/net/mod.rs
mbr: Sounds like a great idea. I'll add it.

mbr:
Alright, I'm thinking this might be the "final" version, unless bugs crop up. I will let the CI finish, but I would be happy to find a green checkmark on this later. =)

PR #218, opened 2018-08-30
Rename Agreement to BinaryAgreement.

### LINKED ISSUE #72: Consistently name individual algorithms.

afck:
Closes #72.


tests/binary_agreement.rs
```diff
@@ -1,7 +1,7 @@
 #![deny(unused_must_use)]
-//! Tests of the Binary Byzantine Agreement protocol. Only one proposer instance
+//! Tests of the Byzantine Binary Agreement protocol. Only one proposer instance
```
vkomenda: Our Agreement is one of Byzantine Agreement protocols which is also binary. So I think the original word order is more appropriate. This is somewhat confirmed by the HBBFT paper. What if you omit "Byzantine" here to avoid introducing a new term?

Issue #43, opened 2018-05-22
Limit message caches and future epochs?


afck:
We need to ensure that an attacker can't fill up a node's memory by sending it lots of messages that it stores forever, particularly in the algorithms that proceed in epochs (agreement, Honey Badger, …) and that, to be fully asynchronous, would in theory need to store any information for any future epoch until they can process it.

It's tempting to enforce some reasonable limits instead, and just assume that we'll never fall behind by more than 1000 epochs. But then an attacker could break consensus by sending a message that is _exactly_ 1000 epochs ahead and will be handled by only some of the honest nodes.

Either way, we should try to limit the amount of memory _per epoch_: E.g. in agreement, `incoming_queue` should be replaced by just the information which of the other nodes sent which of the four possible messages: `BVal(true), BVal(false), Aux(true), Aux(false)`.

* [X] Honey Badger
* [ ] Binary Agreement

afck:
See also: https://github.com/amiller/HoneyBadgerBFT/issues/57

mbr:
Well, you asked for some ideas:

Isn't this a typical problem of the attacker being able to force the peer to allocate a resource with very little cost themselves? This is not without precedent, reminding me [Slowloris](https://en.wikipedia.org/wiki/Slowloris_(computer_security)) or [SYN floods](https://en.wikipedia.org/wiki/SYN_cookies).

Neither of these are a perfect match for our problem, but both are solved by forcing more work on the attacker, in the examples above by constraining the time window for sending data more strictly, or moving the memory burden back onto the attacker.

What about limiting the underlying transmission channel, e.g. adding a "queue full, please retry" response? In that case, the attacker would have to buffer the message themselves and is expected to retry transmitting it.

A sender could simply use a priority queue that sends out messages with the lowest epochs first, so this would only be an issue if two honest nodes have vastly diverged in epochs. These are parameters that can be tweaked though and we are not going to have infinite resources either way.

Maybe even normal TCP backpressure is enough as well; if messages from honest nodes are rarely out of order, we could just stop reading from the TCP connection for a while. This in turn will cause sending to stop or slow down on the other peer, naturally causing them to buffer.

Finally, we should probably define: How should the network behave if a single node is vastly inferior in computing or network capacity than the others? Imagine eight high-performance servers and a ninth node running on a slow machine behind tor. The network can either throttle itself, dropping down the performance enough to include this node, or mark it as faulty and run at a much higher speed.

afck:
[...]

PR #219, opened 2018-08-30
Spam protection part 1: remote epoch tracking in HoneyBadger

vkomenda:
This PR adds a basic part of the spam protection functionality according to #43. The following are added:

1. a map of known `HoneyBadger` epochs of remote `HoneyBadger` nodes;
2. a message broadcast to all remote nodes signalling the start of an epoch;
3. a step output from the builders of `HoneyBadger`, `DynamicHoneyBadger` and `QueueingHoneyBadger`, with the necessary adjustments in tests and in the simulation.

The tests now pass and the PR is ready for review.

### DIFF
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -435,11 +435,22 @@ fn main() {
         .map(|_| Transaction::new(args.flag_tx_size))
         .collect();
     let new_honey_badger = |netinfo: NetworkInfo<NodeId>| {
-        let dyn_hb = DynamicHoneyBadger::builder().build(netinfo);
-        QueueingHoneyBadger::builder(dyn_hb)
+        let (dyn_hb, dyn_hb_step) = DynamicHoneyBadger::builder()
+            .build(netinfo)
+            .expect("`DynamicHoneyBadger` builder failed");
+        // Convert the initial step of the `DynamicHoneyBadger` instance into the initial step of
+        // the `QueueingHoneyBadger` instance.
+        let mut step = Step {
+            output: dyn_hb_step.output,
+            fault_log: dyn_hb_step.fault_log,
+            messages: dyn_hb_step.messages,
+        };
+        let (qhb, qhb_step) = QueueingHoneyBadger::builder(dyn_hb)
             .batch_size(args.flag_b)
             .build_with_transactions(txs.clone())
-            .expect("instantiate QueueingHoneyBadger")
+            .expect("instantiate QueueingHoneyBadger");
+        step.extend(qhb_step);
+        (qhb, step)
     };
     let hw_quality = HwQuality {
         latency: Duration::from_millis(args.flag_lag),
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -435,20 +435,14 @@ fn main() {
         .map(|_| Transaction::new(args.flag_tx_size))
         .collect();
     let new_honey_badger = |netinfo: NetworkInfo<NodeId>| {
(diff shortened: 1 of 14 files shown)


src/honey_badger/honey_badger.rs
```diff
@@ -126,13 +145,27 @@ where
         self.epoch += 1;
         self.has_input = false;
         let max_epoch = self.epoch + self.max_future_epochs;
-        let mut step = Step::default();
+        let mut msgs = VecDeque::new();
+        // The first message in an epoch announces the epoch transition.
+        msgs.push_back(Target::All.message(MessageContent::EpochStarted.with_epoch(self.epoch)));
+        let mut step = Step::new(Default::default(), Default::default(), msgs);
```
afck: This can also just be `Target::All.message(MessageContent::EpochStarted.with_epoch(self.epoch)).into()`.

In general, I'd try to use (and, where necessary, create) as many helper methods for `Step` as possible to avoid using lots of lines in the algorithm code for constructing it.

tests/queueing_honey_badger.rs
```diff
@@ -107,8 +107,19 @@ where
...
+    let (dhb, dhb_step) = DynamicHoneyBadger::builder()
+        .build((*netinfo).clone())
+        .expect("`new_queueing_hb` failed");
+    // Convert the initial step of the `DynamicHoneyBadger` instance into the initial step of
+    // the `QueueingHoneyBadger` instance.
+    let mut step = Step {
+        output: dhb_step.output,
+        fault_log: dhb_step.fault_log,
+        messages: dhb_step.messages,
+    };
```
afck: This could also use `convert()`.

examples/simulation.rs
```diff
@@ -437,11 +437,16 @@ fn main() {
...
+            .build(netinfo)
+            .expect("`DynamicHoneyBadger` builder failed");
+        let (qhb, qhb_step) = QueueingHoneyBadger::builder(dhb)
             .batch_size(args.flag_b)
             .build_with_transactions(txs.clone())
-            .expect("instantiate QueueingHoneyBadger")
+            .expect("instantiate QueueingHoneyBadger");
+        let mut step = dhb_step.convert();
+        step.extend(qhb_step);
+        (qhb, step)
```
afck: (We'll need to think about a way to make the process of creating a QHB easier. Not in this PR, of course.)

src/honey_badger/honey_badger.rs
```diff
@@ -106,6 +125,22 @@ where
         Ok(Step::default())
     }
 
+    /// Handles an epoch start announcement.
+    fn handle_epoch_started(&mut self, sender_id: &N, epoch: u64) {
+        self.update_remote_epoch(sender_id, epoch);
```
afck: Are `handle_epoch_started` and `update_remote_epoch` identical now? Then one of them can be removed.

src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -309,7 +306,9 @@ where
         } {
             info!("{:?} No-op change: {:?}", self.our_id(), change);
         }
-        self.restart_honey_badger(epoch);
+        let mut step: Step<C, N> = Step::default();
+        let step_on_restart = self.restart_honey_badger(epoch)?;
+        step.extend(step_on_restart);
```
afck: Could these three lines be merged?

```rust
let mut step = self.restart_honey_badger(epoch)?;
```

PR #220, opened 2018-08-30
Pinned threshold_crypto to version 0.1.0.

DrPeterVanNostrand:


DrPeterVanNostrand:
@afck yup, I agree.

c0gent:
Let me know when we're ready to publish and I'll take care of it.

PR #221, opened 2018-08-31
Advance to Rust beta

mbr:
For the planned migration to Rust 1.29, switch the CI over to the now working beta (tested with `rustc 1.29.0-beta.7 (49720ea9c 2018-08-26)`). The following advance to 1.29 should ideally be seamless afterwards.


mbr:
Alright, `threshold_crypto` is up to date as well.

PR #223, opened 2018-09-03
Adversary traits

mbr:
This PR adds the API for adversaries that (hopefully) restricts them to capabilities listed in the hbbft paper by wrapping all references in wrappers that enforce these invariants. Ideally, this should be free (except for the additional tests), at least in `--release` mode.

The resulting API will hopefully prove ergonomic. I also smuggled in another adversary implementation, which is a one-liner simulating the previous test-framework's behavior.

### DIFF
--- a/tests/net/adversary.rs
+++ b/tests/net/adversary.rs
@@ -7,14 +7,291 @@
 //! any `DistAlgorithm`. Specific adversaries tailored to individual algorithms are implemented
 //! alongside their other test cases.
 //!
-//! Note: Currently, adversaries are not "limited in power". In future versions, the references
-//!       passed to various callbacks might receive a smaller interface, to prevent manipulations
-//!       that are beyond the modelled adversaries capabilities. Currently, the implementor is
-//!       responsible for ensuring no invariants are violated.
+//! ## Adversary model
+//!
+//! The adversary is assumed to have the following capabilities:
+//!
+//! 1. Manipulation of the order in which messages are delivered.
+//! 1. Eavesdropping on any message on the wire, regardless of sender or receiver.
+//! 1. Full control over any node marked as faulty.
+//!
+//! As a consequence, injecting arbitrary messages from faulty nodes into the network is possible,
+//! by sending a message and re-ordering it.
+//!
+//! The following capabilities are explicitly not included:
+//!
+//! 1. Dropping of messages. The networking layer is expected to ensure that no messages
+//!    are lost. A node that drops messages regardless is considered faulty in real-world
+//!    deployments.
+//! 1. Forging message senders. The networking layer is also expected to sign messages and ensure
+//!    that they are not forged.
+//!
+//! ## Handles
+//!
+//! The adversary manipulates the network and nodes exclusively through handles that ensure they do
+//! not violate the constraints defined above. Handles are either mutable or immutable and can, in
+//! some cases be upgraded to actual references, if the underlying node is faulty (see
+//! `NodeHandle::node()` and `NodeHandle::node_mut()`).
 
-use net::{CrankError, NetMessage, VirtualNet};
+use std::cmp;
 
 use hbbft::messaging::{DistAlgorithm, Step};
+use net::{CrankError, NetMessage, Node, VirtualNet};
+
+/// Immutable network handle.
+///
+/// Allows querying public information of the network or getting immutable handles to any node.
+pub struct NetHandle<'a, D: 'a>(&'a VirtualNet<D>)
+where
+    D: DistAlgorithm;
+
+impl<'a, D: 'a> NetHandle<'a, D>
+where
+    D: DistAlgorithm,
+{
+    /// Returns a node handle iterator over all nodes in the network.
+    #[inline]
+    pub fn nodes(&self) -> impl Iterator<Item = NodeHandle<D>> {
+        self.0.nodes().map(NodeHandle::new)
+    }
+
+    /// Returns an iterator over all faulty in the network.
+    ///
+    /// Instead of a handle, returns the node directly, as the adversary gets full access to all
+    /// nodes in the network.
+    #[inline]
+    pub fn faulty_nodes(&self) -> impl Iterator<Item = &Node<D>> {
+        // FIXME: Add an API to handle a step?
+        // Not wrapped in a `NodeHandle`, the adversary gets full access to their own nodes.
+        self.0.faulty_nodes()
+    }
+
+    /// Returns a node handle iterator over all correct nodes in the network.
+    #[inline]
+    pub fn correct_nodes(&self) -> impl Iterator<Item = NodeHandle<D>> {
+        self.0.correct_nodes().map(NodeHandle::new)
+    }
+
+    /// Returns an iterator over all messages in the network.
+    #[inline]
+    pub fn messages(&'a self) -> impl Iterator<Item = &'a NetMessage<D>> {
+        self.0.messages()
+    }
+
+    /// Returns a handle to a specific node handle.
+    #[inline]
+    pub fn get(&self, id: D::NodeId) -> Option<NodeHandle<D>> {
+        self.0.get(id).map(NodeHandle::new)
+    }
+}
+
+/// Queue position.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum QueuePosition {
+    /// Front of the queue (equivalent to `Before(0)`).
+    Front,
+    /// Back of the queue.
+    Back,
+    /// Before a specific a position.
+    Before(usize),
+}
+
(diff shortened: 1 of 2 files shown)


tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+//! The adversary is assumed to have the following capabilities:
+//!
+//! 1. Manipulation of the order in which messages are delivered.
+//! 1. Eavesdropping on any message on the wire, regardless of sender or receiver.
+//! 1. Full control over any node marked as faulty.
+//!
+//! As a consequence, injecting arbitrary messages from faulty nodes into the network is possible,
+//! by sending a message and re-ordering it.
+//!
+//! The following capabilities are explicitly not included:
```
DrPeterVanNostrand: This is a really good addition to the docs imho.

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+where
+    D: DistAlgorithm,
+{
+    /// Returns a node handle iterator over all nodes in the network.
+    #[inline]
+    pub fn nodes(&self) -> impl Iterator<Item = NodeHandle<D>> {
+        self.0.nodes().map(NodeHandle::new)
+    }
+
+    /// Returns an iterator over all faulty in the network.
```
DrPeterVanNostrand: "faulty" -> "faulty nodes"

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+    }
+
+    /// Returns a handle to a specific node handle.
+    #[inline]
+    pub fn get(&self, id: D::NodeId) -> Option<NodeHandle<D>> {
+        self.0.get(id).map(NodeHandle::new)
+    }
+}
+
+/// Queue position.
```
DrPeterVanNostrand: Maybe a better doc would be something like: "position in the network's message queue"?

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+        NetMutHandle(net)
+    }
+
+    /// Returns a mutable node handle iterator over all nodes in the network.
+    #[inline]
+    pub fn nodes_mut(&mut self) -> impl Iterator<Item = NodeMutHandle<D>> {
+        self.0.nodes_mut().map(NodeMutHandle::new)
+    }
+
+    /// Returns an iterator that allows changes all faulty nodes in the network.
```
DrPeterVanNostrand: "allows changes all" -> "allows changes to all"

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+        self.0.dispatch_message(msg)
+    }
+
+    /// Injects a message into the network.
+    ///
+    /// Allows the injection of `msg` at `position` into the message queue.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `msg.from` is not a faulty or either `msg.from` or `msg.to` do not exist. Panics
```
DrPeterVanNostrand: "is not a faulty" -> "is not a faulty node"

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+// Downgrade-conversion.
+impl<'a, D: DistAlgorithm> From<NetMutHandle<'a, D>> for NetHandle<'a, D> {
+    #[inline]
+    fn from(n: NetMutHandle<D>) -> NetHandle<D> {
+        NetHandle(n.0)
+    }
+}
+
+/// Immutable node handle.
+pub struct NodeHandle<'a, D: 'a>(&'a Node<D>)
```
DrPeterVanNostrand: Is there a reason why you used `where` here instead of:

```rust
pub struct NodeHandle<'a, D: 'a, D: DistAlgorithm>(&'a Node<D>);
```

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+    Back,
+    /// Before a specific a position.
+    Before(usize),
+}
+
+/// Mutable network handle.
+///
+/// Allows reordering of messages, injecting new ones into the network queue and getting mutable
+/// handles to nodes.
+pub struct NetMutHandle<'a, D: 'a>(&'a mut VirtualNet<D>)
```
DrPeterVanNostrand: Is there a reason why you used `where` here instead of keeping everything on a single line?

```rust
pub struct NetMutHandle<'a, D: 'a, D: DistAlgorithm>(&'a mut VirtualNet<D>);
```

tests/net/adversary.rs
DrPeterVanNostrand: Same question goes for all the other places where the `where` keyword is separated out from the `<'a, D: 'a>` trait and lifetime bounds.

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+        NetHandle(n.0)
+    }
+}
+
+/// Immutable node handle.
+pub struct NodeHandle<'a, D: 'a>(&'a Node<D>)
+where
+    D: DistAlgorithm;
+
+impl<'a, D: DistAlgorithm> NodeHandle<'a, D>
```
DrPeterVanNostrand: Duplicate trait bound on `D: DistAlgorithm`?

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+    /// Panics if the node is not faulty.
+    #[inline]
+    pub fn node(&self) -> &'a Node<D> {
+        self.try_node()
+            .expect("could not access inner node of handle, node is not faulty")
+    }
+
+    /// Try to return a reference to the inner faulty node.
+    ///
+    /// If the inner node is faulty, returns a reference to it.
```
DrPeterVanNostrand: Duplicate information in docs. I think that `/// If the inner node is faulty, returns a reference to it.` would suffice as the docstring for this method.

tests/net/adversary.rs
DrPeterVanNostrand: Same question applies for everywhere the `where D: DistAlgorithm` is separated from the `<'a, D: 'a>` bounds.

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+/// Mutable node handle.
+pub struct NodeMutHandle<'a, D: 'a>(&'a mut Node<D>)
+where
+    D: DistAlgorithm;
+
+impl<'a, D: 'a> NodeMutHandle<'a, D>
+where
+    D: DistAlgorithm,
+{
+    /// Construct new mutable node handle.
```
DrPeterVanNostrand: "Construct new" -> "Construct a new"

tests/net/adversary.rs
```diff
@@ -7,14 +7,291 @@
...
+    fn new(inner: &'a mut Node<D>) -> Self {
+        NodeMutHandle(inner)
+    }
+
+    /// Returns a mutable reference to the faulty node.
+    ///
+    /// Allows mutable access to the wrapped node, if it is faulty.
```
DrPeterVanNostrand: First and second lines in the method docs are saying the same thing.

DrPeterVanNostrand:
Everything looks good to me besides a handful of nitpicks.

tests/net/adversary.rs
mbr: The main issue is that things tend to grow and most functions/structs have additional requirements (typically things like `D::NodeId: Clone`.

The Lifetimes make it extra weird. I think I'll probably do every*where*.

tests/net/adversary.rs
mbr: Fixed.

tests/net/adversary.rs
```diff
@@ -353,3 +353,30 @@ where
     D::Message: Clone,
     D::Output: Clone,
 {}
+
+/// Ascending node id message order adversary.
+///
+/// Simulates the behavior of the previous test framework, in which messages were processed by each
+/// node in order, with the lowest node IDs always being chosen first.
+#[derive(Debug, Default)]
+pub struct NodeOrderAdversary {}
```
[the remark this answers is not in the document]
mbr: Nothing, will be cleaned up.

tests/net/adversary.rs
```diff
@@ -7,14 +7,286 @@
...
+
+/// Insert-position for networking queue.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum QueuePosition {
+    /// Front of the queue (equivalent to `Before(0)`).
+    Front,
+    /// Back of the queue.
+    Back,
+    /// Before a specific a position.
+    Before(usize),
```
vkomenda: I'm thinking why `Before` is enough to have no `After`. Would `After(n)` be equivalent to `Before(n + 1)`?

PR #233, opened 2018-09-14
Outputing subset messages as received (initial work)

### LINKED ISSUE #170: Optimization: output early from Subset?
Subset could output multiple times: Once for each accepted contribution, and then once more when it terminates, to say that the remaining contributions have been rejected. That way, Honey Badger can start exchanging decryption shares sooner, and by the time the agreement instances for lagging nodes' contributions finally output `false`, the accepted ones will already be decrypted and ready to be output.

(We should measure, of course, whether it actually improves performance and whether it's worth it.)

### What needs to be changed

The Subset output needs to be an enum with two variants, e.g. `Contribution(Vec<u8>)` and `Done`.

Instead of outputting _all_ accepted contributions [when all have been decided](https://github.com/poanetwork/hbbft/blob/20193ab009104f46f89448673e118c908450bc23/src/subset.rs#L306), we should output a single `Contribution(_)` [as soon as it gets accepted](https://github.com/poanetwork/hbbft/blob/20193ab009104f46f89448673e118c908450bc23/src/subset.rs#L252), and then only output an empty `Done` once all have been decided.

In Honey Badger, we wouldn't [set the `SubsetState` to `Complete`](https://github.com/poanetwork/hbbft/blob/20193ab009104f46f89448673e118c908450bc23/src/honey_badger/epoch_state.rs#L225) on a `Contribution(_)` output, but only on `Done`.

The [simulation example](https://github.com/poanetwork/hbbft/blob/master/examples/simulation.rs) should be run a few times with and without the change, and maybe with different parameters, to test whether it improves performance.

DemiMarie:
Closes #170.

This is still incomplete; in particular, I need to fix up the test suite.


src/subset.rs
```diff
@@ -295,15 +301,16 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
         if self.decided || self.count_true() < self.netinfo.num_correct() {
             return None;
         }
         // Once all instances of BA have completed, let C ⊂ [1..N] be
         // the indexes of each BA that delivered 1. Wait for the output
         // v_j for each RBC_j such that j∈C. Finally output ∪ j∈C v_j.
         if self.ba_results.len() < self.netinfo.num_nodes() {
-            return None;
+            return Some(SubsetOutput::Contribution(proposer_id.clone(),
+             self.broadcast_results.get(proposer_id).map(|x|x.clone()).expect("internal error: already inserted entry not found")));
```
vkomenda: `rustfmt` won't accept this in Travis. Please check the error message.

tests/subset.rs
```diff
@@ -54,13 +54,18 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
     let output = expected.unwrap();
     assert!(once(&output).eq(network.observer.outputs()));
-    // The Subset algorithm guarantees that more than two thirds of the proposed elements
-    // are in the set.
-    assert!(output.len() * 3 > inputs.len() * 2);
-    // Verify that the set's elements match the proposed values.
-    for (id, value) in output {
-        assert_eq!(inputs[&id], value);
-    }
+    match output {
```
vkomenda: I think that in the test each node should output 5 same contributions - one for each node - and then `Done`. On master every node outputs a map:
```
[{NodeId(0): [65, 108, 112, 104, 97], NodeId(1): [66, 114, 97, 118, 111], NodeId(2): [67, 104, 97, 114, 108, 105, 101], NodeId(3): [68, 101, 108, 116, 97], NodeId(4): [69, 99, 104, 111]}]
```
In your test I see all outputs having only one contribution and then `Done`. Other contributions are missing.

src/subset.rs
```diff
@@ -124,6 +124,12 @@ impl<N: NodeIdT + Rand> DistAlgorithm for Subset<N> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SubsetOutput<N> {
```
vkomenda: Looks perfect!

src/subset.rs
DemiMarie: @vkomenda Fixed

DemiMarie:
I made no attempt to avoid unnecessary calls to `.clone()`.  I figured that it was more important to get the functionality working.

src/honey_badger/epoch_state.rs
```diff
@@ -219,12 +222,22 @@ where
...
         });
         if let Some(cs_output) = cs_outputs.pop_front() {
-            self.subset = SubsetState::Complete(cs_output.keys().cloned().collect());
-            step.extend(self.send_decryption_shares(cs_output)?);
+            match cs_output {
+                SubsetOutput::Contribution(k, v) => {
+                    let mut map: BTreeMap<N, Vec<u8>> = Default::default();
+                    self.keys.insert(k.clone());
+                    map.insert(k, v);
+                    step.extend(self.send_decryption_shares(map)?);
```
DemiMarie: Is this correct?  Should I be sending the decryption shares at this point, or is this premature?

src/honey_badger/epoch_state.rs
```diff
@@ -219,12 +222,22 @@ where
...
+                }
+                SubsetOutput::Done(output) => {
+                    self.subset = SubsetState::Complete(output.keys().cloned().collect());
+                    step.extend(self.send_decryption_shares(output)?);
+                }
```
DemiMarie: Or should I be sending them here?

src/subset.rs
```diff
@@ -295,15 +302,21 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
         // the indexes of each BA that delivered 1. Wait for the output
         // v_j for each RBC_j such that j∈C. Finally output ∪ j∈C v_j.
         if self.ba_results.len() < self.netinfo.num_nodes() {
-            return None;
+            return Some(SubsetOutput::Contribution(
+                proposer_id.clone(),
+                self.broadcast_results
+                    .get(proposer_id)
+                    .map(|x| x.clone())
+                    .expect("internal error: already inserted entry not found"),
```
DemiMarie: Is it reasonable to assume this?  I keep getting test failures, but am not sure that that is not due to something happening somewhere else.

DemiMarie:
Squashed.

src/subset.rs
```diff
@@ -295,15 +314,19 @@ impl<N: NodeIdT + Rand> Subset<N> {
         self.ba_results.values().filter(|v| **v).count()
     }
 
-    fn try_binary_agreement_completion(&mut self) -> Option<BTreeMap<N, ProposedValue>> {
+    fn try_binary_agreement_completion(
+        &mut self,
+        proposer_id: &N,
+        result: Option<Vec<u8>>,
+    ) -> Option<SubsetOutput<N>> {
```
afck: Maybe the "true" count check in this method should be moved further down, at least if we now want to output `Contribution`s (but not `Done`) even before we have `num_correct()` instances that returned `true`.

In any case, we need to make sure that for each accepted value (where the agreement instance returns `true`), we output `Contribution` exactly once. Note that this can be triggered by two things:
* An agreement instance outputs `true` and we already have the value—the output of the broadcast instance—for it.
* A broadcast instance outputs a value and we already have received the output `true` from the corresponding agreement instance. (Not very likely, but it can happen if our node is lagging.)

src/subset.rs
```diff
@@ -220,21 +228,31 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
     }
 
     /// Callback to be invoked on receipt of the decision value of the Binary Agreement
     /// instance `id`.
-    fn process_binary_agreement<F>(&mut self, proposer_id: &N, f: F) -> Result<Step<N>>
+    fn process_binary_agreement<F>(
+        &mut self,
+        proposer_id: &N,
+        f: F,
+        result: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
DemiMarie: Fixed with my latest commit

src/subset.rs
```diff
@@ -295,15 +314,19 @@ impl<N: NodeIdT + Rand> Subset<N> {
         self.ba_results.values().filter(|v| **v).count()
     }
 
-    fn try_binary_agreement_completion(&mut self) -> Option<BTreeMap<N, ProposedValue>> {
+    fn try_binary_agreement_completion(
+        &mut self,
+        proposer_id: &N,
+        result: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
DemiMarie: Fixed.

src/honey_badger/epoch_state.rs
```diff
@@ -219,12 +222,22 @@ where
...
     fn process_subset(&mut self, cs_step: cs::Step<N>) -> Result<Step<C, N>> {
         let mut step = Step::default();
-        let mut cs_outputs = step.extend_with(cs_step, |cs_msg| {
+        let mut cs_outputs: VecDeque<_> = step.extend_with(cs_step, |cs_msg| {
             MessageContent::Subset(cs_msg).with_epoch(self.epoch)
         });
         if let Some(cs_output) = cs_outputs.pop_front() {
-            self.subset = SubsetState::Complete(cs_output.keys().cloned().collect());
-            step.extend(self.send_decryption_shares(cs_output)?);
+            match cs_output {
```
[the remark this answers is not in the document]
DemiMarie: Fixed with my latest commit.

src/honey_badger/epoch_state.rs
```diff
@@ -219,12 +222,22 @@ where
...
+                SubsetOutput::Contribution(k, v) => {
+                    let mut map: BTreeMap<N, Vec<u8>> = Default::default();
+                    self.keys.insert(k.clone());
+                    map.insert(k, v);
+                    //step.extend(self.send_decryption_shares(map)?);
```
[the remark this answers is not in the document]
DemiMarie: @afck is this a testsuite bug or a bug in my code?

tests/subset.rs
```diff
@@ -45,21 +45,19 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
-    assert!(once(&output).eq(network.observer.outputs()));
+
     // The Subset algorithm guarantees that more than two thirds of the proposed elements
     // are in the set.
-    assert!(output.len() * 3 > inputs.len() * 2);
-    // Verify that the set's elements match the proposed values.
-    for (id, value) in output {
-        assert_eq!(inputs[&id], value);
+    assert!((output.len() - 1) * 3 > inputs.len() * 2);
+    for i in output {
```
afck: I'd remove this loop now. Instead, we should check in the loop above _for each node_ that the `Contribution` values are the same set, and that `Done` is output exactly once and at the end.

DemiMarie: Done

src/subset.rs
```diff
@@ -286,6 +296,16 @@ impl<N: NodeIdT + Rand> Subset<N> {
                 }
             }
         }
+        if let Some(x) = self.broadcast_results.insert(proposer_id.clone(), Default::default()) {
+            let y: Vec<u8> = x;
```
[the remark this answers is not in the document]
DemiMarie: It was.  Done.

src/honey_badger/epoch_state.rs
```diff
@@ -219,12 +222,26 @@ where
...
-        if let Some(cs_output) = cs_outputs.pop_front() {
-            self.subset = SubsetState::Complete(cs_output.keys().cloned().collect());
-            step.extend(self.send_decryption_shares(cs_output)?);
+        while let Some(cs_output) = cs_outputs.pop_front() {
+            match cs_output {
+                SubsetOutput::Contribution(k, v) => {
+                    self.map.insert(k.clone(), v.clone());
+                    let mut map = BTreeMap::default();
+                    map.insert(k, v);
+                    step.extend(self.send_decryption_shares(map)?);
```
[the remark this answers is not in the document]
DemiMarie: Agreed.  Done.

src/honey_badger/epoch_state.rs
```diff
@@ -219,12 +222,26 @@ where
...
     fn process_subset(&mut self, cs_step: cs::Step<N>) -> Result<Step<C, N>> {
         let mut step = Step::default();
-        let mut cs_outputs = step.extend_with(cs_step, |cs_msg| {
+        let mut cs_outputs: VecDeque<_> = step.extend_with(cs_step, |cs_msg| {
             MessageContent::Subset(cs_msg).with_epoch(self.epoch)
         });
-        if let Some(cs_output) = cs_outputs.pop_front() {
-            self.subset = SubsetState::Complete(cs_output.keys().cloned().collect());
-            step.extend(self.send_decryption_shares(cs_output)?);
+        while let Some(cs_output) = cs_outputs.pop_front() {
```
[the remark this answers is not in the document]
DemiMarie: Agreed and fixed.

src/subset.rs
```diff
@@ -295,11 +296,14 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
-                proposer_id.clone(),
-                x.clone(),
-            )));
+        if let Some(x) = self.broadcast_results.insert(proposer_id.clone(), Default::default()) {
+            let y: Vec<u8> = x;
+            if y.len() > 0 {
+                step.output.extend(Some(SubsetOutput::Contribution(
+                    proposer_id.clone(),
+                    y,
+                )));
```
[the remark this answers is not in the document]
DemiMarie: I agree with both of these.  Fixed.

src/subset.rs
```diff
@@ -295,11 +296,14 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
             }
         }
-        if let Some(x) = self.broadcast_results.get(proposer_id) {
-            step.output.extend(Some(SubsetOutput::Contribution(
-                proposer_id.clone(),
-                x.clone(),
-            )));
+        if let Some(x) = self.broadcast_results.insert(proposer_id.clone(), Default::default()) {
+            let y: Vec<u8> = x;
+            if y.len() > 0 {
```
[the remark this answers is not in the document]
DemiMarie: Agreed and fixed.

tests/subset.rs
```diff
@@ -45,27 +45,25 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
     let mut expected = None;
     for node in network.nodes.values() {
-        if let Some(output) = expected.as_ref() {
-            assert!(once(output).eq(node.outputs()));
-            continue;
-        }
-        assert_eq!(1, node.outputs().len());
-        expected = Some(node.outputs()[0].clone());
+        assert!(1 < node.outputs().len());
+        expected = Some(node.outputs().clone());
```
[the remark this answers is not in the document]
DemiMarie: Fixed.

src/subset.rs
```diff
@@ -276,15 +265,17 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
         debug!(
             "{:?} Updated Binary Agreement results: {:?}",
             self.netinfo.our_id(),
             self.ba_results
         );
 
+        // Binary agreement result accepted.
+        if self.ba_results.insert(proposer_id.clone(), value).is_some() {
+            return Err(Error::MultipleBinaryAgreementResults);
+        }
```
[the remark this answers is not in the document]
DemiMarie: That makes sense.  Fixed.

DemiMarie:
@afck Test suite still fails.  Is there a bug in https://github.com/DemiMarie/hbbft/blob/1b37e14d865e2443857379c7b5d32f237af75a73/tests/network/mod.rs#L91-L116?

DemiMarie:
@afck what should I change to fix the test suite?

c0gent:
@DemiMarie The most recent error is simply a formatting error. Just run `cargo +beta fmt` and it should clear up.

src/subset.rs
```diff
@@ -322,23 +345,20 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
             debug!(
                 "{:?} Binary Agreement instances completed:",
                 self.netinfo.our_id()
             );
-            for (id, result) in &broadcast_results {
-                debug!("    {:?} → {:?}", id, HexBytes(&result));
-            }
             self.decided = true;
-            Some(broadcast_results)
+            Some(SubsetOutput::Done)
```
afck: The removed `debug!` statement above should probably be moved to the point where we now output a `Contribution`.

src/subset.rs
```diff
@@ -220,7 +227,9 @@ impl<N: NodeIdT + Rand> Subset<N> {
                 return Ok(step);
             }
         };
-        self.broadcast_results.insert(proposer_id.clone(), value);
+        self.broadcast_results
+            .entry(proposer_id.clone())
+            .or_insert(Some(value));
```
afck: We need to make sure that we return the `Contribution` output even if our node is somehow lagging and receives the agreement output before the broadcast itself. I.e. at this point we need to check whether we already have an agreement output, and if it's `true`, return the `Contribution`.

(Alternatively `process_binary_agreement` could be modified to return the output in that case, too—not sure whether there's a clean way to do that.)

tests/subset.rs
```diff
@@ -45,22 +45,32 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
     // Verify that all instances output the same set.
     let mut expected = None;
     for node in network.nodes.values() {
-        if let Some(output) = expected.as_ref() {
-            assert!(once(output).eq(node.outputs()));
-            continue;
+        assert!(1 < node.outputs().len());
```
afck: This line can probably go, because there's a stronger assertion in line 66 now.

tests/subset.rs
```diff
@@ -45,22 +45,32 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
+                SubsetOutput::Contribution(k, v) => {
+                    actual_num_outputs += 1;
+                    assert!(actual.insert(k, v).is_none());
+                }
+                SubsetOutput::Done => break,
+            }
         }
-        assert_eq!(1, node.outputs().len());
-        expected = Some(node.outputs()[0].clone());
+        assert_eq!(expected_num_outputs, actual_num_outputs + 1);
```
afck: No need to count `actual_num_outputs`. That will just be `actual.len()`.

tests/subset.rs
```diff
@@ -45,22 +45,32 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
+        // The Subset algorithm guarantees that more than two thirds of the proposed elements
+        // are in the set.
+        assert!(actual_num_outputs * 3 > inputs.len() * 2);
+        match expected {
+            None      => expected = Some(actual),
+            Some(ref set) => assert_eq!(&actual, set),
+        };
+
     }
     let output = expected.unwrap();
```
vkomenda: @DemiMarie, you are now missing a comparison of outputs with inputs. You need this check to verify that the set of output contributions is the same as the set of input proposals. Do this for the observer node `network.observer` and for all correct nodes `network.nodes`.

src/subset.rs
DemiMarie: Done.

DemiMarie:
Is it normal that the testsuite seems to run in an infinite loop?

DemiMarie:
All tests pass on my system.

DemiMarie:
@afck Thank you for your suggestion; the logs allowed me to figure out the problem quickly

DemiMarie:
Should I squash?

vkomenda:
We can squash-merge in github.

tests/subset.rs
```diff
@@ -43,23 +43,26 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
+        // are in the set.
+        assert!(actual.len() * 3 > inputs.len() * 2);
+        for (id, value) in actual {
+            assert_eq!(&inputs[&id], value);
         }
-        assert_eq!(1, node.outputs().len());
-        expected = Some(node.outputs()[0].clone());
-    }
-    let output = expected.unwrap();
-    assert!(once(&output).eq(network.observer.outputs()));
```
vkomenda: Can you please add back the test of outputs from the observer node? It looks that you've forgotten to add it back while refactoring.

src/honey_badger/epoch_state.rs
```diff
@@ -116,6 +116,8 @@ pub struct EpochState<C, N: Rand> {
     subset: SubsetState<N>,
     /// The status of threshold decryption, by proposer.
     decryption: BTreeMap<N, DecryptionState<N>>,
+    /// N seen so far
```
vkomenda: "Nodes found so far."

src/honey_badger/epoch_state.rs
```diff
@@ -219,15 +222,35 @@ where
...
+                        .collect();
+                    for id in faulty_shares {
+                        if let Some(DecryptionState::Ongoing(td)) = self.decryption.remove(&id) {
+                            for id in td.sender_ids() {
+                                let fault_kind = FaultKind::UnexpectedDecryptionShare;
+                                step.fault_log.append(id.clone(), fault_kind);
+                            }
+                        }
+                    }
+                    break;
```
vkomenda: I think it's better to continue with the `for` loop and remove this `break`. In case there are more outputs in `cs_output`, don't ignore them but rather `panic!` in case the program finds any. This way we would prevent bugs later.

src/honey_badger/epoch_state.rs
```diff
@@ -116,6 +116,8 @@ pub struct EpochState<C, N: Rand> {
     subset: SubsetState<N>,
     /// The status of threshold decryption, by proposer.
     decryption: BTreeMap<N, DecryptionState<N>>,
+    /// Nodes found so far in `Subset` output.
+    nodes_found_so_far: BTreeSet<N>,
```
afck: Maybe `accepted_proposers`?
This is the set of IDs of those nodes whose contributions we already know have been voted into the next batch.

src/honey_badger/epoch_state.rs
```diff
@@ -219,15 +222,39 @@ where
...
-        if let Some(cs_output) = cs_outputs.pop_front() {
-            self.subset = SubsetState::Complete(cs_output.keys().cloned().collect());
-            step.extend(self.send_decryption_shares(cs_output)?);
-        }
-        if !cs_outputs.is_empty() {
-            error!("Multiple outputs from a single Subset instance.");
+        let mut has_seen_done = false;
+        for cs_output in cs_outputs {
+            if has_seen_done {
+                panic!("`SubsetOutput::Done` was not the last `SubsetOutput`");
```
afck: Not sure whether we want a `panic!` in production code. I agree that this should never happen and would be a bug, but maybe we should just log it as `error!` or not do the check at all. After all, we do have a test for this, too.

src/subset.rs
```diff
@@ -220,7 +228,25 @@ impl<N: NodeIdT + Rand> Subset<N> {
                 return Ok(step);
             }
         };
-        self.broadcast_results.insert(proposer_id.clone(), value);
+        {
+            let ba_results = &self.ba_results;
```
afck: There's probably no need for the curly braces and for borrowing `self.ba_results` anymore.

tests/subset.rs
```diff
@@ -42,24 +42,43 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
+        match x.clone() {
+            SubsetOutput::Contribution(x, y) => Some((x, y)),
+            SubsetOutput::Done               => None
+        }
+    };
+
     // Verify that all instances output the same set.
-    let mut expected = None;
+    let mut observer: Vec<_> = network.observer.outputs().iter().map(subsetoutput_to_option).collect();
+    observer.sort();
```
afck: Instead of sorting, you could collect it as a `BTreeSet` or `HashSet` (or a map).

(Good idea, though, to replace `expected` with the observer output! Much simpler. :+1: )

src/honey_badger/epoch_state.rs
vkomenda: The `panic!` is unreachable code, so it won't be executed in production, like `expect`s and indeed a few `panic!`s that we have. However I agree with @afck, an `error!` can be a good compromise in this case.
If we want a systematic approach to panics, we might refactor the library at some point to return critical errors to the user of the library instead of causing a critical error.

tests/subset.rs
DemiMarie: Thank you!

The most annoying part of this is that `SubsetOutput` doesn’t implement `Ord`, and if I add such a requirement for `T` then it bubbles up into the public API.  So I am forced to use `subsetoutput_to_option`.

vkomenda: Since you mentioned it, you could simplify the test by deriving `PartialOrd` and `Ord` for `SubsetOutput`. The derived instances use the lexicographic ordering, so `Done` is greater than `Contribution`.

vkomenda: Also see the Rust convention [C-COMMON-TRAITS](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits). We adopted the standard conventions for `hbbft`.

DemiMarie: @vkomenda I tried deriving `Ord` but it resulted in compiler errors.  I didn’t think to derive `PartialOrd` as well until I looked at the Rust source code.

src/honey_badger/epoch_state.rs
DemiMarie: Done

DemiMarie:
@vkomenda @afck is this version good?

tests/subset.rs
DemiMarie: Done

src/honey_badger/epoch_state.rs
DemiMarie: Done.

tests/subset.rs
```diff
@@ -42,24 +42,33 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
     while !network.nodes.values().all(TestNode::terminated) {
         network.step();
     }
+
     // Verify that all instances output the same set.
-    let mut expected = None;
+    let observer: BTreeSet<_> = network.observer.outputs().iter().cloned().collect();
```
vkomenda: `.iter().cloned().collect()` is equivalent to `.to_vec()`, I think.
Also you removed `observer.sort()`. BTW the test appears to pass without sorting. Maybe for the future (i.e., moving to the new `proptest` framework) it would still be better to sort.

tests/subset.rs
```diff
@@ -42,24 +42,33 @@ fn test_subset<A: Adversary<Subset<NodeId>>>(
...
+        assert_eq!(outputs.len(), actual.len() + 1);
+
+        // The Subset algorithm guarantees that more than two thirds of the proposed elements
+        // are in the set.
+        assert!(actual.len() * 3 > inputs.len() * 2);
+        for (id, value) in actual {
+            assert_eq!(&inputs[id], value);
+        }
+
+        assert_eq!(outputs.iter().cloned().collect::<BTreeSet<_>>(), observer);
```
vkomenda: Same observation: `.iter().cloned().collect()` is equivalent to `.to_vec()`
and you removed `outputs.sort()`.

DemiMarie:
I collected into a BTreeSet, not a Vec.  So the order doesn't matter.

vkomenda:
A `BTreeSet` would collapse multiple occurrences of the same output to one while a sorted `Vec` would keep those occurrences. So a sorted `Vec` represents a multiset - a set with multiplicities of its elements - rather than a simple set, and is more expressive.

tests/subset.rs
afck: Sorry, @vkomenda is right (see below), of course, and the `Vec` was probably a good idea.
But since we also count the number of outputs, double entries should automatically cause the test to fail, too, I guess, so it should be fine either way?

tests/subset.rs
afck: Since it's a `BTreeSet` now, `to_vec` wouldn't work, and sorting isn't possible/necessary anymore.

tests/subset.rs
afck: That's also obsolete, since it's a `BTreeSet` now, I think.

src/subset.rs
```diff
@@ -220,7 +227,22 @@ impl<N: NodeIdT + Rand> Subset<N> {
                 return Ok(step);
             }
         };
-        self.broadcast_results.insert(proposer_id.clone(), value);
+
+        let val_to_insert = if let Some(true) = self.ba_results.get(proposer_id) {
```
afck: No need to pattern-match here: This could just be `if Some(true) == self.ba_results.get(proposer_id)`.

On the other hand, maybe it should be a `match` instead: I'm wondering whether in the `Some(false)` case, we should also just insert `None`, since we won't need the value anyway—we already know it has been rejected?

tests/subset.rs
vkomenda: That's right, there is a uniqueness check in the test when a new contribution is inserted into a map. I think the difference is only in testing the observer output which is converted to a set without this check. It's a minor detail and I'll leave it to @DemiMarie whether to take care of it or not.

Issue #226, opened 2018-09-10
Dynamic Honey Badger sender queue


vkomenda:
Issue #43 needs an extra bit of design for queueing and dequeueing Honey Badger (HB) messages. An instance of Dynamic Honey Badger (DHB) contains an embedded instance of HB. This embedded HB instance is restarted on DKG change or completion events. When restarting it, any queued messages should be stored in the DHB instance for later delivery.

vkomenda:
In case of the DHB sender queue the extra complexity arises from the management issues associated with the embedded HB instance. In particular, the messages send by HB should be post-processed by the containing DHB. Post-processing allows to queue HB messages beyond the restart of the embedded HB instance.

Two kinds of post-processing is required.
1. Any `EpochStarted(epoch)` message from HB is annotated with the DHB epoch number `start_epoch` -- call it `DynamicEpochStarted(start_epoch, epoch) -- `and broadcast to all remote nodes. Remote nodes at DHB epoch `start_epoch` pass a message `EpochStarted(epoch)` to the embedded HB instance. Remote nodes at earlier DHB epochs drop the message. Remote nodes at later DHB epochs dequeue any queued outgoing messages for `(start_epoch, epoch)`.
1. The HB instance keeps a queue of messages for remote nodes with the same DHB `start_epoch`. So, any messages from HB other than `EpochStarted` can be sent without delay. If however there is a non-empty queue before the HB instance is restarted, the queue is appended to the DHB queue of postponed HB messages. An implementation of this queue is possible within an outgoing queue of DHB which is searchable by keys containing both a DHB `start_epoch` and an HB `epoch`.

afck:
> Remote nodes at earlier DHB epochs drop the message.

I don't think they should drop it: If I receive `DynamicEpochStarted(6, 8)` from you, and I'm still at `start_epoch = 2` and `epoch = 5`, then I know that:
* My current HB instance won't need to send anything to you ever again. (Basically we could make it handle an `EpochStarted(u64::MAX)` or something… although if we literally do that we'll need to be careful to avoid `+` and use `saturating_add` everywhere on epochs.)
* My _next_ HB instance will need to handle the `EpochStarted(8)` message as soon as I create it.

vkomenda:
> I don't think they should drop it

[...]

PR #237, opened 2018-09-17
Allowed building a HB with a non-0 starting epoch

### LINKED ISSUE #275: Allow starting a new network in an epoch other than 0.
E.g. when using Honey Badger for a blockchain, and the network is shut down and restarted, we'd probably want to start with a later epoch, to keep the epoch number in sync with the block number.

vkomenda:
This is a basic tool for starting `HoneyBadger` instances in non-trivial epochs as suggested by @afck in https://github.com/poanetwork/hbbft/issues/226#issuecomment-421761224.

Closes #275

### DIFF
--- a/src/honey_badger/builder.rs
+++ b/src/honey_badger/builder.rs
@@ -14,6 +14,8 @@ use {Contribution, NetworkInfo, NodeIdT};
 pub struct HoneyBadgerBuilder<C, N> {
     /// Shared network data.
     netinfo: Arc<NetworkInfo<N>>,
+    /// Start in this epoch.
+    epoch: u64,
     /// The maximum number of future epochs for which we handle messages simultaneously.
     max_future_epochs: usize,
     /// Random number generator passed on to algorithm instance for signing and encrypting.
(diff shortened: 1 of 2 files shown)


afck:
Sorry, I think that was a misunderstanding: We shouldn't ever need to start a Honey Badger instance in an epoch other than 0. Our _own_ epochs will never be skipped. What I meant by that comment was that the next Honey Badger instance should, at or immediately after its creation, handle any previously received `EpochStarted` messages concerning it (i.e. all that have the new `start_epoch` and therefore couldn't be handled previously) to update its knowledge of the _other_ nodes' current epochs.
(But maybe I'm just confused—please reopen if I'm wrong.)

afck:
> We shouldn't ever need to start a Honey Badger instance in an epoch other than 0.

I strongly disagree with @afck here! :grin: 
See https://github.com/poanetwork/hbbft/issues/275.

src/honey_badger/builder.rs
```diff
@@ -56,7 +66,7 @@ where
         };
         let step = if self.netinfo.is_validator() {
             // The first message in an epoch announces the epoch transition.
-            Target::All.message(Message::EpochStarted(0)).into()
+            Target::All.message(Message::EpochStarted(epoch)).into()
```
afck: I think we'd just need to remove this one line and rebase.
Then the DHB builder will need the same functionality.

vkomenda:
Let me know if you want a `HoneyBadger` starting epoch to be configurable from the `DynamicHoneyBadger` builder. At the moment it isn't.

PR #240, opened 2018-09-20
Remove itertools dependency.

afck:
With `Iterator::flatten` stabilized, we're not using `Itertools` anymore, other than in tests and examples.
On the other hand, this would break the build with anything less than Rust 1.29.0.


c0gent:
It might be wise to either add a note about the minimum version requirement to the readme or check the version with a `bulld.rs`.

mbr:
Might take a while for official support, even though it's a pain point. https://github.com/rust-lang/rfcs/pull/2495. So three months at the very least, if the core team already had agreed upon it.

I'll add something to the README.

PR #248, opened 2018-10-01
WIP: Better proptest persistence

mbr:


mbr:
Should I add some issues regarding a refactoring to redesign the API to be able to consistently pass in `<R: rand::Rng>`? This will break upstream somewhat and might have some far-reaching consequences, but it seems the right thing to do.

The chosen PRNGs are probably safe (Isaac64 *is* a CSPRNG), but they would be running without reseeding (unlike OsRng), which might be a nice safety net.

src/util.rs
```diff
@@ -0,0 +1,24 @@
+//! Utility functions
+//!
+//! Functions not large enough to warrant their own crate or module, but flexible enough to be used
+//! in multiple disjunct places in the library. May also contain backports, workarounds.
+
+use rand;
+
+/// Work-around trait for creating new random number generators
```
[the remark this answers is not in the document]
mbr: Let's debate this in depth later on!

src/util.rs
```diff
@@ -0,0 +1,24 @@
...
+pub trait SubRng {
+    fn sub_rng(&mut self) -> Box<dyn rand::Rng>;
+}
+
+impl<R> SubRng for R
+where
+    R: rand::Rng,
+{
+    fn sub_rng(&mut self) -> Box<dyn rand::Rng> {
```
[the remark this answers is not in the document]
mbr: It's still inferior to not having to create new random number generators at all, I think. If you have a high-quality RNG that can produce an infinite amount of real random numbers at a fast rate, you probably do not want to use PRNGs at all.

Also, auditing is probably much simpler if there's only one RNG. =)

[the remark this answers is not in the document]
mbr: As for the SubRng for specific rngs: We'd have to require SubRng as a trait for RNGs passed in then. I did experiment with a `DebuggableRng` trait, but multiple trait bounds in a Box is still something that Rust only does through workarounds and those really do not work well in other areas =(.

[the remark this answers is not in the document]
afck: But if you have a `Box<dyn SubRng>` and have `SubRng: Rng` as a trait bound, that should work, shouldn't it?

src/honey_badger/builder.rs
```diff
@@ -32,10 +35,17 @@ where
         HoneyBadgerBuilder {
             netinfo,
             max_future_epochs: 3,
+            rng: Box::new(rand::thread_rng()),
             _phantom: PhantomData,
         }
     }
 
+    /// Sets the random number generator for the public key cryptography.
+    pub fn rng<R: Rng + 'static>(&mut self, rng: R) -> &mut Self {
```
c0gent: This is a bad idea performance-wise. Putting an rng (which needs to be accessed in tight loops) behind dynamic dispatch is suboptimal. If the type of RNG is optional, make an enum for it. If you want to get fancy, create an `enum MySpecialRngEnumThing` type which impl's `From<VariousRngTypes>`. then make `R: Into<MySpecialRngEnumThing>`. If that doesn't make sense I'll elaborate.

src/honey_badger/builder.rs
```diff
@@ -18,6 +19,8 @@ where
     netinfo: Arc<NetworkInfo<N>>,
     /// The maximum number of future epochs for which we handle messages simultaneously.
     max_future_epochs: usize,
+    /// Random number generator passed on to algorithm instance for signing and encrypting.
+    rng: Box<dyn Rng>,
```
c0gent: See below note.

src/util.rs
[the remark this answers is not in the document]
c0gent: As I mentioned in another comment, boxing `Rng` is a bad idea for performance. Use an enum (with `From`/`Into` for convenience if needed).

src/honey_badger/builder.rs
afck: In our case I think we never use an RNG in any inner loop, so it shouldn't be a bottleneck.

c0gent: Ummm... really? Regardless you don't want to use dynamic dispatch for stuff like this. It's bad practice. You never know what that rng will end up being used for in the future.

afck: We pretty much know what our inner loops and our CPU bottlenecks are, and where we use RNGs, and I don't think that's going to change fundamentally. And an enum would be a bit less flexible, limiting the user's choices for an RNG.
(I'm fine with either approach, I just don't think performance is a concern here.)

mbr: Without mocktography, we are spending over 99% in `threshold_crypto` (most of it not on the RNG part). Real-life impact should be not measurable, although I would love to be proven wrong here =).

I think getting rid of the boxed RNG has two main benefits: Removing the pointer lookup on call and allowing inlining. The removal of pointer lookups does not apply here, we would be replacing that lookup with a custom one (`match`ing on the `enum`), as its type cannot always be known at compile time, we are likely to not be able to get rid of that. So by my estimate we are paying with a single added function call.

All in all, I do not think the rather small gains in speed will outweigh the added cost in code complexity here, otherwise we should probably add a generic type parameter as described in the comments, to get the RNG type nailed down during instantiation and everything fully inlined.

In addition, the `Isaac64` RNG is not free to compute either; if the `OsRng` is used (used in threshold crypto before), a full syscall was required for every random value. I would assume that this PR already runs faster.

We might see more substantial relative gains with `XorShiftRng` and mocktography, I do not think we should optimize for that though. Since there's still the possibility of moving RNGs out of the algorithms state later, which hopefully would solve all of these issues, I'd like to postpone action here. =)

c0gent: The added cost in code complexity of an enum is negligible. Trait objects are bad for performance and we need every bit we can get. We should be willing to trade 100 lines of code for a 1% performance gain because those gains add up and compound quickly. I'm not sure about the precise gain to be had here but it doesn't matter. We need to squeeze every bit out. 

Leave this alone for now and I'll submit a PR later to change it.

afck: I agree, let's move this discussion out of this PR. :+1:

Issue #249, opened 2018-10-01
Double panic when trying to run `cargo test` on Win10


ErichDonGubler:
For some reason tests inside of `src/dynamic_honey_badger/votes.rs` are throwing panics inside of panics. This causes an immediate abort, and blocks tests from running. This seems specific to my Windows 10 machine, since `master` seems to pass just fine in Travis (a *nix environment).

Output of `cargo test`:

```
C:\Users\egubler\workspace\personal\pull-requests\hbbft>cargo test
    Finished dev [unoptimized + debuginfo] target(s) in 0.53s
     Running target\debug\deps\hbbft-0cddedd180c2030e.exe

running 3 tests
test broadcast::merkle::tests::test_merkle ... ok
test dynamic_honey_badger::votes::tests::test_pending_votes ... FAILED
thread panicked while panicking. aborting.
error: test failed, to rerun pass '--lib'
```

Output of `rustup show`:

```
active toolchain
----------------

stable-x86_64-pc-windows-gnu (default)
rustc 1.29.0 (aa3ca1994 2018-09-11)
```

<details>
<summary>

**Output of `RUST_BACKTRACE=1 cargo test` in MSYS2 environment**

</summary>

```
~/…/pull-requests/hbbft master$ RUST_BACKTRACE=1 cargo test
    Finished dev [unoptimized + debuginfo] target(s) in 0.57s
     Running target\debug\deps\hbbft-0cddedd180c2030e.exe

running 3 tests
test broadcast::merkle::tests::test_merkle ... ok
test dynamic_honey_badger::votes::tests::test_committed_votes ... FAILED
test dynamic_honey_badger::votes::tests::test_pending_votes ... FAILED

failures:

---- dynamic_honey_badger::votes::tests::test_committed_votes stdout ----
thread 'dynamic_honey_badger::votes::tests::test_committed_votes' panicked at 'Failed to drop `SecretKey`: Failed to `munlock` 32 bytes starting at address: 0x1a90c30', C:\Users\egubler\.cargo\git\checkouts\threshold_crypto-3605c1a1492f376b\02109b5\src\lib.rs:308:13
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::windows::backtrace::unwind_backtrace
             at libstd\sys\windows\backtrace/mod.rs:104
             at libstd\sys\windows\backtrace/mod.rs:91
   1: std::sys_common::backtrace::print
[...]

PR #250, opened 2018-10-01
Make Subset optimization optional

### LINKED ISSUE #243: Make Subset optimization optional.
Let's make the optimization introduced in #170 an option, configurable per Honey Badger instance. We should allow disabling it to
* allow behavior exactly as specified in the paper,
* make it easy to compare performance with and without it,
* enable other features that might conflict with it, like #183.

ErichDonGubler:
Fixes #243. Allow client code to control the behavior of `EpochState` and when it decides to send decryption shares. There are two options for the behavior for handling the output of a `Subset` algorithm, which are represented by the new `SubsetHandlingStrategy` (bikeshedding welcome!) enum:

1) `Incremental`, which is the current (optimized) behavior.
2) `AllAtEnd`, which is supposed to be the old behavior. *Reviewers crunched for time might want to place priority on making sure I've done this right.*

`SubsetHandlingStrategy` is exposed at the builder level. Internally, an `EpochState` instance now stores a `SubsetHandler` enum member that models the necessary state for each different strategy. To resolve ownership issues, I've changed `EpochState::{process_decryption,send_decryption_share}` to take members of the `EpochState` struct instead of the entire struct (`self`). I've also created the `finalize_subset_with_step` method which also uses members and not `self`.

~~Tests compile, but I'm currently blocked on https://github.com/poanetwork/hbbft/issues/249 for testing locally until I get home and have access to my *nix machines.~~

## Current concerns

- [ ] Bikeshedding on new symbol names?

### DIFF
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -21,6 +21,7 @@ pub struct DynamicHoneyBadgerBuilder<C, N> {
     /// Random number generator passed on to algorithm instance for key generation. Also used to
     /// instantiate `HoneyBadger`.
     rng: Box<dyn rand::Rng>,
+    /// Strategy used to handle the output of the `Subset` algorithm.
     subset_handling_strategy: SubsetHandlingStrategy,
     _phantom: PhantomData<(C, N)>,
 }
@@ -60,6 +61,7 @@ where
         self
     }
 
+    /// Sets the strategy to use when handling `Subset` output.
     pub fn subset_handling_strategy(
         &mut self,
         subset_handling_strategy: SubsetHandlingStrategy,
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -8,7 +8,7 @@ use rand::{self, Rand, Rng};
 use serde::{Deserialize, Serialize};
 
 use super::{ChangeState, DynamicHoneyBadger, JoinPlan, Result, Step, VoteCounter};
-use honey_badger::HoneyBadger;
+use honey_badger::{HoneyBadger, SubsetHandlingStrategy};
 use messaging::NetworkInfo;
 use traits::{Contribution, NodeIdT};
 use util::SubRng;
@@ -21,6 +21,7 @@ pub struct DynamicHoneyBadgerBuilder<C, N> {
     /// Random number generator passed on to algorithm instance for key generation. Also used to
     /// instantiate `HoneyBadger`.
     rng: Box<dyn rand::Rng>,
+    subset_handling_strategy: SubsetHandlingStrategy,
     _phantom: PhantomData<(C, N)>,
 }
 
@@ -30,6 +31,7 @@ impl<C, N> Default for DynamicHoneyBadgerBuilder<C, N> {
         DynamicHoneyBadgerBuilder {
             max_future_epochs: 3,
             rng: Box::new(rand::thread_rng()),
+            subset_handling_strategy: SubsetHandlingStrategy::Incremental,
             _phantom: PhantomData,
         }
     }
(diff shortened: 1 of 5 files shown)


src/honey_badger/epoch_state.rs
```diff
@@ -230,54 +263,122 @@ where
...
+                    )?);
+                };
+            }
+
+            macro_rules! finalize_subset_with_step {
+                () => {{
+                    has_seen_done = true;
+                    Self::finalize_subset_with_step(subset, accepted_proposers, decryption, &mut step);
+                }};
+            }
```
afck: It's a matter of taste, but I generally prefer using macros sparingly. I find it harder to read, and I think tools like Rustfmt and Clippy do, too.

src/honey_badger/epoch_state.rs
```diff
@@ -230,54 +263,122 @@ where
...
+                                .drain(..)
+                                .map(|(proposer_id, data)| {
+                                    send_decryption_share!(proposer_id.clone(), &data);
+                                    Ok(proposer_id)
+                                }).collect::<Result<Vec<_>>>()?,
+                        );
+
+                        finalize_subset_with_step!();
+                    }
+                },
```
afck: Maybe this could be simplified—and the macros, `Self` destructuring and static methods avoided—if `SubsetHandler` had a `handle` method that takes a `SubsetOutput` and returns a `Vec<SubsetOutput>`:
* If it's `Incremental`, it returns its argument immediately.
* If it's `AllAtEnd` it stores each `Contribution` input, and then returns all of them later, when `Done` arrives.
We wouldn't need an explicit distinction here, and the code could remain almost unchanged.

ErichDonGubler: That makes sense -- I thought about doing that and decided to accept the complexity to avoid heap allocations and copying. You definitely are more familiar with what tradeoffs the codebase needs, though, so I'll make that change once I get to a dev machine.

src/honey_badger/epoch_state.rs
ErichDonGubler: The motivation here was to avoid typing out the (somewhat long) list of arguments here -- a closure would have resulted in ownership problems, and I wanted to factor out this call specifically.

That said, I agree that rolling macros in this way is definitely harder on maintenance. Do you think it'd just be better to duplicate the calls?

vkomenda: Doesn't a macro compile to a call every time you call it? It's better to call the intended function instead of hiding it behind an obscure macro.

ErichDonGubler: @vkomenda: Not sure what you mean here -- Rust will compile the content of this particular call to a function call, since that's the content of the macro, but it won't create another layer of calling a function (which I'm not sure you meant, but it sounded like). It's basically a copy-paste in any context you use it. Point taken, though -- I'll de-normalize the calls. :)

Again, the intent of the macro was to "bind" arguments without running afoul of `borrowck` -- if one were to use a lambda expression, for instance, then there'd be borrowing clashes with `accepted_proposers` because `accepted_proposers` would get used before this ever got used.

afck: I think with the approach below we should be able to have simpler code without macros and without much repetition.

src/honey_badger/epoch_state.rs
afck: I think that would be premature optimization: The bottleneck in hbbft is on way lower layers (mainly in the cryptography), so our priority should be to keep the code simple.

src/honey_badger/epoch_state.rs
vkomenda: @ErichDonGubler, please correct me if I'm wrong but two occurrences of a macro that is calling a function in Rust expand to two different function calls. In any case, the bottomline is that macros are a last resort option, and there are multiple ways you can continue without introducing a macro here. These macros increase code complexity.

ErichDonGubler: @vkomenda Yep, okay, we're on the same page. Sorry for any confusion!

ErichDonGubler:
@afck: I've changed the static method/macro binding to be `SubsetHandler::handle`, like you've suggested -- I have it returning a `struct SubsetHandleData` that also includes whether or not `Done` was found, but it essentially follows your suggestion. How does it look?

ErichDonGubler:
Rebased changes to what has been pulled in today.

src/dynamic_honey_badger/builder.rs
```diff
@@ -58,26 +60,42 @@ where
         self
     }
 
+    pub fn subset_handling_strategy(
```
afck: Please add a short doc comment explaining subset handling. (Also in `HoneyBadgerBuilder`.)

ErichDonGubler: I noticed I was missing documentation and intended to get to it today -- thank you for reminding me! :)

ErichDonGubler:
@afck: I've added doc comments written in terms of what I've learned about internals, but would it be better/more user-friendly if they were written in terms of how `{Dynamic,}HoneyBadger` uses them, instead of referring to algorithms like `EpochState` and `Subset`? I'm worried that my lack of domain knowledge doesn't let me write them to be as useful as they could/should be.

That could be saved for a future PR, of course -- just thinking out loud.

PR #254, opened 2018-10-04
Fix test-case shrinking issues

mbr:
Replaces the `NetworkDimensionStrategy`'s implementation with one based on a mapping of network dimensions to the natural numbers. This fixed panics during shrinking, making tests not exit without information when failing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ itertools = "0.7"
 serde_derive = "1.0.55"
 signifix = "0.9"
 proptest = "0.8.7"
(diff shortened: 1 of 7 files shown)


mbr:
I'm having issues finding a place in the docs where the precise requirements for `simply` and `complicate` are pinned down and cannot seem to find anything about the precise behavior of shrinking. This has kept me from implementing better suited shrinking methods; the previous iteration of that functionality did something very similar to what you described, but it was hard guaranteeing what I assumed would be the reversion of `simplify` through `complicate`.

PR #255, opened 2018-10-04
Make `DistAlgorithm` require `Send`.

mbr:
Using `hbbft` in other software that is based on frameworks like [tokio](https://tokio.rs), being `Send` is often a requirement on shared data structures.

This goes slightly beyond just making sure HoneyBadger is suitable for tokio by adding a `Send` requirement on `DistAlgorithm`. It was tempting to do so and ensures we won't accidentally break things in the future at the cost of complicating `coin.rs` slightly; we can restrict to just adding the `Box<Rng + Send>` type bounds instead if this is an issue.


mbr:
@c0gent This should hopefully suffice for parity integration; let me know if it does not please.

PR #256, opened 2018-10-04
Add `Sync` req. on boxed rngs.

c0gent:


c0gent:
Whoever sees this just approve :P

PR #257, opened 2018-10-05
Support a global RNG in transaction queue

vkomenda:
Transaction queue is the component in old tests which is not currently tested by `proptest`-based tests. This PR adds functionality which enables such tests with a given random seed.

Only the minimal support is provided in the old non-`proptest` based tests, the intention being to extend the new framework wherever possible.

### DIFF
--- a/src/dynamic_honey_badger/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger/dynamic_honey_badger.rs
@@ -38,7 +38,7 @@ pub struct DynamicHoneyBadger<C, N: Rand> {
     pub(super) incoming_queue: Vec<(N, Message<N>)>,
     /// A random number generator used for secret key generation.
     // Boxed to avoid overloading the algorithm's type with more generics.
-    pub(super) rng: Box<dyn rand::Rng + Send + Sync>,
+    pub(crate) rng: Box<dyn rand::Rng + Send + Sync>,
 }
 
 impl<C, N> fmt::Debug for DynamicHoneyBadger<C, N>
--- a/src/dynamic_honey_badger/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger/dynamic_honey_badger.rs
@@ -38,7 +38,7 @@ pub struct DynamicHoneyBadger<C, N: Rand> {
     pub(super) incoming_queue: Vec<(N, Message<N>)>,
     /// A random number generator used for secret key generation.
     // Boxed to avoid overloading the algorithm's type with more generics.
-    pub(crate) rng: Box<dyn rand::Rng + Send + Sync>,
+    pub(super) rng: Box<dyn rand::Rng + Send + Sync>,
 }
 
 impl<C, N> fmt::Debug for DynamicHoneyBadger<C, N>
--- a/src/queueing_honey_badger.rs
+++ b/src/queueing_honey_badger.rs
@@ -93,18 +93,23 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 
 /// A Queueing Honey Badger builder, to configure the parameters and create new instances of
 /// `QueueingHoneyBadger`.
-pub struct QueueingHoneyBadgerBuilder<T, N: Rand> {
+pub struct QueueingHoneyBadgerBuilder<T, N: Rand, Q> {
     /// Shared network data.
     dyn_hb: DynamicHoneyBadger<Vec<T>, N>,
     /// The target number of transactions to be included in each batch.
     batch_size: usize,
+    /// The queue of pending transactions that haven't been output in a batch yet.
+    queue: Q,
     _phantom: PhantomData<T>,
 }
 
-impl<T, N> QueueingHoneyBadgerBuilder<T, N>
+pub type QueueingHoneyBadgerWithStep<T, N, Q> = (QueueingHoneyBadger<T, N, Q>, Step<T, N, Q>);
+
+impl<T, N, Q> QueueingHoneyBadgerBuilder<T, N, Q>
 where
     T: Contribution + Serialize + for<'r> Deserialize<'r> + Clone,
     N: NodeIdT + Serialize + for<'r> Deserialize<'r> + Rand,
+    Q: TransactionQueue<T> + Debug + Default + Extend<T> + Sync + Send,
 {
     /// Returns a new `QueueingHoneyBadgerBuilder` configured to use the node IDs and cryptographic
     /// keys specified by `netinfo`.
@@ -115,6 +120,7 @@ where
         QueueingHoneyBadgerBuilder {
             dyn_hb,
             batch_size: 100,
+            queue: Default::default(),
             _phantom: PhantomData,
         }
     }
@@ -125,8 +131,14 @@ where
         self
     }
 
+    /// Sets the transaction queue object.
+    pub fn queue(mut self, queue: Q) -> Self {
+        self.queue = queue;
+        self
+    }
+
     /// Creates a new Queueing Honey Badger instance with an empty buffer.
-    pub fn build(self) -> (QueueingHoneyBadger<T, N>, Step<T, N>)
+    pub fn build(self) -> QueueingHoneyBadgerWithStep<T, N, Q>
     where
         T: Contribution + Serialize + for<'r> Deserialize<'r>,
     {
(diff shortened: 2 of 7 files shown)


src/queueing_honey_badger.rs
```diff
@@ -172,6 +171,20 @@ where
...
+where
+    T: Contribution + Serialize + for<'r> Deserialize<'r> + Debug,
+    N: NodeIdT + Serialize + for<'r> Deserialize<'r> + Rand + Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("QueueingHoneyBadger")
+            .field("batch_size", &self.batch_size)
+            .field("dyn_hb", &self.dyn_hb)
+            .field("queue", &self.queue)
+            .finish()
```
afck: Isn't that precisely the derived implementation?

src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -39,7 +39,7 @@ pub struct DynamicHoneyBadger<C, N: Rand> {
     pub(super) incoming_queue: Vec<(N, Message<N>)>,
     /// A random number generator used for secret key generation.
     // Boxed to avoid overloading the algorithm's type with more generics.
-    pub(super) rng: Box<dyn rand::Rng + Send + Sync>,
+    pub(crate) rng: Box<dyn rand::Rng + Send + Sync>,
```
afck: I'd prefer having the user pass another RNG (or the whole `TransactionQueue`?) into `QueueingHoneyBadgerBuilder` instead: that makes it more transparent which one is used for the queue and which one e.g. for DKG. (Although eventually we'll move to not storing them as fields at all anymore.)

Also, ideally each algorithm's code should serve as an example for a user who wants to build something similar: e.g. a user who wants to use a different kind of queue should be able to look at the QHB code to build their own implementation based on DHB, so it shouldn't use anything from DHB that's only available within the crate.

src/queueing_honey_badger.rs
vkomenda: Sorry, I forgot to derive it after removing an `rng` filed from `QueueingHoneyBadger`.

vkomenda:
@afck, please check if you think it's OK to pass an object of `TransactionQueue` trait to `QueueingHoneyBadgerBuilder` the way I did it.

src/transaction_queue.rs
```diff
@@ -1,35 +1,107 @@
...
+use util::SubRng;
+
+/// An interface to the transaction queue. A transaction queue is a structural part of
+/// `QueueingHoneyBadger` that manages enqueueing of transactions for a future batch and dequeueing
+/// of transactions to become part of a current batch.
+pub trait TransactionQueue<T> {
+    fn is_empty(&self) -> bool;
+    fn push_back(&mut self, t: T);
+    fn choose(&mut self, amount: usize, batch_size: usize) -> Vec<T>;
+    fn remove_all<'a, I>(&mut self, txs: I)
```
afck: I think the method docs should be here in the trait, not below in the impl. (Or both?)

src/transaction_queue.rs
```diff
@@ -1,35 +1,107 @@
...
+    where
+        I: IntoIterator<Item = &'a T>,
+        T: 'a + Contribution;
+}
 
 /// A wrapper providing a few convenience methods for a queue of pending transactions.
-#[derive(Debug)]
-pub struct TransactionQueue<T>(pub VecDeque<T>);
+pub struct VecDequeTransactionQueue<T> {
+    /// Random number generator passed on from the algorithm instance.
```
afck: I'd document what it's used for rather than where it happens to come from in the caller. Maybe: "Random number generator used for choosing transactions from the queue"?

Alternatively, we could not even store it in here, but just make it an argument of `choose`?

vkomenda: In some queues `choose` could be a deterministic choice, in others it could use a thread-local RNG, so that passing it between threads wouldn't be an issue at the trait level. So I'm leaning towards keeping `rng` internal to a particular queue implementation.

src/transaction_queue.rs
vkomenda: I chose both to be on the safe side.

src/transaction_queue.rs
```diff
@@ -1,35 +1,113 @@
...
     }
 
     /// Returns a new set of `amount` transactions, randomly chosen from the first `batch_size`.
     /// No transactions are removed from the queue.
     // TODO: Return references, once the `HoneyBadger` API accepts them. Remove `Clone` bound.
-    pub fn choose(&self, amount: usize, batch_size: usize) -> Vec<T> {
-        let mut rng = rand::thread_rng();
-        let limit = cmp::min(batch_size, self.0.len());
-        let sample = match rand::seq::sample_iter(&mut rng, self.0.iter().take(limit), amount) {
+    fn choose(&mut self, amount: usize, batch_size: usize) -> Vec<T> {
```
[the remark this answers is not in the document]
vkomenda: Well, in that case we store the RNG in `QueueingHoneyBadger`.

src/queueing_honey_badger.rs
```diff
@@ -116,6 +121,7 @@ where
         QueueingHoneyBadgerBuilder {
             dyn_hb,
             batch_size: 100,
+            queue: Default::default(),
```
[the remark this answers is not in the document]
vkomenda: If I understand correctly, the criticism applies mostly to the RNG. If the RNG is removed from the queue, the queue would be allowed to have a `Default` implementation, that is, the empty queue?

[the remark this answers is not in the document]
mbr: > If I understand correctly, the criticism applies mostly to the RNG. If the RNG is removed from the queue, the queue would be allowed to have a `Default` implementation, that is, the empty queue?

Good point! Probably no harm in an empty queue then.

src/transaction_queue.rs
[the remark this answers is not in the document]
mbr: > Well, in that case we store the RNG in `QueueingHoneyBadger`.

For now! Not having it on the queue means we have to pass it in from the outside, which at some point should be true for the full HoneyBadger. It's a good first step to use the one from `QueueingHoneyBadger` right now.

vkomenda:
I attempted to satisfy the both reviewers' comments. So, `QueueingHoneyBadger` effectively has two RNGs: one in the `DynamicHoneyBadger` instance and another provided to the `TransactionQueue::choose` method. Also, the default RNG is removed and a requirement is made to provide an RNG at construction time of `QueueingHoneyBadger`. Unfortunately, having the RNG stored in `QueueingHoneyBadger` rules out those kinds of RNG that cannot be sent or synced.

src/queueing_honey_badger.rs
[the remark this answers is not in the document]
afck: Not sure what you mean by "empty queue"? The QHB default certainly shouldn't be to drop all transactions and only output empty batches.
I'd much rather just add the `rng` parameter to `choose`.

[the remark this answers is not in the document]
vkomenda: I added that parameter to `choose`. Default is for the empty vector, no RNG attached. Is this what you are asking? Or something different?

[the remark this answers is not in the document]
afck: Ah, you're right. Sorry, I misunderstood the discussion. :+1:

vkomenda:
I rebased the PR again. @mbr, please check if your comments are satisfied.

src/transaction_queue.rs
```diff
@@ -1,32 +1,56 @@
...
-
-impl<T: Clone> TransactionQueue<T> {
+/// An interface to the transaction queue. A transaction queue is a structural part of
+/// `QueueingHoneyBadger` that manages enqueueing of transactions for a future batch and dequeueing
+/// of transactions to become part of a current batch.
+pub trait TransactionQueue<T>: fmt::Debug + Default + Extend<T> + Sync + Send {
+    /// Checks whether the queue is empty.
+    fn is_empty(&self) -> bool;
+    /// Appends an element at the end of the queue.
+    fn push(&mut self, t: T);
```
mbr: https://play.rust-lang.org/?version=stable&mode=release&edition=2015  --  Comparing the ASM output, it's almost exactly the same, except that operands are flipped on a single `cmpq` instruction =).

I still think we should use extend here and/or provide a default impl.

mbr:
I vote squash. There's litte left in the discrete commits.

PR #261, opened 2018-10-10
Fixed a condition for accepting HB messages in future epochs

vkomenda:


src/honey_badger/honey_badger.rs
```diff
@@ -125,7 +125,7 @@ where
                 .entry(epoch)
                 .or_insert_with(Vec::new)
                 .push((sender_id.clone(), content));
-        } else if epoch == self.epoch {
+        } else if self.epoch <= epoch || epoch <= self.epoch + self.max_future_epochs {
```
afck: Shouldn't that just be `if self.epoch <= epoch`? From the first condition in line 122 we already know that `self.epoch + self.max_future_epochs`. (The way it is now it looks wrong to me: `||` should be `&&`, I think?)

vkomenda: You are right. It should be an `&&` half of which is already implied by the previous check. I made a correction.

afck:
Good catch! :+1:

PR #266, opened 2018-10-11
Replace HexBytes with the hex_fmt crate.

afck:
Also move `Message` out of the `broadcast::broadcast` module.


mbr:
Looks good to me, provided it passes the CI =).

PR #269, opened 2018-10-15
Added regression for integer underflow.

mbr:
Just a maybe - ideally, we would isolate this case and add it as a test to `threshold_crypto`.


afck:
I'll add a test to `threshold_crypto` instead.
(Let's keep this open until then.)

afck:
Fix and tests are in https://github.com/poanetwork/threshold_crypto/pull/52.

PR #270, opened 2018-10-16
Add a test for the reordering attack on BinaryAgreement

### LINKED ISSUE #228: Test that the reordering attack on `BinaryAgreement` no longer succeeds
The reordering attack described in https://github.com/amiller/HoneyBadgerBFT/issues/59#issue-310368284 was avoided by adding `Conf` messages in #44, see also #37. We need a test that replays the attack and verifies that the attack does not succeed.

Closes #37.
Closes #45.

d33a94975ba60d59:
Fixes #228

### DIFF
--- a/src/binary_agreement/mod.rs
+++ b/src/binary_agreement/mod.rs
@@ -78,6 +78,7 @@ use self::bool_set::BoolSet;
 use threshold_sign;
 
 pub use self::binary_agreement::BinaryAgreement;
+pub use self::sbv_broadcast::Message as SbvMessage;
 
 /// An Binary Agreement error.
 #[derive(Clone, Eq, PartialEq, Debug, Fail)]
--- a/tests/binary_agreement_mitm.rs
+++ b/tests/binary_agreement_mitm.rs
@@ -1,0 +1,480 @@
+#![deny(unused_must_use)]
+//! Tests the BinaryAgreement protocol with a MTIM adversary.
+
+extern crate env_logger;
+extern crate failure;
+extern crate hbbft;
+extern crate integer_sqrt;
+extern crate proptest;
+extern crate rand;
+extern crate threshold_crypto;
+
+pub mod net;
+
+use std::iter;
+use std::sync::{Arc, Mutex};
+
+use hbbft::binary_agreement::{BinaryAgreement, MessageContent, SbvMessage};
+use hbbft::threshold_sign::ThresholdSign;
+use hbbft::{DistAlgorithm, NetworkInfo, Step};
+
+use net::adversary::{NetMutHandle, QueuePosition};
+use net::err::CrankError;
+use net::{Adversary, NetBuilder, NetMessage};
+
+type NodeId = usize;
+type SessionId = u8;
+type Algo = BinaryAgreement<NodeId, SessionId>;
+
+/// The state of the current epoch's coin. In some epochs this is fixed, in others it starts
+/// with in `InProgress`.
+#[derive(Debug)]
+enum CoinState<N> {
+    /// The value was fixed in the current epoch, or the coin has already terminated.
+    Decided(bool),
+    /// The coin value is not known yet.
+    InProgress(Box<ThresholdSign<N>>),
+}
+
+impl<N> CoinState<N> {
+    /// Returns the value, if this coin has already decided.
+    fn value(&self) -> Option<bool> {
+        match self {
+            CoinState::Decided(value) => Some(*value),
+            CoinState::InProgress(_) => None,
+        }
+    }
+}
+
+impl<N> From<bool> for CoinState<N> {
+    fn from(value: bool) -> Self {
+        CoinState::Decided(value)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum MessageType {
+    BVal,
+    Aux,
+    Coin,
+}
+
+fn message_type_and_content(msg: &MessageContent) -> Option<(MessageType, Option<bool>)> {
+    match msg {
+        MessageContent::SbvBroadcast(sbv_msg) => match sbv_msg {
+            SbvMessage::BVal(v) => Some((MessageType::BVal, Some(*v))),
+            SbvMessage::Aux(v) => Some((MessageType::Aux, Some(*v))),
+        },
+        MessageContent::Coin(_) => Some((MessageType::Coin, None)),
+        _ => None,
+    }
+}
+
+/// A boolean XOR a value from the state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum BoolFromState {
+    AEstimated(bool),
+    CoinValue(bool),
+}
+
+struct Stage {
+    source_groups: &'static [usize],
+    dest_groups: &'static [usize],
+    msg_type: MessageType,
+    msg_contents: Option<BoolFromState>,
+    msg_count: usize,
+}
+
+// Group IDs
+const A0: usize = 0;
+const A1: usize = 1;
+const B: usize = 2;
+const F: usize = 3;
+
+// Comments from https://github.com/amiller/HoneyBadgerBFT/issues/59#issue-310368284
+const STAGES: &[Stage] = &[
+    // x sends BVAL(\neg v) to the nodes in A0
+    Stage {
+        source_groups: &[F],
+        dest_groups: &[A0],
+        msg_type: MessageType::BVal,
+        msg_contents: Some(BoolFromState::AEstimated(true)),
+        msg_count: NODES_PER_GROUP,
+    },
+    // and BVAL(v) to the nodes in A1.
+    Stage {
+        source_groups: &[F],
+        dest_groups: &[A1],
+        msg_type: MessageType::BVal,
+        msg_contents: Some(BoolFromState::AEstimated(false)),
+        msg_count: NODES_PER_GROUP,
+    },
+    // Also, all votes from nodes in B are delivered to all nodes in A.
+    Stage {
+        source_groups: &[B],
+        dest_groups: &[A0, A1],
+        msg_type: MessageType::BVal,
+        msg_contents: None,
+        msg_count: NODES_PER_GROUP * (NODES_PER_GROUP * 2),
+    },
+    // Messages within A0 are delivered.
+    // Thus nodes in A0 see |B|+|F|=f+1 votes for \neg v;
+    // so all nodes in A0 broadcast BVAL(\neg v)
+    // and all nodes in A0 see |A0|+|B|+|F|=2f+1 votes for \neg v;
+    // so all nodes in A0 broadcast AUX(\neg v).
+    Stage {
+        source_groups: &[A0],
+        dest_groups: &[A0],
+        msg_type: MessageType::BVal,
+        msg_contents: None,
+        msg_count: NODES_PER_GROUP * (NODES_PER_GROUP - 1),
+    },
+    // Then all messages within A1 are delivered,
+    Stage {
+        source_groups: &[A1],
+        dest_groups: &[A1],
+        msg_type: MessageType::BVal,
+        msg_contents: None,
+        msg_count: NODES_PER_GROUP * (NODES_PER_GROUP - 1),
+    },
+    // as well as the BVAL(v) messages from A0 to A1.
+    // Thus the nodes in A1 see |A0|+|A1|+|F|=2f+1 votes for v and broadcast AUX(v).
+    Stage {
+        source_groups: &[A0],
+        dest_groups: &[A1],
+        msg_type: MessageType::BVal,
+        msg_contents: Some(BoolFromState::AEstimated(false)),
+        msg_count: NODES_PER_GROUP * NODES_PER_GROUP,
+    },
+    // After this all messages within A are delivered
+    Stage {
+        source_groups: &[A0, A1],
+        dest_groups: &[A0, A1],
+        msg_type: MessageType::BVal,
+        msg_contents: None,
+        msg_count: (NODES_PER_GROUP * 2) * (NODES_PER_GROUP * 2 - 1),
+    },
+    Stage {
+        source_groups: &[A0, A1],
+        dest_groups: &[A0, A1],
+        msg_type: MessageType::Aux,
+        msg_contents: None,
+        msg_count: (NODES_PER_GROUP * 2) * (NODES_PER_GROUP * 2 - 1),
+    },
+    // and x sends both BVAL(0) and BVAL(1) to every node in A.
+    // Thus every node in A broadcasts both BVAL(0) and BVAL(1) and sets bin_values=\{0,1\}.
+    Stage {
+        source_groups: &[F],
+        dest_groups: &[A0, A1],
+        msg_type: MessageType::BVal,
+        msg_contents: Some(BoolFromState::AEstimated(false)),
+        msg_count: NODES_PER_GROUP * 2,
+    },
+    Stage {
+        source_groups: &[F],
+        dest_groups: &[A0, A1],
+        msg_type: MessageType::BVal,
+        msg_contents: Some(BoolFromState::AEstimated(true)),
+        msg_count: NODES_PER_GROUP * 2,
+    },
+    // !! Not mentioned in the GitHub issue, but seems necessary.
+    // F sends Aux(_) to A, because nodes in A need 2f+1 Aux messages
+    // before they broadcast their coins.
+    Stage {
+        source_groups: &[F],
+        dest_groups: &[A0, A1],
+        msg_type: MessageType::Aux,
+        msg_contents: Some(BoolFromState::AEstimated(false)),
+        msg_count: NODES_PER_GROUP * 2,
+    },
+    // Now all nodes in A broadcast their threshold shares over the coin,
+    // so since |A|+|F|=2f+1, the adversary can construct the random coin value s.
+    Stage {
+        source_groups: &[A0, A1],
+        dest_groups: &[F],
+        msg_type: MessageType::Coin,
+        msg_contents: None,
+        msg_count: NODES_PER_GROUP * 2,
+    },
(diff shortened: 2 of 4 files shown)


tests/binary_agreement.rs
```diff
@@ -27,31 +27,139 @@ extern crate threshold_crypto as crypto;
...
+            self.sent_attack = false;
+        }
+    }
+
+    fn step(&mut self) -> Vec<MessageWithSender<BinaryAgreement<NodeId>>> {
+        let mut messages = Vec::new();
+        if !self.sent_attack {
+            self.sent_attack = true;
+            for &our_node_id in &self.known_adversarial_ids {
+                let f = self.known_node_ids.len() / 3;
```
afck: `self.known_node_ids` doesn't change inside the loop, I think? Then computing `f` could be moved outside of the loop.

tests/binary_agreement.rs
```diff
@@ -108,3 +221,9 @@ fn test_binary_agreement_first_silent() {
     let new_adversary = |_: usize, _: usize| SilentAdversary::new(MessageScheduler::First);
     test_binary_agreement_different_sizes(new_adversary);
 }
+
+#[test]
+fn test_binary_agreement_reordering() {
+    let new_adversary = |_: usize, _: usize| ReorderingAdversary::default();
+    test_binary_agreement_different_sizes(new_adversary);
```
afck: Does it really make sense for this particular test to run with `InputConfig`s other than `Split`?

tests/binary_agreement.rs
```diff
@@ -27,31 +27,139 @@ extern crate threshold_crypto as crypto;
...
 ) {
+    // The borrow checker gets in our way here, so we clone the node ids.
     let ids: Vec<NodeId> = network.nodes.keys().cloned().collect();
-    for id in ids {
-        network.input(id, input.unwrap_or_else(rand::random));
+    for (i, id) in ids.into_iter().enumerate() {
+        let value = match input {
+            InputConfig::Fixed(b) => b,
+            InputConfig::Random => rand::random(),
+            InputConfig::Split => i < (2 * network.nodes.len() / 3),
```
afck: Won't this input `true` for the first third? The comment and the code above suggest it should be the other way round.

vkomenda: This will only test constant input: `true` for the set `A` and `false` for the set `B`. There can be a second attack with these values negated. That possible attack is not covered here.

tests/binary_agreement.rs
```diff
@@ -27,31 +27,139 @@ extern crate threshold_crypto as crypto;
...
-use hbbft::binary_agreement::BinaryAgreement;
-use hbbft::NetworkInfo;
+use hbbft::binary_agreement::sbv_broadcast;
+use hbbft::binary_agreement::{BinaryAgreement, Message, MessageContent};
+use hbbft::{NetworkInfo, Target, TargetedMessage};
 
-use network::{Adversary, MessageScheduler, NodeId, SilentAdversary, TestNetwork, TestNode};
+use network::{
+    Adversary, MessageScheduler, MessageWithSender, NodeId, SilentAdversary, TestNetwork, TestNode,
+};
```
afck: We're currently in the process of transitioning from the old (`network`) to the new (`net`) test framework. I'm not sure how complicated it would be to port this to the new one?

tests/binary_agreement.rs
```diff
@@ -27,31 +27,139 @@ extern crate threshold_crypto as crypto;
...
+                                content: MessageContent::SbvBroadcast(
+                                    sbv_broadcast::Message::BVal(false),
+                                ),
+                            },
+                        },
+                    ));
+                }
+            }
+        }
+        messages
```
afck: This doesn't really reorder any messages, it just forges new messages from correct nodes, which the adversary can't do in reality. These messages may then arrive in any order (_in addition_ to the original ones, sent by the correct nodes themselves), so I don't think it reliably replicates the attack.

I think the new `net` framework provides better controls over message reordering. (Please correct me if I'm wrong, @mbr.)

Also, the attacker needs to reconstruct the coin value before anyone else, and then send the `Aux` messages accordingly.

Once the test is complete, please try (without checking it in, or maybe just in a separate temporary branch) disabling the `Conf` messages in `binary_agreement`, and use `aux_vals` itself instead of `conf_values` to try and update the epoch. With that change, this test should fail (i.e. the attack works) but all the other tests should still pass (i.e. `binary_agreement` is still working as originally described in the paper).

afck:
Ah, sorry, you already wrote that it's not an MITM adversary.
I do think it should be, though. I'd like to avoid merging "cheating" adversaries, that forge correct nodes' messages.

d33a94975ba60d59:
I'll rewrite this as a MITM adversary using the net framework if possible. I don't believe this is currently a cheating adversary, though.

src/binary_agreement/mod.rs
```diff
@@ -66,7 +66,7 @@
 mod binary_agreement;
 mod bool_multimap;
 pub mod bool_set;
-mod sbv_broadcast;
+pub mod sbv_broadcast;
```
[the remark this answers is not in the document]
afck: Yes, I don't think that algorithm on its own is generally useful and maybe it would just be confusing to expose it. Maybe we should just re-export `sbv_broadcast::Message as SbvMessage` from `binary_agreement`?

d33a94975ba60d59:
I rewrote the test using the net framework. I'll attempt to verify that it's working next (there's a good chance I still messed something up).

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,210 @@
...
+/// A0 is the first third of nodes, A1 is the second third, and the rest are B.
+#[derive(Default)]
+struct ReorderingAdversary {
+    stage: ReorderingStage,
+    stage_progress: usize,
+    sent_initial_messages: bool,
+    sent_final_messages: bool,
+}
+
+const NODES_PER_GROUP: usize = 10;
```
vkomenda: Can this value be randomized? For example, sampled randomly from a given range?

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,210 @@
...
+        _: NetMessage<BinaryAgreement<NodeId>>,
+    ) -> Result<Step<BinaryAgreement<NodeId>>, CrankError<BinaryAgreement<NodeId>>> {
+        const BVAL_FALSE_MSG: Message = Message {
+            epoch: 0,
+            content: MessageContent::SbvBroadcast(SbvMessage::BVal(false)),
+        };
+        const BVAL_TRUE_MSG: Message = Message {
+            epoch: 0,
+            content: MessageContent::SbvBroadcast(SbvMessage::BVal(true)),
+        };
```
afck: I think you can also write:
```rust
MessageContent::SbvBroadcast(SbvMessage::BVal(false)).with_epoch(epoch)
```

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    fn tamper(
+        &mut self,
+        mut net: NetMutHandle<BinaryAgreement<NodeId>>,
+        _: NetMessage<BinaryAgreement<NodeId>>,
+    ) -> Result<Step<BinaryAgreement<NodeId>>, CrankError<BinaryAgreement<NodeId>>> {
+        const BVAL_FALSE_MSG: Message = Message {
+            epoch: 0,
```
afck: I think this attack needs to be repeated in every epoch, not just `0`.

tests/binary_agreement_mitm.rs
afck: On the other hand this is just a regression test for one specific case. I wonder whether we should just set this to `1`?

vkomenda:
I agree with @afck, we possibly need `self.conf_values = Some(aux_vals)` before the call to `try_finish_conf_round`. This is what `send_conf` would have done. Thanks for the correction!

In that case changing `send_conf` instead is better. Just remove the last line `self.send(MessageContent::Conf(values))` and adjust the returned `Result` value accordingly.

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,210 @@
...
+
+#[test]
+fn reordering_attack() {
+    let _ = env_logger::try_init();
+    let ids: Vec<NodeId> = (0..NUM_NODES).collect();
+    let mut net = NetBuilder::new(ids.iter().cloned())
+        .adversary(ReorderingAdversary::default())
+        .crank_limit(10000)
+        .using(|info| {
+            BinaryAgreement::new(Arc::new(info.netinfo), 0, info.id)
```
afck: The third argument ("`proposer_id`") needs to be the same in all nodes.

That's indeed a confusing API, of course, and we should definitely change it. But for now, you can just replace `info.id` with `0`.

d33a94975ba60d59:
It's now passing, but failing with this patch:

```patch
diff --git a/src/binary_agreement/binary_agreement.rs b/src/binary_agreement/binary_agreement.rs
index 5beac06..e1c1265 100644
--- a/src/binary_agreement/binary_agreement.rs
+++ b/src/binary_agreement/binary_agreement.rs
@@ -247,7 +247,8 @@ impl<N: NodeIdT> BinaryAgreement<N> {
             return Ok(self.try_finish_conf_round()?);
         }

-        self.send(MessageContent::Conf(values))
+        //self.send(MessageContent::Conf(values))
+        Ok(Step::default())
     }

     /// Multicasts and handles a message. Does nothing if we are only an observer.
```

@afck pointed out that I'm still missing the last part of the attack, so I'm not sure why this is happening.

afck:
Just not sending the `Conf` messages will make it always fail. You have to make @vkomenda's change and in addition set `self.conf_values = Some(aux_vals)`, as discussed above. (I.e. you need to change it so it implements ABA as described in the paper.)

d33a94975ba60d59:
@afck The function `send_conf` sets those values right above the line I commented out :)

As @vkomenda said,

> In that case changing send_conf instead is better. Just remove the last line self.send(MessageContent::Conf(values)) and adjust the returned Result value accordingly

And that's what I did

afck:
But if you apply only the above 1-line patch, without adding the `try_finish_conf_round`, doesn't that make _all_ tests fail? Doesn't it simply make `BinaryAgreement` never output?

afck:
But if you apply only the above 1-line patch, without adding the `try_finish_conf_round`, doesn't that make _all_ tests fail? Doesn't it simply make `BinaryAgreement` never output?

afck:
But if you apply only the above 1-line patch, without adding the `try_finish_conf_round`, doesn't that make _all_ tests fail? Doesn't it simply make `BinaryAgreement` never output?

afck:
But if you apply only the above 1-line patch, without adding the `try_finish_conf_round`, doesn't that make _all_ tests fail? Doesn't it simply make `BinaryAgreement` never output?

vkomenda:
@d33a94975ba60d59 , sorry for the misunderstanding. I was too brief. We need to call `try_finish_conf_round` without sending any `Conf` messages. You can do so by adding ` self.conf_values = Some(aux_vals)` to my earlier patch as @afck suggested.

vkomenda:
@d33a94975ba60d59 , sorry for confusion. I wasn't descriptive enough. You would need to return `Ok(self.try_finish_conf_round()` from `send_conf`, not the empty `Step` as you did. Alternatively you can use the patch I sent with the correction made by @afck.

vkomenda:
@d33a94975ba60d59 , sorry for confusion. I wasn't descriptive enough. You would need to return `Ok(self.try_finish_conf_round()` from `send_conf`, not the empty `Step` as you did. Alternatively you can use the patch I sent with the correction made by @afck.

tests/binary_agreement_mitm.rs
mbr: In that case, making it random will answer both questions :P

d33a94975ba60d59:
This has been much more complicated than I expected, but once again, I think it's working. It's passing, but failing with `crank: node failed to process step: CrankLimitExceeded(10000)` with this patch:

```patch
diff --git a/src/binary_agreement/binary_agreement.rs b/src/binary_agreement/binary_agreement.rs
index c9ff256..5d9414e 100644
--- a/src/binary_agreement/binary_agreement.rs
+++ b/src/binary_agreement/binary_agreement.rs
@@ -187,7 +187,9 @@ impl<N: NodeIdT> BinaryAgreement<N> {
                 }
                 CoinState::InProgress(_) => {
                     // Start the `Conf` message round.
-                    step.extend(self.send_conf(aux_vals)?)
+                    //step.extend(self.send_conf(aux_vals)?)
+                    self.conf_values = Some(aux_vals);
+                    step.extend(self.try_finish_conf_round()?)
                 }
             }
         }
@@ -347,9 +349,11 @@ impl<N: NodeIdT> BinaryAgreement<N> {
 
     /// Checks whether the _N - f_ `Conf` messages have arrived, and if so, activates the coin.
     fn try_finish_conf_round(&mut self) -> Result<Step<N>> {
+        /*
         if self.conf_values.is_none() || self.count_conf() < self.netinfo.num_correct() {
             return Ok(Step::default());
         }
+        */
 
         // Invoke the coin.
         let coin_step = match self.coin_state {

```

The BinaryAgreement gets to around epoch 50 before the crank limit is exceeded, and logging the estimated values for each node, it seems clear to me that the attack is now working.

I've also rewritten this to use a much more modular/configurable system. The STAGES array defines the attack, and the rest of the code just carries it out.

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,440 @@
...
+        }
+    }
+
+    fn stage_matches_msg(&self, message: &NetMessage<BinaryAgreement<NodeId>>) -> bool {
+        if let Some(stage) = STAGES.get(self.stage) {
+            let from = *message.from();
+            let src_group = if from == 0 {
+                3
+            } else {
+                (from - 1) / NODES_PER_GROUP
```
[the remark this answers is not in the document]
d33a94975ba60d59: It makes other parts a bit more complicated, like iterating over the correct algorithms. I think having the faulty node first is fine.

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,440 @@
...
+struct ReorderingAdversary {
+    stage: usize,
+    stage_progress: usize,
+    sent_stage_messages: bool,
+    epoch: u32,
+    coin_state: CoinState<NodeId>,
+    /// The estimated value for nodes in A.
+    a_estimated: bool,
+    // TODO this is really hacky but there's no better way to get this value
+    netinfo_mutex: Arc<Mutex<Option<Arc<NetworkInfo<NodeId>>>>>,
```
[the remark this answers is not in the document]
d33a94975ba60d59: Getting the networkinfo into the adversary is a surprisingly difficult task. The closure that gets the netinfo must be created after the adversary is created, but cannot have a mutable reference to the adversary because it needs to be moved into the NetBuilder. I tried to extract the adversary from the NetBuilder in the closure, but with only a mutable reference to a box I can't downcast it to my adversary type. If I could convert it to an Any it'd work, but even making Adversary a super-trait requiring Any, I run into https://github.com/rust-lang/rust/issues/5665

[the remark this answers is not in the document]
d33a94975ba60d59: Another option is having the Adversary type as a type parameter of NetBuilder. This would make creating a network without an adversary more difficult, though we could just have a helper for that using NullAdversary. What do you think about that?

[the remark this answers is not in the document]
vkomenda: Is `netinfo_mutex` only used to get the `invocation_id`? If so, don't use `NetworkInfo` inside the adversary. Pass only data that are needed to construct the nonce, or, in the current master, [`coin_id`](https://github.com/poanetwork/hbbft/blob/5f21c9a4e49acc9a95e72ba13ef1f594030c8f24/src/binary_agreement/binary_agreement.rs#L326).

tests/binary_agreement_mitm.rs
```diff
@@ -0,0 +1,506 @@
...
+}
+
+const NODES_PER_GROUP: usize = 2;
```
vkomenda: Please randomize the number of nodes in a group. The attack targets groups of any size. The tests might try a few sizes in a range `[1..max_size]` for some `max_size`.

vkomenda: On a second thought, no randomization would replace formal analysis because it would still generate finitely many tests. Since this is a regression test for a particular message ordering, a constant group size suffices here.

d33a94975ba60d59:
Rebased. I added Algo and SessionId typedefs in the process.

tests/binary_agreement_mitm.rs
d33a94975ba60d59: FYI, in my testing, I never noticed group sizes having an effect on how the attack worked.

tests/binary_agreement_mitm.rs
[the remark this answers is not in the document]
vkomenda: Yes, you are right. Another option would be to clone the `NetworkInfo` of the `NewNodeInfo`. It's read-only in `BinaryAgreement`, so it should be fine to clone it at the time of building the `VirtualNet`.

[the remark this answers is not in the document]
d33a94975ba60d59: That's what's currently done :)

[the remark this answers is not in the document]
vkomenda: If you clone, you don't need the enclosing `Arc<Mutex<Option<_>>>` in the constructor argument of the adversary.

vkomenda:
Looks fine by me. Please update to Rust 1.30.0.

d33a94975ba60d59:
> Please update to Rust 1.30.0.

What do you mean by that? Did I miss a `dyn` or something? I'm on Rust 1.30.0. The Travis failure was a formatting error from my rebase.

vkomenda:
Sorry, I meant the formatting errors. The PR is ready for merging :)

PR #274, opened 2018-10-16
Added a NetworkInfo getter trait

vkomenda:
I added `HasNetworkInfo` to access the `netinfo` field of a `DistAlgorithm` if it has one. This can be used in the generic sender queue.


c0gent:
You may want to consider renaming this trait as it goes against conventions (recently summed up nicely here: https://github.com/rust-lang-nursery/api-guidelines/issues/28#issuecomment-418509628).

The current name gives the impression that it's method returns a bool.

c0gent:
Well the method is named `netinfo`...

vkomenda:
Maybe just cloning the `Arc<NetworkInfo<N>>` into the sender queue is a better idea. There isn't anything that the queue has to access that is too large to be cloned, so I think I'll clone instead of adding the getter.

vkomenda:
Closing this since the feature is not required.

PR #276, opened 2018-10-18
Avoid BTreeMap in EpochState.

afck:
Profiling the simulation example (with mock crypto) identified this as a hot spot, and using a `Vec` instead and returning early where possible speeds it up.

### DIFF
--- a/src/honey_badger/epoch_state.rs
+++ b/src/honey_badger/epoch_state.rs
@@ -48,14 +48,6 @@ where
             DecryptionState::Complete(_) => Ok(td::Step::default()),
         }
     }
-
-    /// Returns the plaintext, if it has already been decrypted.
-    fn plaintext(&self) -> Option<&[u8]> {
-        match self {
-            DecryptionState::Ongoing(_) => None,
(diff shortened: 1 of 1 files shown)


src/honey_badger/epoch_state.rs
```diff
@@ -265,13 +257,13 @@ where
...
-            .iter()
-            .flat_map(|(id, dec_state)| dec_state.plaintext().map(|pt| (id.clone(), pt)))
-            .collect();
-        if !proposer_ids.iter().eq(plaintexts.keys()) {
-            return None; // Not all accepted contributions are decrypted yet.
+        let mut plaintexts = Vec::new();
+        // Collect accepted plaintexts. Return if some are not decrypted yet.
+        for id in proposer_ids {
+            match self.decryption.get(id) {
+                None | Some(DecryptionState::Ongoing(_)) => return None,
```
vkomenda: I'm not sure the returned value of `push` is relevant. You could make this branch return `()` instead and discard the returned value of `push` on the next line.

vkomenda: Everything is fine. I misunderstood this line at first.

PR #277, opened 2018-10-18
Changed the defaults for tracing to `false`.

mbr:
Saves valuable harddrive space!

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -102,7 +102,7 @@ Adversaries can be introduced through the `.adversary` method on the constructor
 
 By default, all network tests write traces of every network message into logfiles, named `net-trace_*.txt` in the current working directory. Each log stores one message per line, in the format of `[SENDER] -> [RECEIVER]: MSG`.
 
-This behavior can be controlled using the `HBBFT_TEST_TRACE` environment variable; if set and equal to `0` or `false`, this functionality is disabled. Tracing is disabled by default.
+This behavior can be controlled using the `HBBFT_TEST_TRACE` environment variable; if set and equal to `1` or `true`, this functionality is enabled. Tracing is disabled by default.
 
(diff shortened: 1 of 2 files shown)


tests/README.md
```diff
@@ -102,7 +102,7 @@ Adversaries can be introduced through the `.adversary` method on the constructor
 
 By default, all network tests write traces of every network message into logfiles, named `net-trace_*.txt` in the current working directory. Each log stores one message per line, in the format of `[SENDER] -> [RECEIVER]: MSG`.
 
-This behavior can be controlled using the `HBBFT_TEST_TRACE` environment variable; if set and equal to `0` or `false`, this functionality is disabled. Tracing is enabled by default.
+This behavior can be controlled using the `HBBFT_TEST_TRACE` environment variable; if set and equal to `0` or `false`, this functionality is disabled. Tracing is disabled by default.
```
afck: It should probably say "`1`" and "`true`" and "enabled" in the first sentence now.

tests/net/mod.rs
```diff
@@ -502,7 +502,7 @@ where
 
         let trace = self.trace.unwrap_or_else(|| {
             // If the trace setting is not overriden, we use the setting from the environment.
-            let setting = env::var("HBBFT_TEST_TRACE").unwrap_or_else(|_| "true".to_string());
+            let setting = env::var("HBBFT_TEST_TRACE").unwrap_or_else(|_| "false".to_string());
```
afck: We could avoid creating another string:

```rust
env::var("HBBFT_TEST_TRACE").map_or_else(false, |setting| setting == "true" || setting == "1")
```

mbr: We could, I considered it and chalked it up to copying it verbatim from elsewhere. Shame on me!

PR #278, opened 2018-10-20
Use published threshold_crypto 0.2.

afck:
Also: Follow the soon-to-be-official formatting guidelines:
https://github.com/nrc/rfcs/blob/style-guide/style-guide/cargo.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,18 +1,20 @@
 [package]
 name = "hbbft"
 version = "0.1.0"
-authors = ["Vladimir Komendantskiy <komendantsky@gmail.com>",
-           "Andreas Fackler <AndreasFackler@gmx.de>",
-           "Peter van Nostrand <jnz@riseup.net>",
-           "Andrew Gross <andogro@gmail.com>",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,18 +1,20 @@
...
+]
+categories = ["algorithms", "asynchronous", "cryptography", "network-programming"]
+keywords = ["consensus", "asynchronous", "threshold"]
 license = "MIT/Apache-2.0"
-repository = "https://github.com/poanetwork/hbbft"
 readme = "README.md"
-keywords = ["consensus", "asynchronous", "threshold"]
-categories = ["algorithms", "asynchronous", "cryptography", "network-programming"]
+repository = "https://github.com/poanetwork/hbbft"
+description = "Honey Badger Byzantine fault tolerant consensus algorithm"
```
c0gent: Capitalization

afck: Good point. How about:
"The Honey Badger of Byzantine Fault Tolerant Consensus Protocols"?
That would also be closer to the paper's title.

c0gent: Yup :)

PR #287, opened 2018-10-23
Extend API to work without DistAlgorithm.

afck:
Extends all APIs so that the user doesn't need to import the
`DistAlgorithm` trait.

Also, removes the error on inputting too late into `BinaryAgreement`:
whether the input still matters is an implementation detail.


afck:
I agree, there's still a lot of work to do in hbbft in terms of documentation.

c0gent:
> I agree, there's still a lot of work to do in hbbft in terms of documentation
> 
> *afck merged commit bc6a67d into master...*

That's usually my line...

Issue #209, opened 2018-08-14
Make threshold encryption optional.


afck:
Threshold encryption of batches adds a round of messages (although this could be partly ameliorated with #170) and is very CPU intensive (see https://github.com/amiller/HoneyBadgerBFT/issues/9).

We could add a parameter to enable encryption only in a certain percentage of rounds, or "every _n_-th" round. In DHB we could even make that dynamic (#103), so that the network can adapt if censorship is suspected.

### Implementation

As a first step, a new `encryption_schedule` configuration option should be added to `[Dynamic]HoneyBadger[Builder]`. I'd suggest an enum with variants `Always`, `Never` and `EveryNthEpoch(n)`; or should it be a tuple instead that specifies how many encrypted epochs, followed by how many unencrypted ones?

`HoneyBadger` then needs to decide, based on the epoch number, whether to encrypt or not, and initialize `EpochState` accordingly. `EpochState` needs to be updated to be able to skip the encryption and decryption steps.

Let's also extend the [simulation example](https://github.com/poanetwork/hbbft/blob/master/examples/simulation.rs) to make the new parameter configurable there, and test whether it actually improves performance.

The second step would be to make it dynamic: We'd probably want a new [`Change`](https://github.com/poanetwork/hbbft/blob/20193ab009104f46f89448673e118c908450bc23/src/dynamic_honey_badger/change.rs#L5) variant `EncryptionSchedule`. It would allow validators to vote for encryption schedule changes in the same way they vote for adding or removing validators.

Once such a change has a majority, no key generation needs to be initialized. We'd just restart `HoneyBadger` with the new parameter.

afck:
@ejmg offered his help with this and will be working on it. Thank you! :+1:

logannc:
I started looking at this today. So far I've mostly gotten the configuration sorted out, including dynamically changing the encryption schedule (I probably need to commit some analog to `KeyGenMessage` when we update, but I'm deferring that for the moment).

What I'm having a little trouble identifying is how to handle `EpochState`. I'm a little fuzzy on how everything ties together. Part of this is that I haven't yet read the paper so I'm a little weak on the primitives involved. I'll read up on that, but I was hoping I could get a bit of direction until then.

[...]

PR #288, opened 2018-10-24
Optional threshold encryption 209

logannc:
@afck  #209 

I'm sure I've missed something, but tests are passing with Always and Never, so I think it's ready for review. :)

Running the simulation example with encryption on my machine takes 43s (the simulated time is 12932). Without encryption, those numbers drop to 0.355s (w/ sim at 8153). Every N, for a few arbitrary values of N: N=2 (22s, 10546), N=5 (11s, 9852), N=8 (6s, 8751). I didn't bother running for `TickTock`, which is N encrypted then M unencrypted.

The changes to `Change` were a bit more invasive than I'd have liked, but generally cleans up places where it's used because the `NodeChange` case and `EncryptionSchedule` case are generally not both dealt with.

Something I didn't do was move the encryption into the `EpochState` because that would have required passing in the RNG, which seemed undesireable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,3 @@ src/proto/message.rs
 **/*.rs.bk
 Cargo.lock
 net-trace_*.txt
-TODOs
-.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ src/proto/message.rs
 **/*.rs.bk
 Cargo.lock
 net-trace_*.txt
+TODOs
+.gitignore
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 
 use crypto::{SecretKey, SecretKeySet, SecretKeyShare};
 use rand::{self, Rand, Rng};
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Serialize};
 
 use super::{Change, ChangeState, DynamicHoneyBadger, JoinPlan, Result, Step, VoteCounter};
 use honey_badger::{HoneyBadger, SubsetHandlingStrategy};
@@ -46,8 +46,8 @@ impl<C, N> Default for DynamicHoneyBadgerBuilder<C, N> {
 
 impl<C, N> DynamicHoneyBadgerBuilder<C, N>
 where
-    C: Contribution + Serialize + for<'r> Deserialize<'r>,
-    N: NodeIdT + Serialize + for<'r> Deserialize<'r> + Rand,
+    C: Contribution + Serialize + DeserializeOwned,
+    N: NodeIdT + Serialize + DeserializeOwned + Rand,
 {
     /// Returns a new `DynamicHoneyBadgerBuilder` configured to use the node IDs and cryptographic
     /// keys specified by `netinfo`.
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -9,9 +9,9 @@ use serde::{Deserialize, Serialize};
 
 use super::{Change, ChangeState, DynamicHoneyBadger, JoinPlan, Result, Step, VoteCounter};
 use honey_badger::{HoneyBadger, SubsetHandlingStrategy};
+use threshold_decryption::EncryptionSchedule;
 use util::SubRng;
 use {Contribution, NetworkInfo, NodeIdT};
-use threshold_decryption::EncryptionSchedule;
 
 /// A Dynamic Honey Badger builder, to configure the parameters and create new instances of
 /// `DynamicHoneyBadger`.
@@ -163,11 +163,9 @@ where
             rng: Box::new(self.rng.sub_rng()),
         };
         let step = match join_plan.change {
-            ChangeState::InProgress(ref change) => {
-                match change {
-                    Change::NodeChange(change) => dhb.update_key_gen(join_plan.epoch, change)?,
-                    _ => Step::default(),
-                }
+            ChangeState::InProgress(ref change) => match change {
+                Change::NodeChange(change) => dhb.update_key_gen(join_plan.epoch, change)?,
+                _ => Step::default(),
             },
             ChangeState::None | ChangeState::Complete(..) => Step::default(),
         };
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -150,7 +150,7 @@ where
         let arc_netinfo = Arc::new(netinfo.clone());
         let honey_badger = HoneyBadger::builder(arc_netinfo.clone())
             .max_future_epochs(self.max_future_epochs)
-            .encryption_schedule(self.encryption_schedule)
+            .encryption_schedule(join_plan.encryption_schedule)
             .build();
         let mut dhb = DynamicHoneyBadger {
             netinfo,
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -149,6 +149,7 @@ where
         let arc_netinfo = Arc::new(netinfo.clone());
         let honey_badger = HoneyBadger::builder(arc_netinfo.clone())
             .max_future_epochs(self.max_future_epochs)
+            .encryption_schedule(self.encryption_schedule)
             .build();
         let mut dhb = DynamicHoneyBadger {
             netinfo,
@@ -164,7 +165,7 @@ where
         let step = match join_plan.change {
             ChangeState::InProgress(ref change) => {
                 match change {
-                    Change::Add(_,_) | Change::Remove(_) => dhb.update_key_gen(join_plan.epoch, change)?,
+                    Change::NodeChange(change) => dhb.update_key_gen(join_plan.epoch, change)?,
                     _ => Step::default(),
                 }
             },
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -102,6 +102,7 @@ where
         let max_future_epochs = *max_future_epochs;
         let arc_netinfo = Arc::new(netinfo.clone());
         let honey_badger = HoneyBadger::builder(arc_netinfo.clone())
+            .session_id(epoch)
             .max_future_epochs(max_future_epochs)
             .rng(rng.sub_rng())
             .subset_handling_strategy(subset_handling_strategy.clone())
--- a/src/dynamic_honey_badger/builder.rs
+++ b/src/dynamic_honey_badger/builder.rs
@@ -7,10 +7,11 @@ use crypto::{SecretKey, SecretKeySet, SecretKeyShare};
 use rand::{self, Rand, Rng};
 use serde::{Deserialize, Serialize};
 
-use super::{ChangeState, DynamicHoneyBadger, JoinPlan, Result, Step, VoteCounter};
+use super::{Change, ChangeState, DynamicHoneyBadger, JoinPlan, Result, Step, VoteCounter};
 use honey_badger::{HoneyBadger, SubsetHandlingStrategy};
 use util::SubRng;
 use {Contribution, NetworkInfo, NodeIdT};
+use threshold_decryption::EncryptionSchedule;
 
 /// A Dynamic Honey Badger builder, to configure the parameters and create new instances of
 /// `DynamicHoneyBadger`.
@@ -24,6 +25,8 @@ pub struct DynamicHoneyBadgerBuilder<C, N> {
     rng: Box<dyn rand::Rng>,
     /// Strategy used to handle the output of the `Subset` algorithm.
     subset_handling_strategy: SubsetHandlingStrategy,
+    /// Schedule for adding threshold encryption to some percentage of rounds
+    encryption_schedule: EncryptionSchedule,
     _phantom: PhantomData<(C, N)>,
 }
 
@@ -35,6 +38,7 @@ impl<C, N> Default for DynamicHoneyBadgerBuilder<C, N> {
             max_future_epochs: 3,
             rng: Box::new(rand::thread_rng()),
             subset_handling_strategy: SubsetHandlingStrategy::Incremental,
+            encryption_schedule: EncryptionSchedule::Always,
             _phantom: PhantomData,
         }
     }
@@ -78,6 +82,12 @@ where
         self
     }
 
+    /// Sets the schedule to use for threshold encryption.
+    pub fn encryption_schedule(&mut self, encryption_schedule: EncryptionSchedule) -> &mut Self {
+        self.encryption_schedule = encryption_schedule;
+        self
+    }
+
     /// Creates a new Dynamic Honey Badger instance with an empty buffer.
     pub fn build(&mut self, netinfo: NetworkInfo<N>) -> DynamicHoneyBadger<C, N> {
         let DynamicHoneyBadgerBuilder {
@@ -85,6 +95,7 @@ where
             max_future_epochs,
             rng,
             subset_handling_strategy,
+            encryption_schedule,
             _phantom,
         } = self;
         let epoch = *epoch;
@@ -94,6 +105,7 @@ where
             .max_future_epochs(max_future_epochs)
             .rng(rng.sub_rng())
             .subset_handling_strategy(subset_handling_strategy.clone())
+            .encryption_schedule(*encryption_schedule)
             .build();
         DynamicHoneyBadger {
             netinfo,
@@ -150,7 +162,12 @@ where
             rng: Box::new(self.rng.sub_rng()),
         };
         let step = match join_plan.change {
-            ChangeState::InProgress(ref change) => dhb.update_key_gen(join_plan.epoch, change)?,
+            ChangeState::InProgress(ref change) => {
+                match change {
+                    Change::Add(_,_) | Change::Remove(_) => dhb.update_key_gen(join_plan.epoch, change)?,
+                    _ => Step::default(),
+                }
+            },
             ChangeState::None | ChangeState::Complete(..) => Step::default(),
         };
         Ok((dhb, step))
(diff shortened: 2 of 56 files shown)


logannc:
:man_facepalming:  I should have read the contributing guide first.

.gitignore
```diff
@@ -3,3 +3,5 @@ src/proto/message.rs
 **/*.rs.bk
 Cargo.lock
 net-trace_*.txt
+TODOs
+.gitignore
```
afck: I don't think we want git to ignore the `.gitignore`?

src/dynamic_honey_badger/change.rs
```diff
@@ -1,30 +1,49 @@
...
+        }
+    }
+}
+
+/// A node change action: adding or removing a node.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Hash, Debug)]
+pub enum Change<N> {
+    // Add or Remove a node from the set of validators
+    NodeChange(NodeChange<N>),
+    /// Change the threshold encryption schedule. i.e., to increase frequency to prevent censorship or decrease for performance.
```
afck: Please add a line break: maximum width is 100 characters.

src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -321,18 +327,34 @@ where
...
     }
 
+    pub(super) fn update_encryption_schedule(
+        &mut self,
+        epoch: u64,
+        change: &Change<N>,
+    ) -> Result<Step<C, N>> {
+        self.restart_honey_badger(epoch);
+        if let Change::EncryptionSchedule(schedule) = change {
+            self.honey_badger.encryption_schedule = *schedule;
```
afck: I'd prefer adding that to `restart_honey_badger` as an argument. Let's not change fields while an algorithm is already running, and add it to the `HoneyBadgerBuilder` instead.

src/dynamic_honey_badger/mod.rs
```diff
@@ -149,14 +149,14 @@ struct KeyGenState<N> {
     /// The key generation instance.
     key_gen: SyncKeyGen<N>,
     /// The change for which key generation is performed.
-    change: Change<N>,
-    /// The number of key generation messages received from each peer. At most _N + 1_ are
+    change: NodeChange<N>,
+    /// The number of key generation messages received from the candidate. At most _N² + 1_ are
```
afck: Sorry, I think that "`²`" was a mistake. I changed it to _N + 1_ yesterday.

src/honey_badger/honey_badger.rs
```diff
@@ -101,13 +105,19 @@ where
...
+        let prop = if require_decryption {
+            let ciphertext = self
+                .netinfo
+                .public_key_set()
+                .public_key()
+                .encrypt_with_rng(&mut self.rng, ser_prop);
+            bincode::serialize(&ciphertext).map_err(|err| ErrorKind::ProposeBincode(*err))?
+        } else {
+            ser_prop
+        };
```
afck: I'd prefer to move that part of the logic entirely into `EpochState`.
(Not your fault: It was already inconsistent in the first place to do the serialization and encryption in `HoneyBadger` and the deserialization and decryption in `EpochState`.)
Let's just input the `proposal` itself into `EpochState::propose`.

logannc: This will require accepting or adding an RNG to `EpochState`. Thoughts?

logannc: That's what I meant by 'accepting'. Does it matter if the RNG comes from `HoneyBadger`? IIRC, Rust doesn't have a good idiom for default/optional parameters yet. I'll probably end up doing this, it just feels like there should be a way to do it without passing in an RNG that is only sometimes used... It feels like we're going to end up with duplication somewhere no matter what though.

.gitignore
logannc: tbh, I committed this just to push my work so I could work from a different machine and forgot to remove it. In my personal projects, I actually like to have it ignored so I can ignore junk in my working directory without seeing `.gitignore` as one of the changed files. You can always `git add -f` if you actually need to commit a change to `.gitignore`, which is rare.

But I'll remove it. :)

src/honey_badger/honey_badger.rs
afck: Yes, it needs to come from `HoneyBadger`: We went to great lengths to make sure that we can configure all sources of randomness and have deterministic tests. Passing in a reference to the RNG is not expensive: It's no problem if it isn't always used.

logannc: Hm, running into a borrow checker issue with:

```rust
let epoch_state = self.epoch_state_mut(epoch)?;
let rng = &mut self.rng;
epoch_state.propose(ser_prop, rng)?
```

I think I've got it sorted, but I'm not sure if there is a more idiomatic way to appease rust. Running tests then I'll push.

logannc: I also might have an issue where clippy is telling me to do something but the compiler is unhappy when I do... Clippy might just be wrong on this one. I'm not sure if there is a way to tell it to ignore this specific case?

src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -288,11 +290,16 @@ where
...
+                ChangeState::Complete(Change::NodeChange(kgs.change))
             } else if let Some(change) = self.vote_counter.compute_winner().cloned() {
                 // If there is a new change, restart DKG. Inform the user about the current change.
-                step.extend(self.update_key_gen(batch_epoch + 1, &change)?);
+                step.extend(match &change {
+                    Change::NodeChange(change) => self.update_key_gen(batch_epoch + 1, &change)?,
+                    Change::EncryptionSchedule(schedule) => {
+                        self.update_encryption_schedule(batch_epoch + 1, *schedule)?
+                    }
+                });
```
afck: Shouldn't we go straight to `ChangeState::Complete` for this? Changes other than `NodeChange` doesn't need a key generation period—after the restart of Honey Badger, it's done—, so I wonder whether the `InProgress` variant should only take a `NodeChange` argument in the first place, instead of a `Change`?

src/dynamic_honey_badger/mod.rs
```diff
@@ -149,14 +149,14 @@ struct KeyGenState<N> {
     /// The key generation instance.
     key_gen: SyncKeyGen<N>,
     /// The change for which key generation is performed.
-    change: Change<N>,
-    /// The number of key generation messages received from each peer. At most _N + 1_ are
+    change: NodeChange<N>,
+    /// The number of key generation messages received from the candidate. At most _N + 1_ are
```
afck: Sorry, the other change in this comment on master should also stay. We're tracking the key generation message count for each peer now, instead of just the candidate.

src/honey_badger/epoch_state.rs
```diff
@@ -205,15 +209,27 @@ where
...
 
     /// If the instance hasn't terminated yet, inputs our encrypted contribution.
-    pub fn propose(&mut self, ciphertext: &Ciphertext) -> Result<Step<C, N>> {
-        let ser_ct =
-            bincode::serialize(ciphertext).map_err(|err| ErrorKind::ProposeBincode(*err))?;
-        let cs_step = self.subset.handle_input(ser_ct)?;
+    pub fn propose(
+        &mut self,
+        proposal: Vec<u8>,
+        rng: &mut Box<dyn Rng + Send + Sync>,
```
afck: I think Clippy is right here: why not make this a `&mut dyn Rng + Send + Sync`?

src/honey_badger/epoch_state.rs
```diff
@@ -205,15 +209,27 @@ where
...
     }
 
     /// If the instance hasn't terminated yet, inputs our encrypted contribution.
-    pub fn propose(&mut self, ciphertext: &Ciphertext) -> Result<Step<C, N>> {
-        let ser_ct =
-            bincode::serialize(ciphertext).map_err(|err| ErrorKind::ProposeBincode(*err))?;
-        let cs_step = self.subset.handle_input(ser_ct)?;
+    pub fn propose(
+        &mut self,
+        proposal: Vec<u8>,
```
afck: Sorry, I meant to also move the serialization (not only of the ciphertext, but also of the proposal in the first place) into `EpochState`: It is already generic in the type `C` and handles deserialization, and this would make it more consistent. So it should probably be `proposal: &C` here.

src/honey_badger/epoch_state.rs
```diff
@@ -307,7 +323,11 @@ where
             } = self.subset_handler.handle(cs_output);
 
             for (k, v) in contributions {
-                step.extend(self.send_decryption_share(k.clone(), &v)?);
+                step.extend(if self.require_decryption {
+                    self.send_decryption_share(k.clone(), &v)?
+                } else {
+                    self.process_decryption(k.clone(), td::Step::default().with_output(v))?
```
afck: Let's just directly

```rust
self.decryption.insert(k.clone(), DecryptionState::Complete(v));
```

instead of instantiating a threshold decryption `Step`.

src/honey_badger/honey_badger.rs
```diff
@@ -34,6 +35,8 @@ pub struct HoneyBadger<C, N: Rand> {
     pub(super) rng: Box<dyn Rng + Send + Sync>,
     /// Represents the optimization strategy to use for output of the `Subset` algorithm.
     pub(super) subset_handling_strategy: SubsetHandlingStrategy,
+    /// The schedule for which rounds we should use threshold encryption.
+    pub(crate) encryption_schedule: EncryptionSchedule,
```
afck: I'd prefer making this `pub(super)`, like the other fields, and adding a getter for it:
The idea is that the algorithm modules can be used independently, and the user should be able to implement all of the `DynamicHoneyBadger` functionality by themselves, if needed, outside of this crate.

src/dynamic_honey_badger/dynamic_honey_badger.rs
logannc: Probably.  We could certainly make that change, though I wouldn't be surprised about needing to reverse the change to `InProgress` later. I'll change the former, let me know if you feel strongly about the `InProgress` change.

src/honey_badger/epoch_state.rs
logannc: If the function signature is `&mut (dyn Rng + Send + Sync)` then I need `.encrypt_with_rng(&mut Box::from(rng), ...)` later instead of just `.encrypt_with_rng(rng, ...)` because Rust complains when the argument to `encrypt_with_rng(...)` does not implement `Sized`.

I am a little fuzzy on the difference between the two. As I understand it, as it currently is, we have a reference to a pointer (I use 'pointer' here, and afterwards, loosely) on the stack which points to a heap allocated `Rng` trait object. Under the new signature, we would receive a reference to a stack allocated `Rng` trait object, then we create a heap allocated pointer TO that stack object, then we create a stack allocated reference to that heap allocated pointer and THAT is the argument we give to `encrypt_with_rng`.... 

From an efficiency standpoint, the former seems better (though I haven't gone through Godbolt or anything to see what it'd actually compile down to for release). Are there any ergonomics concerns that would offset that? Or, alternatively, is there another way to appease Rust that doesn't involve `Box`ing the reference argument?

src/dynamic_honey_badger/dynamic_honey_badger.rs
afck: I don't think we'll ever have `InProgress` for another change: DKG only needs to run if we changed the set of validators, but not for any parameter changes. The only thing I can think of is that we may want to regenerate our keys periodically, even without a validator change, but we could just add a `NodeChange::NewKeys` for that later, if needed.

On the other hand, the API is possibly more intuitive if both variants contain the same `Change` type.

@vkomenda, @c0gent: I'm fine with either. Let me know if you have a preference.

src/honey_badger/epoch_state.rs
afck: You can also omit the `Send + Sync` from the signature (sorry, I didn't think of that above): just `&mut Rng`, like all the other methods do that take an RNG argument, including the ones in the `rand` crate itself.

And you shouldn't need pass in a pointer to a pointer at all. `encrypt_with_rng(&mut *self.rng, …)` would borrow the RNG from its box, and only pass in a pointer to the RNG itself, I think?

src/honey_badger/honey_badger.rs
afck: I think the problem here is that `epoch_state_mut`, being a method on `self`, borrows _all_ of self and conflicts with the borrowed `rng`. Your workaround looks reasonable. (It's unfortunate that it's necessary, of course.)

src/honey_badger/epoch_state.rs
logannc: If I'm doing `&mut *self.rng` then I'm leaving the function signature of `propose` as `&mut Box<...>` instead of `&mut Rng` so we'd be ignoring clippy and this would all be moot anyway. (This also doesn't borrow the Rng from the box because we have `&mut Box<...>` so `&mut *rng` types back to `&mut Box<...>`.)

To reiterate, if I use `&mut Rng` in the function signature, it complains about `dyn rang::Rng` being unsized at the callsite of `encrypt_with_rng(...)`. Related, I'm not 100% clear how passing a `&mut Box<dyn Rng>` works in the first place since `encrypt_with_rng<R: Rng>(..., rng: R)` and AFAIK there isn't an `impl<T: Rng> Rng for Box<T>`. I guess somehow from the `Borrow` impl on `Box`  and then it autoderefs inside `encrypt_with_rng(...)`?

In any case, `encrypt_with_rng` requires something sized. AFAIK, that means we have a concrete type or a boxed trait object, right? So, unless there are ergonomic reasons that we'd prefer for `propose(...)` to accept an unboxed trait object, we might as well ignore clippy and require a boxed trait because we're gunna have to box it anyway.

src/dynamic_honey_badger/dynamic_honey_badger.rs
logannc: Note to self: changing this to `Complete` is what broke the tests.

src/honey_badger/epoch_state.rs
afck: Right… but this works for me:

```rust
fn print_n<R: Rng>(rng: &mut R) {
    let n: u8 = rng.gen();
    println!("Random number: {}", n);
}

fn print_n_using_boxed(mut rng: Box<dyn Rng + Sync + Send>) {
    print_n(&mut rng);
}
```

So it looks like the function signature still doesn't need a `Box`, but you can just pass in `&mut self.rng`?

logannc: This may all be moot because clippy is no longer reporting this as an issue in the recently released stable 1.30. So consider the rest of this more curiosity than necessity.

Where/how do you get that to work? https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=5026f41cd25e55fff07cc2fbd4c961de

1. I can't seem to get a direct translation to work in the playground using the current code as a base. It says that `Rng` is not implemented for `Box<dyn Rng>`. Which makes sense, I don't really understand why the current code compiles.
2. When I use `&mut dyn Rng` in the function signature, I get errors about it being unsized, which is what I've been saying this whole time. AFAIK, this is because `print_n` has an implicit `Sized` bound on `rng` which is why we need `Box`. And since it's defined in `threshold_encryption`, we can't change it inside this PR.

afck: Sorry, you're right, I mixed that up: let's just not make `propose`'s argument a fat pointer at all, just `&mut R` with a generic `R: Rng`. That's what all the other methods do, and it works with just `&mut self.rng`, I think.

src/dynamic_honey_badger/builder.rs
```diff
@@ -138,6 +150,7 @@ where
         let arc_netinfo = Arc::new(netinfo.clone());
         let honey_badger = HoneyBadger::builder(arc_netinfo.clone())
             .max_future_epochs(self.max_future_epochs)
+            .encryption_schedule(self.encryption_schedule)
```
afck: Actually, I'm wondering whether the encryption schedule needs to be part of the `JoinPlan`, since the joining node must use the same one as the others.

src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -356,7 +384,11 @@ where
             .session_id(epoch)
             .max_future_epochs(self.max_future_epochs)
             .rng(self.rng.sub_rng())
-            .build();
+            .encryption_schedule(if let Some(schedule) = encryption_schedule {
+                schedule
+            } else {
+                self.honey_badger.get_encryption_schedule()
```
afck:
```rust
encryption_schedule.or_else(|| self.honey_badger.get_encryption_schedule())
```

src/honey_badger/epoch_state.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "cargo-clippy", allow(borrowed_box))]
```
afck: That probably isn't necessary anymore.

src/dynamic_honey_badger/builder.rs
logannc: Ah, I see. That's what I had intended, just got it wrong and didn't revisit it.

src/dynamic_honey_badger/change.rs
```diff
@@ -1,31 +1,51 @@
...
+}
+
+/// A node change action: adding or removing a node.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Hash, Debug)]
+pub enum Change<N> {
+    // Add or Remove a node from the set of validators
+    NodeChange(NodeChange<N>),
+    /// Change the threshold encryption schedule.
+    /// Increase frequency to prevent censorship or decrease frequency for increased throughput.
+    EncryptionSchedule(EncryptionSchedule),
```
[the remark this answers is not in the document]
logannc: I didn't remove `None` from `Change` nor did I place `None` inside any variant of `Change`. The only `None` involved is `ChangeState::None`, which I think is what you're talking about regarding `HoneyBadger` being restarted (from `fn join_plan(...)` in `dynamic_honey_badger/batch.rs`.

Please let me know if I've misunderstood.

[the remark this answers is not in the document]
vkomenda: Yes, you are absolutely right! In that case everything is OK.

PR #289, opened 2018-10-24
Make the BA session ID generic.

afck:


src/binary_agreement/binary_agreement.rs
```diff
@@ -39,10 +43,13 @@ impl<N> From<bool> for CoinState<N> {
...
-    session_id: u64,
-    /// The ID of the proposer of the value for this Binary Agreement instance.
-    proposer_id: N,
+    // We store the session ID twice: once in its serialized form, to create the `ThresholdSign`
+    // nonces, and once boxed, for debug output.
+    /// Session identifier, to prevent replaying messages in other instances,
+    /// e.g, the Honey Badger algorithm epoch plus the proposer ID.
+    session_id: Box<Debug + Send + Sync>,
+    /// The serialized session identifier.
+    ser_session_id: Vec<u8>,
```
afck: Alternatives to this stunt include:
* Add `T` as a parameter to `BinaryAgreement` and its `Step`, which is pretty ugly.
* Use [erased-serde](https://crates.io/crates/erased-serde) and store clones of the ID in two separate boxes: `Box<erased_serde::Serialize>` and `Box<Debug>`.
* Create a new, boxable `trait SerializeDerive` with: `fn serialize(&self) -> Result<Vec<u8>, bincode::Error>` and `fn debug(&self) -> &Debug`.
* The obvious, elegant solution that I just can't come up with.

afck:
As discussed, we'll introduce generic session IDs instead (as an additional type parameter) for `BinaryAgreement` and `Subset`, but not for for the Honey Badgers themselves. That way we don't couple the layers (i.e. we remove the notions of "proposer ID" and "Honey Badger epoch" from `BinaryAgreement`) without unnecessarily complicating the user of DHB and HB. Users of `BinaryAgreement` and `Subset` can pick anything `Serialize + Display` as their session ID.

PR #292, opened 2018-10-24
Minor improvements to the Step API.

afck:


src/traits.rs
```diff
@@ -107,6 +106,12 @@ where
         self.messages.extend(other.messages);
     }
 
+    /// Extends this step with `other` and returns the result.
+    pub fn and(mut self, other: Self) -> Self {
```
mbr: `#[inline]`?

mbr: Why do we need this in addition to `extend_with`?

mbr: Isn't `and` very misleading here? There's no "falsy" `Step`, is there?

afck: I have trouble coming up with a better name. `union`? `with`?
I'm also happy with not adding it at all, if you think it's not worth it. I just found so many places where we would first `extend` and then return that I thought it might be handy.

vkomenda: Probably not `union` because `extend` is not commutative as a binary operation. Either `and` or `with` sound OK. Might be `append` as well.

mbr: `append` is for appending single elements; `with` is weird to me as a Python programmer ;).

`merge`? `combine`?

mbr:
> I'm also happy with not adding it at all, if you think it's not worth it. I just found so many places where we would first extend and then return that I thought it might be handy.

I guess this is an example?

```diff
-            let mut step = self.handle_sbvb_step(sbvb_step)?;
-            step.extend(self.handle_conf(sender_id, BoolSet::from(b))?);
-            Ok(step)
+            let step = self.handle_sbvb_step(sbvb_step)?;
+            Ok(step.and(self.handle_conf(sender_id, BoolSet::from(b))?))
```

I'm probably not writing enough `tokio` code, that's I why I find explicit three-line version a easier to read at the moment. I'd be fine with adding the convenience function, iff we find a better name for it than `and` though.

src/traits.rs
vkomenda: List append is an operation on two lists that returns a concatenation of the first list and second list. It's much older than Python, and Python got it differently (and wrong, to me as a functional programmer) with `extend` and `append`. Let's not use `append` to conform to the naming convention.

`merge` has a reordering aspect. `combine` or `combined` sound like the output doesn't preserve the exact shape of inputs. `join` is a synonym of "or" rather than `and`: https://en.wikipedia.org/wiki/Join_and_meet

I think `join` is the closest match out of the ones you listed.

afck:
I renamed it to `join` for now… happy to discuss tomorrow; I'm fine with either: removing it, or finding a better name.

PR #294, opened 2018-10-24
Fixed early output from Subset

### LINKED ISSUE #282: Remove synchrony in HB Epoch State
At the moment, `EpochState` expects decryption shares only for "accepted proposers", that is, nodes found in the output of `Subset`. If the order of messages changes within the same epoch, early decryption shares are discarded with a fault report. "Accepted proposers" aren't really necessary and removing them would allow to reorder messages without breaking HB output.

vkomenda:
Fixes #282 

Let me know if you think a specialized test is needed. This bug is only covered by tests in https://github.com/poanetwork/hbbft/tree/vk-extern-sender-queue. It should be possible to trigger the bug (and the fault) by sending _some_ `Aux` messages before the corresponding `Proof` messages, so that `Agreement` instances finish before _some_ `Broadcast` instances.

### DIFF
--- a/src/subset.rs
+++ b/src/subset.rs
@@ -311,7 +311,11 @@ impl<N: NodeIdT + Rand> Subset<N> {
                     }
                 }
             }
-            if let Some(Some(value)) = self.broadcast_results.insert(proposer_id.clone(), None) {
+            if let Some(value) = self
+                .broadcast_results
+                .get_mut(proposer_id)
+                .and_then(Option::take)
+            {
                 debug!("    {:?} → {:0.10}", proposer_id, HexFmt(&value));
                 step.output
                     .extend(Some(SubsetOutput::Contribution(proposer_id.clone(), value)));


src/subset.rs
```diff
@@ -322,6 +332,26 @@ impl<N: NodeIdT + Rand> Subset<N> {
...
+            Some(BroadcastResult::Taken) => None,
+            None => {
+                remove_insertion = true;
+                None
+            }
+        };
+        if remove_insertion {
+            self.broadcast_results.remove(proposer_id);
+        }
+        value
```
afck: If you add a `BroadcastResult::take` method (that replaces the instance with `Taken` and returns the `Option<ProposedValue>`), this method could just be

```rust
self.broadcast_results.get_mut(proposer_id).and_then(BroadcastResult::take)
```

without cloning the ID, and without inserting and removing again.

afck: …but then we might as well just use `Option` again, instead of `BroadcastResult`, which is structurally equivalent anyway?

vkomenda:
@afck, your solution is much better. I amended the commit.

PR #295, opened 2018-10-25
Fix abbreviations from the paper: ABA.

afck:
The Binary Agreement is mostly referred to as "ABA" in the [paper](https://eprint.iacr.org/2016/199.pdf).
(Unfortunately not consistently: the box with the message flow just says "BA".)

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,12 +142,12 @@ Our implementation modifies the protocols described in "[The Honey Badger of BFT
 
 We have simplified algorithm naming conventions from the original paper.
 
-|  Algorithm Name  | Original Name                    |
-| ---------------- | -------------------------------- |
-| Honey Badger     | HoneyBadgerBFT                   |
-| Subset           | Asynchronous Common Subset (ACS) |
-| Broadcast        | Reliable Broadcast (RBC)         |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,12 +142,12 @@ Our implementation modifies the protocols described in "[The Honey Badger of BFT
...
-| Honey Badger     | HoneyBadgerBFT                   |
-| Subset           | Asynchronous Common Subset (ACS) |
-| Broadcast        | Reliable Broadcast (RBC)         |
-| Binary Agreement | Binary Byzantine Agreement (BBA) |
+|  Algorithm Name  | Original Name                                 |
+| ---------------- | --------------------------------------------- |
+| Honey Badger     | HoneyBadgerBFT                                |
+| Subset           | Asynchronous Common Subset (ACS)              |
+| Broadcast        | Reliable Broadcast (RBC)                      |
+| Binary Agreement | Asynchronous Binary Byzantine Agreement (ABA) |
```
vkomenda: Do we need the unfortunate abbreviation? It's not a standard one, is it?

This abbreviation should be ABBA.

afck: I think the table is mainly to make it easier to connect the paper with the code. And "ABA" is used quite a few times throughout the paper.

PR #297, opened 2018-10-25
Include advice on how to handle `Step` in docs.

mbr:
These doc additions capture what we currently consider best practices when handling algorithm output.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -10,7 +10,7 @@ use serde::Serialize;
 use fault_log::{Fault, FaultLog};
 use TargetedMessage;
 
-/// A transaction, user message, etc.
+/// A transaction, user message, or other user data.
 pub trait Contribution: Eq + Debug + Hash + Send + Sync {}
 impl<C> Contribution for C where C: Eq + Debug + Hash + Send + Sync {}
 
@@ -26,11 +26,11 @@ impl<M> Message for M where M: Debug + Send + Sync {}
 pub trait SessionIdT: Display + Serialize + Send + Sync + Clone {}
 impl<S> SessionIdT for S where S: Display + Serialize + Send + Sync + Clone {}
 
-/// Single algorithm step result.
+/// Single algorithm step outcome.
 ///
 /// Each time input (typically in the form of user input or incoming network messages) is provided
-/// to an instance of an algorithm, a `Step` is produced as a result, potentially containing
-/// output values, fault log and network messages.
+/// to an instance of an algorithm, a `Step` is produced, potentially containing output values,
+/// a fault log, and network messages.
 ///
 /// Any `Step` **must always be used** by the client application; at the very least the resulting
 /// messages must be queued.
@@ -38,7 +38,7 @@ impl<S> SessionIdT for S where S: Display + Serialize + Send + Sync + Clone {}
 /// ## Handling unused Steps
 ///
 /// In the (rare) case of a `Step` not being of any interest at all, instead of discarding it
-/// through `let _ = ...` or similar constructs, the implicit assumption should explicitliy be
+/// through `let _ = ...` or similar constructs, the implicit assumption should explicitly be
 /// checked instead:
 ///
 /// ```rust,no_run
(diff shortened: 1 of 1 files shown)


src/traits.rs
```diff
@@ -21,8 +21,28 @@ impl<N> NodeIdT for N where N: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
...
+///
+/// ## Handling unused Steps
+///
+/// In the (rare) case of a `Step` not being of any interest at all, instead of discarding it
+/// through `let _ = ...` or similar constructs, the implicit assumption should explicitliy be
+/// checked instead:
+///
+/// ```rust,norun
+/// assert!(alg.propose(123).expect("Could not propose value").is_empty(),
+///         "Algorithm will never output anything on first proposal");
```
afck: I know it's just for the example, but: If we are the only validator, the first proposal usually _will_ cause an immediate output.

src/traits.rs
```diff
@@ -21,8 +21,28 @@ impl<N> NodeIdT for N where N: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
 pub trait Message: Debug + Send + Sync {}
 impl<M> Message for M where M: Debug + Send + Sync {}
 
-/// Result of one step of the local state machine of a distributed algorithm. Such a result should
-/// be used and never discarded by the client of the algorithm.
+/// Single algorithm step result.
```
vkomenda: Did you mean this to be a header? How about
```
## Result of a single algorithm step
```
?

src/traits.rs
mbr: I know, I found out the hard way! `alg` is neutral though, doesn't have to be HoneyBadger =).

I can't think of an example where it would be valid to not handle the step at the moment.

src/traits.rs
mbr: No, it's the docstring for `struct Step`. It's just cut off in the diff view =).

vkomenda: Still might be worthwhile to normalize this docstring. It's a bit misleading to refer to "result" because `Step` does happen to occur as `Result<Step<..>>` sometimes.
What do you think, @andogro ?

andogro: Result seems like the most clear word to explain this. If you think it causes confusion for the user because of Result<Step<..>>  would changing 'result' to 'outcome' make it more obvious?

src/traits.rs
```diff
@@ -21,8 +21,28 @@ impl<N> NodeIdT for N where N: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
...
+///
+/// Each time input (typically in the form of user input or incoming network messages) is provided
+/// to an instance of an algorithm, a `Step` is produced as a result, potentially containing
+/// output values, fault log and network messages.
```
andogro: a fault log,

src/traits.rs
```diff
@@ -21,8 +21,28 @@ impl<N> NodeIdT for N where N: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
...
+///
+/// Any `Step` **must always be used** by the client application; at the very least the resulting
+/// messages must be queued.
+///
+/// ## Handling unused Steps
+///
+/// In the (rare) case of a `Step` not being of any interest at all, instead of discarding it
+/// through `let _ = ...` or similar constructs, the implicit assumption should explicitliy be
```
andogro: explicitly

src/traits.rs
```diff
@@ -21,8 +21,28 @@ impl<N> NodeIdT for N where N: Eq + Ord + Clone + Debug + Hash + Send + Sync {}
...
+/// checked instead:
+///
+/// ```rust,no_run
+/// assert!(alg.propose(123).expect("Could not propose value").is_empty(),
+///         "Algorithm will never output anything on first proposal");
+/// ```
+///
+/// Should an edge case occur resulting in outgoing messages to be generated, the `assert!` will
```
andogro: If an edge case occurs and outgoing messages are generated as a result, the ...

src/traits.rs
mbr: I like 'outcome', I'll change it.

src/traits.rs
mbr: I was not sure what kind of entities the fault log contains and whether it's technically a single log or multiple logs. I decided to cheat by leaving it ambivalent ("[fault log and network] messages" vs "[fault log] and [network messages]").

Changing it now, @afck will correct me if it's wrong =).

mbr:
Updated and rebased.

src/traits.rs
afck: Our `FaultLog` type contains several `Fault`s, each of which blames a `FaultKind` on a `NodeId`.

I'm fine with "a fault log". :+1:

PR #298, opened 2018-10-25
Make the BA session ID generic.

afck:
This change also fixes a subtle security issue: We were potentially reusing the same coins in some rare cases!

If we are in epoch 41 and are just about to e.g. finish voting on a change, and `max_future_epochs == 2`, an attacker might send contributions for epochs 42 and 43 in the current (old) `HoneyBadger` instance. This would trigger the other nodes to start corresponding `BinaryAgreement`s and potentially flip "HB epoch 42" and "43" coins in them.
Once epoch 41 actually concludes, everyone starts a new `HoneyBadger` instance in epoch 42, where the same coins will be flipped again, and the attacker now knows a few values in advance.

That's probably not critical at all, but it's now addressed: We include the `start_epoch` in the coin name.

### DIFF
--- a/src/binary_agreement/binary_agreement.rs
+++ b/src/binary_agreement/binary_agreement.rs
@@ -1,12 +1,16 @@
 use std::collections::BTreeMap;
+use std::fmt::{self, Display};
+use std::result;
 use std::sync::Arc;
+
+use bincode;
 
 use super::bool_multimap::BoolMultimap;
 use super::bool_set::BoolSet;
 use super::sbv_broadcast::{self, SbvBroadcast};
-use super::{Error, Message, MessageContent, Nonce, Result, Step};
+use super::{Error, Message, MessageContent, Result, Step};
 use threshold_sign::{self, ThresholdSign};
-use {DistAlgorithm, NetworkInfo, NodeIdT, Target};
+use {DistAlgorithm, NetworkInfo, NodeIdT, SessionIdT, Target};
 
 /// The state of the current epoch's coin. In some epochs this is fixed, in others it starts
 /// with in `InProgress`.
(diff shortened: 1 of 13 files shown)


src/binary_agreement/binary_agreement.rs
```diff
@@ -321,24 +317,19 @@ impl<N: NodeIdT> BinaryAgreement<N> {
...
             1 => CoinState::Decided(false),
             _ => {
-                let nonce = Nonce::new(
-                    self.netinfo.invocation_id().as_ref(),
-                    self.session_id,
-                    self.netinfo.node_index(&self.proposer_id).unwrap(),
-                    self.epoch,
-                );
-                CoinState::InProgress(Box::new(ThresholdSign::new(self.netinfo.clone(), nonce)))
+                let coin_id = bincode::serialize(&(&self.session_id, self.epoch))?;
```
mbr: Did we sidestep the issue by just using an anonymous type in the form of a tuple to get the epoch info in there?

src/honey_badger/builder.rs
```diff
@@ -35,6 +38,7 @@ where
     pub fn new(netinfo: Arc<NetworkInfo<N>>) -> Self {
         HoneyBadgerBuilder {
             netinfo,
+            session_id: 0,
```
mbr: Is it ever a good idea to have a slightly insecure default here? We could make it a mandatory argument and force people to pass in `0` explicitly. What's the most common use case?

src/honey_badger/epoch_state.rs
```diff
@@ -376,3 +382,17 @@ where
         }
     }
 }
+
+/// A session identifier for a `Subset` sub-algorithm run within an epoch. It consists of the epoch
+/// number, and an optional `HoneyBadger` session identifier.
```
mbr: Did our abstraction leak here again? =)

src/honey_badger/builder.rs
afck: I'm not sure: As soon as you generate new key shares every time you start Honey Badger, you're fine with session ID 0. If you restart a network, but in a new epoch, you're also fine.

And even if you restart the exact same network at epoch 0, all the attacker could do would be delay consensus slightly more effectively than in the first run: If e.g. in epoch 0, one of the coins was flipped twice, they'll now know two coin values in advance, but from the third one, they won't, so they won't be able to stall consensus indefinitely anyway.

I'm fine with either: making it mandatory or not.

src/honey_badger/epoch_state.rs
afck: This is inside the `honey_badger` module, so I think it's fine to refer to its session identifier?

src/honey_badger/builder.rs
c0gent: I'm not 100% sure about this either but I think it's probably better left with a 0 default.

c0gent:
You guys weren't quick enough to sneakily merge this :P

PR #299, opened 2018-10-25
Use the correct NetworkInfos in the broadcast ProposeAdversary.

afck:


tests/broadcast.rs
```diff
@@ -65,31 +62,17 @@ impl Adversary<Broadcast<NodeId>> for ProposeAdversary {
...
-            .map(|msg| MessageWithSender::new(id, msg))
-            .collect()
+            .flat_map(|(&id, netinfo)| {
+                Broadcast::new(netinfo.clone(), id)
+                    .expect("broadcast instance")
+                    .handle_input(b"Fake news".to_vec())
+                    .expect("propose")
+                    .messages
+                    .into_iter()
+                    .map(move |msg| MessageWithSender::new(id, msg))
```
vkomenda: This still doesn't use `step.output`. Probably it's OK but don't we need to keep a note?

afck: Not really: I don't know why we had added that comment. After all, the adversary is just running fake nodes internally here, to create the fake messages to send.

PR #307, opened 2018-10-29
Instead of macro_use, use use.

afck:


mbr:
Does this in any way affect `rustfmt`? Afaik, `rustfmt` does not reorder imports if they have `#[macro_use]` in front, because the order in which macros are imported actually matters.

I am guessing the CI will catch any issue at the very least, still it's something to maybe keep in mind =).

PR #308, opened 2018-10-29
External sender queue implementation

### LINKED ISSUE #226: Dynamic Honey Badger sender queue

vkomenda:
Fixes #226 

This PR implements a reference spam control mechanism - the _sender queue_ - by requiring all peers implement post-processing of outgoing messages as follows:

-  Peer epochs are monitored.

- An outgoing message is sent to a peer if and when the sender reckons the peer is at an epoch that matches the epoch of the message. If the peer is already ahead, the message is discarded by the sender instead of being sent.

- `Target::All` messages may be split into `Target::Node` messages for each peer if peers happen to be at different epochs, some of which match the epoch of the message while others don't.

Incoming message queues have been removed. Message recipients discard obsolete messages and return a `Fault` for messages with epochs too far ahead.

The sender queue is agnostic to the differences between validators and observers. All it cares about is what the peer IDs are and what is the ID of the node it is running on. I believe this is approximately in line with Option 2 of @afck's [comment](https://github.com/poanetwork/hbbft/issues/226#issuecomment-423935571).

There are three implementations of a sender queue, one for each of `HoneyBadger`, `DynamicHoneyBadger` and `QueueingHoneyBadger`, although the last one uses the `DynamicHoneyBadger` implementation, so those two are essentially the same.

**Comments welcome:** I'm not sure whether `Epoched::Epoch` is fine to stay `Copy` rather than `Clone`. Maybe I should have implemented an optional storage for the computed epoch and make `Epoched::Epoch` to be `Clone` instead to avoid possible recomputation of the epoch of the same message multiple times.


c0gent:
Vlad, could you provide a brief (but complete) outline of the changes you've made just for a frame of reference? I haven't been following the this topic closely so it would be helpful for me but it would also make it easier for us in the future if we ever look back at this. I usually like to put a description or list of changes in the commit message but here in the PR is fine too.

vkomenda:
@c0gent, I updated the description as you asked.

c0gent:
Excellent description!

src/dynamic_honey_badger/mod.rs
```diff
@@ -111,19 +115,54 @@ pub enum Message<N: Rand> {
...
             Message::SignedVote(ref signed_vote) => signed_vote.era(),
         }
     }
+}
+
+/// Dynamic Honey Badger epoch. It consists of an era and an epoch of Honey Badger that started in
+/// that era. For messages originating from `DynamicHoneyBadger` as opposed to `HoneyBadger`, that
+/// HoneyBadger epoch is `None`.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, Serialize, Deserialize)]
+pub struct Epoch(pub(super) (u64, Option<u64>));
```
afck: I'd prefer avoiding the double brackets everywhere:

```rust
pub struct Epoch(pub(super) u64, pub(super) Option<u64>);
```

src/dynamic_honey_badger/sender_queue.rs
```diff
@@ -0,0 +1,48 @@
...
+                None
+            },
+        )
+    }
+
+    fn is_accepting_epoch(&self, us: &Message<N>, Epoch((them_era, them_hb_epoch)): Epoch) -> bool {
+        let Epoch((era, hb_epoch)) = us.epoch();
+        era == them_era
+            && (them_hb_epoch <= hb_epoch
+                && hb_epoch <= them_hb_epoch.map(|e| e + self.max_future_epochs as u64))
```
afck: Shouldn't we accept it even if `hb_epoch.is_none()` but `them_hb_epoch.is_some()`? If I'm not mistaken, `Option<u64>` is ordered "lexicographically", and `None` is just smaller than every `Some(_)`. Also, `them_hb_epoch` is always `Some(_)`, isn't it?

So I would expect something like:

```rust
if era != them_era {
    return false;
}
match (hb_epoch, them_hb_epoch) {
    (Some(us), Some(them)) => them <= us && us <= them + self.max_future_epochs as u64,
    (None, Some(them)) => true,
    (_, None) => error!("Peer's Honey Badger epoch unknown."),
}
```

~But actually, if it's always `Some(_)`, it would be better to make that clear statically, and just have method parameters `them_era: u64, them_hb_epoch: u64` directly?~ (Edit: Not possible, of course, because `Epoch` is not always the same type… :confused: )

src/dynamic_honey_badger/sender_queue.rs
```diff
@@ -0,0 +1,48 @@
+use std::cmp;
+
+use rand::Rand;
+use serde::{de::DeserializeOwned, Serialize};
+
+use super::{Batch, Change, ChangeState, DynamicHoneyBadger, Epoch, Message, Step};
+use sender_queue::SenderQueueFunc;
+use {Contribution, Epoched, NodeIdT};
+
+impl<C, N> SenderQueueFunc<DynamicHoneyBadger<C, N>> for DynamicHoneyBadger<C, N>
```
afck: I wonder whether the trait should be called something like `Queueable` or `SenderQueueable`?
In any case, I'd make the file name the same as the trait's.

src/sender_queue/message.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Message<M>
+where
+    M: Epoched,
+{
+    pub epoch: <M as Epoched>::Epoch,
```
afck: Isn't that redundant? Could we just rename `MessageContent` to `Message`, and give `EpochStarted` and `Epoch` argument?

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// Returns a pair containing
+    ///
+    /// - the maximum epoch out of a given epoch and the epoch of a given batch,
+    ///
+    /// - an optional new peer to be added to the set of peers.
+    fn max_epoch_with_batch(
+        &self,
+        epoch: <D::Message as Epoched>::Epoch,
+        batch: &D::Output,
+    ) -> (<D::Message as Epoched>::Epoch, Option<D::NodeId>);
```
afck: This method seems to do two things at once. Maybe it would be cleaner to only return an `Option<D::NodeId>`, and to implement `Epoched` for the `Step`s? So instead of

```rust
let (max_e, added) = alg.max_epoch_with_batch(epoch, batch);
```

the caller would write:

```rust
let max_e = epoch.max(batch.epoch());
let added = alg.added_node(&batch);
```

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+    D::Message: Clone + Epoched + Serialize + DeserializeOwned,
+    D::NodeId: NodeIdT + Rand,
+{
+    type NodeId = D::NodeId;
+    type Input = D::Input;
+    type Output = D::Output;
+    type Message = Message<D::Message>;
+    type Error = D::Error;
+
+    fn handle_input(&mut self, input: Self::Input) -> Result<Step<D>, D> {
```
afck: Even for the `SenderQueue`, it would be nice to be usable without importing `DistAlgorithm`. Let's move the implementations of `handle_input` and `handle_message` to the `impl<D> SenderQueue<D>` below, and just call those from here.

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+        sender_id: &D::NodeId,
+        epoch: <D::Message as Epoched>::Epoch,
+    ) {
+        let earlier_keys: Vec<_> = self
+            .outgoing_queue
+            .keys()
+            .cloned()
+            .filter(|(id, this_epoch)| {
+                id == sender_id
+                    && PartialOrd::partial_cmp(this_epoch, &epoch) == Some(Ordering::Less)
```
afck: Isn't that equivalent to `this_epoch < epoch`? I think `<` doesn't evaluate to `true` if `partial_cmp` returns `None`?

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn process_new_epoch(
+        &mut self,
+        sender_id: &D::NodeId,
+        epoch: <D::Message as Epoched>::Epoch,
+    ) -> Step<D> {
+        // Send any HB messages for the HB epoch.
+        let mut ready_messages = self
+            .outgoing_queue
+            .remove(&(sender_id.clone(), epoch))
+            .unwrap_or_else(|| vec![]);
```
afck: `unwrap_or_default` seems to be stable now. (Alternatively, `unwrap_or_else(Vec::new)` would also avoid the closure.)

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+        let mut updated = false;
+        // Look up `DynamicHoneyBadger` epoch updates and collect any added peers.
+        let (new_epoch, mut new_peers) = step.output.iter().fold(
+            (self.epoch, BTreeSet::new()),
+            |(epoch, mut new_peers), batch| {
+                let (max_epoch, new_peer) = self.algo.max_epoch_with_batch(epoch, batch);
+                if max_epoch != epoch {
+                    updated = true;
+                }
+                new_peer.map(|peer| new_peers.insert(peer));
```
afck: Maybe `new_peers.extend(new_peer)` would be simpler.

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+                (max_epoch, new_peers)
+            },
+        );
+        self.epoch = new_epoch;
+        self.peer_ids.append(&mut new_peers);
```
afck: Would it make sense to set their epoch to the next one right away?

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+        &self,
+        epoch: <D::Message as Epoched>::Epoch,
+        batch: &D::Output,
+    ) -> (<D::Message as Epoched>::Epoch, Option<D::NodeId>);
+
+    /// Whether the epoch `them` accepts the message `us`.
+    fn is_accepting_epoch(&self, us: &D::Message, them: <D::Message as Epoched>::Epoch) -> bool;
+
+    /// Whether the epoch `them` is ahead of the epoch of the message `us`.
+    fn is_later_epoch(&self, us: &D::Message, them: <D::Message as Epoched>::Epoch) -> bool;
```
afck: Maybe `is_message_obsolete`? Otherwise I'm never sure whether the message's or the peer's epoch is the "later" one.

tests/dynamic_honey_badger.rs
```diff
@@ -75,8 +77,8 @@ where
             .iter()
             .filter(|(_, node)| {
                 node_busy(*node)
-                    && !node.instance().has_input()
-                    && node.instance().netinfo().is_validator()
+                    && !node.instance().algo().has_input()
+                    && node.instance().algo().netinfo().is_validator()
```
afck: Maybe we should add these as convenience methods to `SenderQueue` (`has_input` only for `SenderQueue<DHB>` and `SenderQueue<HB>`, of course).

src/dynamic_honey_badger/batch.rs
```diff
@@ -1,14 +1,16 @@
...
+use super::{ChangeState, Epoch, JoinPlan};
 use {NetworkInfo, NodeIdT};
 
 /// A batch of transactions the algorithm has output.
 #[derive(Clone, Debug)]
 pub struct Batch<C, N> {
     /// The sequence number: there is exactly one batch in each epoch.
     pub(super) epoch: u64,
+    /// The next `DynamicHoneyBadger` epoch.
+    pub(super) next_epoch: Epoch,
```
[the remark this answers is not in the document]
vkomenda: Yes if we change `Batch::epoch` to be `Epoch` rather than `u64`. `next_epoch` should be an `Epoch`.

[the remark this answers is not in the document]
afck: Ah, you're right, the `era` is missing… but maybe it would be cleaner to include `era: u64` in `Batch`, instead of `next_epoch` then.

afck:
I just realized that we probably need to include `max_future_epochs` in the `JoinPlan` now (or, alternatively, include more information about which epochs we accept in the `EpochStarted` message). Otherwise a node with `max_future_epochs == 3` will send messages to one with `max_future_epochs == 2` that it won't accept yet.

With this and #288, it would make sense to add a mechanism that on network startup and for joining nodes ensures that the nodes all agree on those parameters (but probably in a separate PR).

src/sender_queue/mod.rs
vkomenda: Having `Batch::next_epoch` would help because in fact we need
```
let max_e = epoch.max(batch.next_epoch);
```
It's not the epoch of the batch we are after but its next epoch.

vkomenda:
I reckon including `max_future_epochs` in every `EpochStarted` message would be a bit chatty. But including it in the `JoinPlan` is doable. I can do it in this PR if you prefer.

src/sender_queue/mod.rs
vkomenda: An issue with making `D::Output: Epoched` is that `epoch` is not of type `<D::Output as Epoched>::Epoch` but of type `<D::Message as Epoched>::Epoch`! And those are not equal.

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+    ///
+    /// - the duration of `e0` is at least that of `e`.
+    ///
+    /// Returned is a list of spanning epochs for the given epoch.
+    ///
+    /// For example, any `DynamicHoneyBadger` epoch `Epoch((x, Some(y)))` has a unique spanning
+    /// epoch `Epoch((x, None))`. In turn, no epoch `Epoch((x, None))` has a spanning epoch.
+    fn spanning_epochs(
+        epoch: <D::Message as Epoched>::Epoch,
+    ) -> Vec<<D::Message as Epoched>::Epoch>;
```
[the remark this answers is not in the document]
vkomenda: `self` is used for type inference. However I'll try what you are suggesting.

src/sender_queue/mod.rs
vkomenda: Possibly, I'm not sure. Is your idea that by doing this we would help the joining nodes to catch up faster?

afck: No, I'm just thinking that we already know what they new node is going to tell us: We already know that the batch's next epoch is the one we will receive in its `EpochStarted` message?
If not, we could set it to the default epoch. I'm just thinking it would make it easier if we didn't have the separate `peer_id`s collection, and just always had every peer as a key in the epoch map?

src/traits.rs
```diff
@@ -163,6 +164,128 @@ impl<D: DistAlgorithm> From<TargetedMessage<D::Message, D::NodeId>> for Step<D>
...
+                        let lagging = |&them| {
+                            !(is_accepting_epoch(&message, them) || is_later_epoch(&message, them))
+                        };
+                        if remote_epochs.get(&id).map_or(true, lagging) {
+                            defer = true;
+                        }
+                    }
+                    if defer {
+                        deferred_msgs.push((id, message));
+                    }
```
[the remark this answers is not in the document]
vkomenda: I rewrote it a bit differently. Please check if that's something you think is OK too.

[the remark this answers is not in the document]
afck: Sure, that's fine with me, too. :+1:

src/sender_queue/mod.rs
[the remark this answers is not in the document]
afck: Four traits… :dizzy_face: 
I'm still hoping there's a simpler solution, but we can also refactor that in a separate PR, of course.

src/dynamic_honey_badger/mod.rs
```diff
@@ -112,19 +116,54 @@ pub enum Message<N: Rand> {
...
             Message::SignedVote(ref signed_vote) => signed_vote.era(),
         }
     }
+}
+
+/// Dynamic Honey Badger epoch. It consists of an era and an epoch of Honey Badger that started in
+/// that era. For messages originating from `DynamicHoneyBadger` as opposed to `HoneyBadger`, that
+/// HoneyBadger epoch is `None`.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, Serialize, Deserialize)]
+pub struct Epoch(pub(super) u64, pub(super) Option<u64>);
```
c0gent: Need an `Add` impl for this.

afck: I don't think so. From the user's perspective, batches will still have `u64` epoch numbers.

c0gent: Or, I guess it would be sufficient to just add a method which returns the epoch as a u64.

c0gent: I guess it would be sufficient to just add a method which returns the epoch as a u64.

afck: Somehow I feel like the user-visible `seqnum` should keep the name `epoch`…

src/honey_badger/batch.rs
```diff
@@ -9,6 +9,15 @@ pub struct Batch<C, N> {
     pub contributions: BTreeMap<N, C>,
 }
 
+impl<C, N: NodeIdT> Epoched for Batch<C, N> {
+    type Epoch = u64;
+
+    /// Returns the **next** `HoneyBadger` epoch after the sequential epoch of the batch.
+    fn epoch(&self) -> u64 {
+        self.epoch + 1
```
afck: I find that very counterintuitive. Should there maybe be a `next_epoch` method in `SenderQueueableOutput` instead?

src/sender_queue/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+
+pub trait SenderQueueableMessage: Epoched {
+    /// Whether the message is accepted in epoch `them`.
+    fn is_accepted(&self, them: <Self as Epoched>::Epoch, max_future_epochs: u64) -> bool;
+
+    /// Whether the epoch of the message is behind `them`.
+    fn is_obsolete(&self, them: <Self as Epoched>::Epoch) -> bool;
+}
+
+pub trait SenderQueueableOutput<N, M>: Epoched
```
afck: If we add `next_epoch`, we can probably remove the `Epoched` trait bound here.

c0gent:
Hydrabadger works fine with these changes fyi. `Batch::seqnum` definitely needs to be renamed.

src/dynamic_honey_badger/batch.rs
```diff
@@ -18,9 +20,30 @@ pub struct Batch<C, N> {
...
+            Epoch(seqnum + 1, Some(0))
+        }
+    }
+}
+
 impl<C, N: NodeIdT> Batch<C, N> {
-    pub fn epoch(&self) -> u64 {
-        self.epoch
+    /// Returns the linear epoch of this `DynamicHoneyBadger` batch.
+    pub fn seqnum(&self) -> u64 {
```
c0gent: Needs a better name.

src/traits.rs
```diff
@@ -183,6 +186,120 @@ impl<D: DistAlgorithm> From<TargetedMessage<D::Message, D::NodeId>> for Step<D>
...
+            messages
+                .drain(..)
+                .partition(|TargetedMessage { target, message }| match target {
+                    Target::All => peer_epochs
+                        .values()
+                        .all(|&them| message.is_accepted(them, max_future_epochs)),
+                    Target::Node(id) => peer_epochs
+                        .get(&id)
+                        .map_or(false, |&them| message.is_accepted(them, max_future_epochs)),
+                });
```
afck: I still feel it would be easier to just iterate through all the messages in the big loop below, and populate `passed_msgs` there, too. That would avoid allocating and then consuming `failed_msgs`.
(Can also be refactored in another PR, of course.)

src/traits.rs
```diff
@@ -163,6 +164,128 @@ impl<D: DistAlgorithm> From<TargetedMessage<D::Message, D::NodeId>> for Step<D>
...
+                    let accepting_nodes: BTreeSet<&D::NodeId> = remote_epochs
+                        .iter()
+                        .filter(|(_, them)| accepts(them))
+                        .map(|(id, _)| id)
+                        .collect();
+                    let non_lagging_nodes: BTreeSet<&D::NodeId> = remote_epochs
+                        .iter()
+                        .filter(|(_, them)| isnt_earlier_epoch(them))
+                        .map(|(id, _)| id)
+                        .collect();
```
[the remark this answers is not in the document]
afck: I still think this :arrow_up: would make the code simpler _and_ avoid allocating the two sets.

afck:
I won't be able to continue the review before Monday, but if @mbr and @c0gent approve, feel free to merge.

src/dynamic_honey_badger/batch.rs
vkomenda: I can rename the `seqnum` field and method back to `epoch` and the `epoch` method to `next_epoch`.

c0gent: Maybe adding a `_num` or `_no` suffix to the methods which return a `u64` as opposed to an `Epoch` would help clarify things? 

Perhaps we should just leave it as is for now and create an issue to figure out a consistent naming scheme in a future PR.

vkomenda: I also think we might need a more catchy name for this user-facing epoch of DHB. We already have two other, epochs that are used internally in HB and DHB. We are currently reusing the term "epoch" for the user-facing reduction of the real DHB epoch.

src/dynamic_honey_badger/mod.rs
vkomenda: `+` won't work on `Epoch`. If you need the next epoch of a batch, please see `SenderQueueableOutput::next_epoch`.

src/dynamic_honey_badger/sender_queueable.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn is_accepted(&self, Epoch(them_era, them_hb_epoch): Epoch, max_future_epochs: u64) -> bool {
+        let Epoch(era, hb_epoch) = self.epoch();
+        if era != them_era {
+            return false;
+        }
+        match (hb_epoch, them_hb_epoch) {
+            (Some(us), Some(them)) => them <= us && us <= them + max_future_epochs,
+            (None, Some(_)) => true,
+            (_, None) => {
+                error!("Peer's Honey Badger epoch undefined");
```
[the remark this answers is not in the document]
vkomenda: TODO added.

vkomenda:
If that is needed, I would rather add `max_future_epochs` to `JoinPlan` in a separate PR not to overcrowd this one.

src/dynamic_honey_badger/sender_queueable.rs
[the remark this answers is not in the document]
vkomenda: ...and I refactored the case out completely. No `Fault` required.

src/traits.rs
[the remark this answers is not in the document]
vkomenda: There should only be one extra negation in your `else if` clause.

vkomenda:
Rebased on top of master.

The last commit adds an intermediate type between non-linear epochs `Epoched::Epoch` and linear epochs of DHB, that is `u64`. This intermediate type is `Epoched::LinEpoch` of "linearizable epochs". It can be converted to `u64` but not vice versa. Also `peer_epochs` now use this type, which allows to avoid unreachable match cases.

src/traits.rs
```diff
@@ -237,59 +236,46 @@ where
...
+                passed_msgs.push(msg);
+            } else {
+                let m = msg.message;
+                match msg.target {
+                    Target::Node(ref id) => {
+                        let defer = {
+                            let lagging = |&them| {
+                                !(m.is_accepted(them, max_future_epochs) || m.is_obsolete(them))
+                            };
+                            peer_epochs.get(&id).map_or(true, lagging)
```
afck: Now that `peer_epochs` contains all peers as keys, we can probably drop the message if the peer isn't found?

And then we already know at this point that `!m.is_accepted(them, max_future_epochs)`. I think the code would look easier if `pass` was fully merged into this `match` statement here, instead of matching the same message twice. Then it would just say something like this here:

```rust
if let Some(them) = peer_epochs.get(&id) {
    if m.is_accepted(them, max_future_epochs) {
        passed_msgs.push(msg);
    } else if !m.is_obsolete(them) {
        deferred_msgs.push(msg);
    }
}
```

vkomenda: You wouldn't be able to push `msg` without cloning since in that case it is borrowed by the match statement.

vkomenda: But we can instead clone the target of the message only, which is cheaper compared to cloning the whole message, even though it will be done on every outgoing message.

PR #310, opened 2018-10-30
Improve the broadcast documentation.

afck:
* Reorganize the explanation, separate algorithm from justification.
* Some of the message numbers were wrong.
* Use `[i]` for indices, even in pseudocode.

### DIFF
--- a/src/broadcast/mod.rs
+++ b/src/broadcast/mod.rs
@@ -11,35 +11,47 @@
 //!
 //! ## How it works
 //!
-//! * The proposer uses a Reed-Solomon code to split the value into _N_ chunks, _N - 2 f_ of which
+//! The proposer uses a Reed-Solomon code to split the value into _N_ chunks, _N - 2 f_ of which
 //! suffice to reconstruct the value. These chunks are put into a Merkle tree, so that with the
-//! tree's root hash `h`, branch `bi` and chunk `si`, the `i`-th chunk `si` can be verified by
-//! anyone as belonging to the Merkle tree with root hash `h`. These values are "proof" number `i`:
-//! `pi = (h, bi, si)`.
-//! * The proposer sends `Value(pi)` to node `i`. It translates to: "I am the proposer, and `pi`
-//! contains the `i`-th share of my value."
-//! * Every (correct) node that receives `Value(pi)` from the proposer sends it on to everyone else
-//! as `Echo(pi)`. An `Echo` translates to: "I have received `pi` directly from the proposer." If
-//! the proposer sends another `Value` message it is ignored.
-//! * So every node that receives at least _f + 1_ `Echo` messages with the same root hash can
-//! decode a value.
-//! * Every node that has received _N - f_ `Echo`s with the same root hash from different nodes
-//! knows that at least _N - 2 f_ _correct_ nodes have sent an `Echo` with that hash to everyone,
-//! and therefore everyone will eventually receive at least _N - f_ of them. So upon receiving
-//! _N - f_ `Echo`s, they send a `Ready(h)` to everyone. It translates to: "I know that everyone
-//! will eventually be able to decode the value with root hash `h`." Moreover, since every correct
-//! node only sends one kind of `Echo` message, there is no danger of receiving _N - f_ `Echo`s
-//! with two different root hashes.
-//! * Even without enough `Echo` messages, if a node receives _2 f + 1_ `Ready` messages, it knows
-//! that at least one _correct_ node has sent `Ready`. It therefore also knows that everyone will
-//! be able to decode eventually, and multicasts `Ready` itself.
-//! * If a node has received _2 f + 1_ `Ready`s (with matching root hash) from different nodes,
-//! it knows that at least _2 f + 1_ _correct_ nodes have sent it. Therefore, every correct node
-//! will eventually receive _2 f + 1_, and multicast it itself. Therefore, every correct node will
-//! eventually receive _2 f + 1_ `Ready`s, too. _And_ we know at this point that every correct
-//! node will eventually be able to decode (i.e. receive at least _2 f + 1_ `Echo` messages).
-//! * So a node with _2 f + 1_ `Ready`s and _N - 2 f_ `Echos` will decode and _output_ the value,
-//! knowing that every other correct node will eventually do the same.
+//! branch `b[i]`, the `i`-th chunk `s[i]` can be verified by anyone as belonging to the Merkle
+//! tree with root hash `h`. These values are "proof" number `i`: `p[i] = (h, b[i], s[i])`.
+//!
+//! * The proposer sends `Value(p[i])` to each node `i`.<br>
+//! It translates to:
+//! "I am the proposer, and `p[i]` contains the `i`-th share of my value."
+//! * A node that receives `Value(p[i])` from the proposer sends it on to everyone
+//! else as `Echo(p[i])`. (Additional `Value`s are ignored.)<br>
+//! It translates to:
+//! "I have received `p[i]` from the proposer."
+//! * A node that has received _N - f_ `Echo`s **or** _f + 1_ `Ready`s with root hash `h`, sends
+//! `Ready(h)` to everyone.<br>
+//! It translates to:
+//! "I know that everyone will eventually be able to decode the value with root hash `h`."
+//! * A node that has received _2 f + 1_ `Ready`s **and** _N - 2 f_ `Echo`s with root hash `h`
+//! decodes and outputs the value, and then terminates.
+//!
+//! The third point is justified as follows:
+//!
+//! A node that receives at least _N - 2 f_ `Echo`s
+//! with root hash `h` can decode the value. So every node that has received _N - f_ `Echo`s with
+//! `h` knows that at least _N - 2 f_ **correct** nodes have sent an `Echo` with `h` to everyone,
+//! and therefore everyone will eventually receive at least _N - 2 f_ of them. So upon receiving
+//! _N - f_ `Echo`s, they know that everyone will be able to decode, and send `Ready(h)`.
+//! Moreover, since every correct node only sends one kind of `Echo` message, there is no danger of
+//! receiving _N - f_ `Echo`s with two different root hashes.
+//!
+//! Even without enough `Echo`s, if a node receives _f + 1_ `Ready(h)` messages, it knows
+//! that at least one _correct_ node has sent `Ready(h)`. It therefore also knows that everyone will
+//! be able to decode eventually, and multicasts `Ready(h)` itself.
+//!
+//! Regarding the last point:
+//!
+//! If a node has received _2 f + 1_ `Ready(h)` messages, it knows that at least _f + 1_
+//! **correct** nodes have sent it. Therefore, every node will eventually receive _f + 1_, and
+//! multicast `Ready(h)` itself. Therefore, every node will eventually receive _2 f + 1_ `Ready(h)`
+//! messages, too. In addition, we know at this point that every correct will eventually be
+//! able to decode, i.e. receive _N - 2 f_ `Echo` (since we know that at least one correct node has
+//! sent `Ready(h)`). So we know that all the other correct nodes will also output the value.
 //!
 //! ## Example
 //!
--- a/src/broadcast/mod.rs
+++ b/src/broadcast/mod.rs
@@ -1,46 +1,63 @@
 //! # Broadcast
 //!
-//! The Broadcast Protocol assumes a network of _N_ nodes that send signed messages to
-//! each other, with at most _f_ of them faulty, where _3 f < N_. Handling the networking and
-//! signing is the responsibility of this crate's user; a message is only handed to the Broadcast
-//! instance after it has been verified to be "from node i". One of the nodes is the "proposer"
-//! who sends a value. It needs to be determined beforehand, and all nodes need to know and agree
-//! who it is. Under the above conditions, the protocol guarantees that either all or none
-//! of the correct nodes output a value, and that if the proposer is correct, all correct nodes
-//! output the proposed value.
+//! The Broadcast Protocol assumes a network of _N_ validators that send signed messages to
+//! each other, with at most _f_ of them faulty, where _3 f < N_. It allows one validator, the
+//! "proposer", to send a value to the other validators, and guarantees that:
+//! * If the proposer is correct, all correct validators will receive the value.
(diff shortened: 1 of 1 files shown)


src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
-//! it knows that at least _2 f + 1_ _correct_ nodes have sent it. Therefore, every correct node
-//! will eventually receive _2 f + 1_, and multicast it itself. Therefore, every correct node will
-//! eventually receive _2 f + 1_ `Ready`s, too. _And_ we know at this point that every correct
-//! node will eventually be able to decode (i.e. receive at least _2 f + 1_ `Echo` messages).
-//! * So a node with _2 f + 1_ `Ready`s and _N - 2 f_ `Echos` will decode and _output_ the value,
-//! knowing that every other correct node will eventually do the same.
+//! branch `b[i]`, the `i`-th chunk `s[i]` can be verified by anyone as belonging to the Merkle
+//! tree with root hash `h`. These values are "proof" number `i`: `p[i] = (h, b[i], s[i])`.
+//!
+//! * The proposer sends `Value(p[i])` to node `i`.<br>
```
mbr: Shouldn't this be "each node `i`"?

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! It translates to:
+//! "I am the proposer, and `p[i]` contains the `i`-th share of my value."
```
mbr: To me, these "colloquial" phrasings are not helping much.

Would something like this be easier to understand?

> This way, the proposer announces itself and distributes shares of the value equally among all nodes, along with a way for each node to verify that all shares are from the value.

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! * A node that receives `Value(p[i])` from the proposer sends it on to everyone
+//! else as `Echo(p[i])`. (Additional `Value`s are ignored.)<br>
+//! It translates to:
+//! "I have received `p[i]` from the proposer."
```
mbr: Alternative:

> An `Echo(p[i])` from a node indicates that it has received a share of the value from the proposer. Since they contain part of the value, they are also used later on to reconstruct it after the algorithm completes.

src/broadcast/mod.rs
afck: I amended it.

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! * A node that has received _N - f_ `Echo`s **or** _f + 1_ `Ready`s with root hash `h`, sends
+//! `Ready(h)` to everyone.<br>
+//! It translates to:
+//! "I know that everyone will eventually be able to decode the value with root hash `h`."
```
mbr: > If a node has received `N-f` parts of the message, the `Ready(h)` is justified because due to the Reed-Solomon-Code this is the minimal amount of shares required to reconstruct it. The accompanying proof let's the node verify that all shares are from the same value.
>
> Sending a `Ready` after receiving `f+1` `Readys` is more speculative; since there can be at most `f` byzantine nodes "faking" a `Ready`, the value may not be complete yet, but the node is sure it will receive it eventually.

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! * A node that has received _2 f + 1_ `Ready`s **and** _N - 2 f_ `Echo`s with root hash `h`
+//! decodes and outputs the value, and then terminates.
+//!
+//! The third point is justified as follows:
```
mbr: This is a bit lengthy; you decide which version to kep ;).

src/broadcast/mod.rs
afck: Hmm… I guess both the "I am the proposer" and "announces itself" aren't technically correct, since the algorithm is already instantiated with that information, i.e. the proposer is known beforehand. So maybe:

> This way, the proposer distributes shares of the value equally among all nodes, along with a way to verify that all shares are from the same value.

src/broadcast/mod.rs
afck: I don't quite agree with that. I really think we send a `Ready` if and only if we know that everyone will eventually be able to decode (always under the assumption that ≤ _f_ are faulty, of course):
* _N - f_ parts are not the minimal amount to reconstruct: that's _N - 2 f_. Having received _N - f_ `Echo`s implies that _N - 2 f_ **correct** nodes have sent it, so **every node** will receive at least _N - 2 f_, so every node will be able to decode.
* _f + 1_ `Ready`s imply that at least one correct node has received _N - f_ `Echo`s, which is as good as if we had received them ourselves: at this point, we also know that everyone will eventually be able to decode (including ourselves). So I don't consider this speculative at all.

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
-//! it knows that at least _2 f + 1_ _correct_ nodes have sent it. Therefore, every correct node
-//! will eventually receive _2 f + 1_, and multicast it itself. Therefore, every correct node will
-//! eventually receive _2 f + 1_ `Ready`s, too. _And_ we know at this point that every correct
-//! node will eventually be able to decode (i.e. receive at least _2 f + 1_ `Echo` messages).
-//! * So a node with _2 f + 1_ `Ready`s and _N - 2 f_ `Echos` will decode and _output_ the value,
-//! knowing that every other correct node will eventually do the same.
+//! branch `b[i]`, the `i`-th chunk `s[i]` can be verified by anyone as belonging to the Merkle
+//! tree with root hash `h`. These values are "proof" number `i`: `p[i] = (h, b[i], s[i])`.
+//!
+//! * The proposer sends `Value(p[i])` to each node `i`.<br>
```
vkomenda: "node" -> "validator"? Could more precise terminology be still applicable here? I know, `Values` are only sent between validators anyway. However, `Echo`s are also sent to observers. So there is a clear difference which could be emphasised in the docs.

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! It translates to:
+//! "I am the proposer, and `p[i]` contains the `i`-th share of my value."
+//! * A node that receives `Value(p[i])` from the proposer sends it on to everyone
```
vkomenda: "A node `i` that receives a valid `Value(p[i])`"...

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! * A node that has received _N - f_ `Echo`s **or** _f + 1_ `Ready`s with root hash `h`, sends
+//! `Ready(h)` to everyone.<br>
+//! It translates to:
+//! "I know that everyone will eventually be able to decode the value with root hash `h`."
+//! * A node that has received _2 f + 1_ `Ready`s **and** _N - 2 f_ `Echo`s with root hash `h`
+//! decodes and outputs the value, and then terminates.
+//!
+//! The third point is justified as follows:
+//!
+//! A node that receives at least _N - 2 f_ `Echo`s
```
vkomenda: `Echo`s -> valid `Echo`s

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! with root hash `h` can decode the value. So every node that has received _N - f_ `Echo`s with
```
vkomenda: `Echo`s -> valid `Echo`s

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! `h` knows that at least _N - 2 f_ **correct** nodes have sent an `Echo` with `h` to everyone,
+//! and therefore everyone will eventually receive at least _N - 2 f_ of them. So upon receiving
+//! _N - f_ `Echo`s, they know that everyone will be able to decode, and send `Ready(h)`.
```
vkomenda: also "valid `Echo`s"

src/broadcast/mod.rs
```diff
@@ -11,35 +11,47 @@
...
+//! that at least one _correct_ node has sent `Ready(h)`. It therefore also knows that everyone will
+//! be able to decode eventually, and multicasts `Ready(h)` itself.
+//!
+//! Regarding the last point:
+//!
+//! If a node has received _2 f + 1_ `Ready(h)` messages, it knows that at least _f + 1_
+//! **correct** nodes have sent it. Therefore, every node will eventually receive _f + 1_, and
+//! multicast `Ready(h)` itself. Therefore, every node will eventually receive _2 f + 1_ `Ready(h)`
+//! messages, too. In addition, we know at this point that every correct will eventually be
+//! able to decode, i.e. receive _N - 2 f_ `Echo` (since we know that at least one correct node has
```
vkomenda: "valid `Echo`s"

src/broadcast/mod.rs
afck: I added a paragraph about validity, to avoid inserting it everywhere. (Also, due to the note above, I think it only applies right at the end, where we construct the value—that's where we have to wait for _N - 2 f_ **valid** messages. In all the other places we could count the invalid ones, too.)

src/broadcast/mod.rs
```diff
@@ -11,35 +11,54 @@
...
+//! * The proposer sends `Value(p[i])` to each validator `i`.
+//! * A validator `i` that receives `Value(p[i])` from the proposer sends it on to everyone else as
+//! `Echo(p[i])`.
+//! * A validator that has received _N - f_ `Echo`s **or** _f + 1_ `Ready`s with root hash `h`,
+//! sends `Ready(h)` to everyone.
+//! * A node that has received _2 f + 1_ `Ready`s **and** _N - 2 f_ `Echo`s with root hash `h`
+//! decodes and outputs the value, and then terminates.
+//!
+//! Only the first valid `Value` from the proposer, and the first valid `Echo` message from every
+//! validator is handled as above. Invalid messages (where the proof isn't correct), `Values`
```
andogro: validator, is handled as above (add comma).

src/broadcast/mod.rs
```diff
@@ -11,35 +11,54 @@
...
+//! received from other nodes, and all further `Value`s and `Echo`s are ignored, and the sender is
+//! reported as faulty.
+//!
+//! In the `Valid(p[i])` messages, the proposer distributes shares of the value equally among all
```
andogro: In a valid `Value(p[i]` message?  or did you mean to just say `Value(p[i])'

src/broadcast/mod.rs
```diff
@@ -11,35 +11,54 @@
...
+//! everyone will be able to decode eventually, and multicasts `Ready(h)` itself.
+//!
+//! Finally, if a node has received _2 f + 1_ `Ready(h)` messages, it knows that at least _f + 1_
+//! **correct** validators have sent it. Thus, every remaining correct validator will eventually
+//! receive _f + 1_, and multicast `Ready(h)` itself. Hence every node will receive _2 f + 1_
+//! `Ready(h)` messages.<br>
+//! In addition, we know at this point that every node will eventually be able to decode, i.e.
+//! receive _N - 2 f_ valid `Echo`s (since we know that at least one correct validator has sent
+//! `Ready(h)`).<br>
+//! In short: Once we satisfy the termination condition in the fourth point above, we know that
```
andogro: Once we satisfy the termination condition in the fourth point above ( we've received _2 f + 1_ `Ready`s **and** _N - 2 f_ `Echo`s with root hash `h`),  

added for convenience, it takes a little hunting and scrolling to see which is the 4th point since they are not numbered.

src/broadcast/mod.rs
```diff
@@ -11,35 +11,54 @@
...
-//! it knows that at least _2 f + 1_ _correct_ nodes have sent it. Therefore, every correct node
-//! will eventually receive _2 f + 1_, and multicast it itself. Therefore, every correct node will
-//! eventually receive _2 f + 1_ `Ready`s, too. _And_ we know at this point that every correct
-//! node will eventually be able to decode (i.e. receive at least _2 f + 1_ `Echo` messages).
-//! * So a node with _2 f + 1_ `Ready`s and _N - 2 f_ `Echos` will decode and _output_ the value,
-//! knowing that every other correct node will eventually do the same.
+//! branch `b[i]`, the `i`-th chunk `s[i]` can be verified by anyone as belonging to the Merkle
+//! tree with root hash `h`. These values are "proof" number `i`: `p[i] = (h, b[i], s[i])`.
+//!
+//! * The proposer sends `Value(p[i])` to each validator `i`.
```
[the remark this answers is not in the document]
afck: If I use the same `i` in two places in any single bullet point, it does mean the same number: The proposer sends chunk number 5, inside `Value(p[5])`, to validator number 5. Validator number 5 then sends `Echo(p[5])` to everyone else.

I agree that the "i" in the introduction above was a bit out of place; it tried to reformulate that, too, and elaborate on the Merkle tree.

src/broadcast/mod.rs
afck: I was meaning to say that, by sending `Value(p[0])` to validator 0, `Value(p[1])` to validator 1, etc., the proposer distributes the proofs (containing the chunks) equally among the validators.

src/broadcast/mod.rs
mbr: I see now that I made a mistake there, as you pointed out, the Merkle tree lets us pinpoint which shares are incorrect, reducing the number of necessary fragments to reconstruct from *N - f* to *N - 2f*. I am not sure that that'll hold for any erasure code though; it might be a property of RS.

The second paragraph is poorly phrased on my part, the "not complete" in this case is meant to apply to just the current node, not the network itself.

I retract my objections, but I maintain we should improve this section =).

src/broadcast/mod.rs
mbr: There are no "validators" in a broadcast, right? Since this is lower level, should we not keep the terminology lower level as well?

afck: Yes, there are: Validators are the ones who receive `Value`s and send `Echo`s and `Ready`s.
An observer—someone who only receives `Echo`s and `Ready`s—can still receive the value, with the same guarantees.

But I think we should explain the general concept in the crate documentation rather than here. (_If_ we keep it.)

src/broadcast/mod.rs
afck: If I understand correctly, we could use any erasure code with a code rate of _(N - 2 f) / N_ here.
(Note that I already rewrote most of this section. Please take a look at the new version.)

andogro:
Looks good, including the basic methods and Merkle tree explanation makes it much clearer.

Issue #286, opened 2018-10-23
Consider making ThresholdDecryption work like ThresholdSign.


afck:
`ThresholdSign` receives the to-be-signed data on construction. Its _input_ is `()`: It's only there to trigger sending the shares. This has the advantage that we can validate incoming signature shares even if `handle_input` wasn't called yet.

`ThresholdDecryption` receives the ciphertext via `handle_input`. This makes its implementation more complicated, but maybe it's more intuitive to use?

This feels inconsistent. We should decide on one approach. I'm leaning towards the former, depending on how much that complicates `ThresholdDecryption`'s use in `EpochState`.

vkomenda:
In the latter approach, did you mean instead that `ThresholdSign` would receive a ciptertext through `handle_input`?

`ThresholdDecryption` currently already handles a ciphertext input.

afck:
Yes, exactly: the second one is how `ThresholdDecryption` currently works; for `ThresholdSign` it would mean inputting the to-be-signed data in `handle_input` instead of the constructor.

afck:
As @alyjak pointed out in https://github.com/poanetwork/hbbft/pull/312#discussion_r229513749, making TD work like TS would move quite a lot of complexity from (simple) TD to (already complex) `EpochState`.

It's probably a good thing that TD currently handles all cases automatically: decryption shares arriving before or after the ciphertext. So maybe it should be the other way round, and TS should do the same, i.e. not accept the data to sign in the constructor, but in the single `handle_input` call that triggers sending the share instead?

Another, even more flexible (but maybe unnecessarily complicated?) option would be to change both:
* No data/ciphertext in the constructor.
* One `set_data`/`set_ciphertext` call to set the data/ciphertext. (At this point, all previously received invalid shares can be discarded, and fault logs generated.)
* One separate `send_share` call to trigger sending our own share to the others.

…because in `EpochState`, we can't send the decryption share at the moment where we receive the ciphertext, and it would be convenient to store it directly in the `ThresholdDecryption` instance right away.

afck:
Implemented in #316.

PR #312, opened 2018-10-30
Issue #286: make ThresholdDecryption consistent with ThresholdSign

alyjak:
Turns threshold_decryption messages into an enum that can pass `DecryptionShare`s or `Ciphertext`. This allows  `ThresholdDecryption` instances to be asynchronously populated through the common handle_message interface. `ThresholdDecryption`'s `handle_input` function can now simply be used to see if output is available for the instance.

### DIFF
--- a/src/honey_badger/epoch_state.rs
+++ b/src/honey_badger/epoch_state.rs
@@ -46,14 +46,6 @@ where
             DecryptionState::Complete(_) => Ok(td::Step::default()),
         }
     }
-
-    /// Handles a ciphertext input.
-    fn set_ciphertext(&mut self, ciphertext: Ciphertext) -> td::Result<td::Step<N>> {
-        match self {
-            DecryptionState::Ongoing(ref mut td) => td.handle_input(ciphertext),
-            DecryptionState::Complete(_) => Ok(td::Step::default()),
-        }
-    }
 }
 
 /// The status of the subset algorithm.
@@ -254,7 +246,7 @@ where
                     Entry::Vacant(entry) => {
                         entry.insert(DecryptionState::new(self.netinfo.clone()))
                     }
-                }.handle_message(sender_id, share)
+                }.handle_message(sender_id, td::Message::Share(share))
                 .map_err(ErrorKind::ThresholdDecryption)?;
                 self.process_decryption(proposer_id, td_step)
             }
@@ -344,11 +336,13 @@ where
     /// Processes a Threshold Decryption step.
     fn process_decryption(&mut self, proposer_id: N, td_step: td::Step<N>) -> Result<Step<C, N>> {
         let mut step = Step::default();
-        let opt_output = step.extend_with(td_step, |share| {
-            MessageContent::DecryptionShare {
-                proposer_id: proposer_id.clone(),
-                share,
-            }.with_epoch(self.epoch)
+        let opt_output = step.extend_with(td_step, |msg_share| {
+            match msg_share {
+                td::Message::Share(share) => {
+                    MessageContent::DecryptionShare{proposer_id: proposer_id.clone(),
+                                                    share: share}.with_epoch(self.epoch)},
+                _ => panic!(),
+            }
         });
         if let Some(output) = opt_output.into_iter().next() {
             self.decryption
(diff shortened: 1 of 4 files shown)


src/honey_badger/epoch_state.rs
```diff
@@ -344,11 +336,13 @@ where
...
-            MessageContent::DecryptionShare {
-                proposer_id: proposer_id.clone(),
-                share,
-            }.with_epoch(self.epoch)
+        let opt_output = step.extend_with(td_step, |msg_share| {
+            match msg_share {
+                td::Message::Share(share) => {
+                    MessageContent::DecryptionShare{proposer_id: proposer_id.clone(),
+                                                    share: share}.with_epoch(self.epoch)},
+                _ => panic!(),
```
alyjak: I don't think this panic is the right thing to do, but my rust-fu and hbbt-fu aren't strong enough to know the proper way to handle this. Is there something like unwrap for enums?

src/threshold_decryption.rs
```diff
@@ -128,18 +133,25 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
...
+                }
+                if self.shares.insert(sender_id.clone(), share).is_some() {
+                    return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleDecryptionShares).into());
+                }
+                Step::default()
+            },
+            Message::Cipher(ciphertext) => self.set_ciphertext(ciphertext)?,
+        });
+        step.extend(self.try_output()?);
+        Ok(step)
```
alyjak: I'm not sure if this is correct -- should I be extending the step like this?

src/threshold_decryption.rs
```diff
@@ -38,8 +38,13 @@ pub enum Error {
...
 
 /// A Threshold Decryption message.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Rand)]
-pub struct Message(pub DecryptionShare);
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum Message {
+    /// A decryption share from a peer.
+    Share(DecryptionShare),
+    /// A message setting nodeId's ciphertext Should only be transmitted locally.
+    Cipher(Ciphertext),
```
alyjak: I might be hacking the handle_message api here. The ciphertext should only be passed internally and not  become network traffic. I think this is enforced through the honeybadger message type -- which intentionally has no definition for a Ciphertext message content, but let me know if this is sketchy or not.

afck: Yes, the `Message ` should stay the same.
Instead, the ciphertext should be passed into the constructor of `ThresholdDecryption`.

src/threshold_decryption.rs
```diff
@@ -58,13 +63,13 @@ pub type Step<N> = ::Step<ThresholdDecryption<N>>;
...
-    type Input = Ciphertext;
+    type Input = ();
     type Output = Vec<u8>;
     type Message = Message;
     type Error = Error;
 
-    fn handle_input(&mut self, input: Ciphertext) -> Result<Step<N>> {
-        self.set_ciphertext(input)
+    fn handle_input(&mut self, _input: ()) -> Result<Step<N>> {
+        self.try_output()
```
afck: We still want to trigger the sending of our own threshold share here.

src/threshold_decryption.rs
afck: In general, extending `step` this way is fine, but we shoudln't introduce `Message::Cipher`, so we won't need this change here.

src/threshold_decryption.rs
alyjak: Does that work? Isn't it possible for threshold shares to arrive prior to the ciphertext being sent from the broadcast?

afck: That's true, it would move the complexity of storing shares before knowing the ciphertext from `threshold_decryption` to `honey_badger::epoch_state`, so maybe it's a bad idea.

#286 is very much a design issue, too: We need to find the best approach.

But we definitely don't want to send the ciphertext across the network again (that's what has already been done at this point, using the `Broadcast`/RBC algorithm): `Message` won't need to change.

alyjak: By validated do you mean something other than attempting to decrypt after F+1 incoming shares are received?

If not, incoming shares can't be validated without the ciphertext, and a node's share can't be sent without the ciphertext right? So, since shares and ciphertext arrive asynchronously I think the current design is optimal already, as the node's share is sent as soon as the ciphertext arrives plus `try_output` is called as well.

afck: > something other than attempting to decrypt

Yes, like the current code already does:
https://github.com/poanetwork/hbbft/blob/2fed8316890f6907d8155d815dd614fdca6978d1/src/threshold_decryption.rs#L169

Exactly, that's only possible with the ciphertext: So currently we validate as soon as we have the ciphertext, and we decrypt as soon as we have ciphertext and _f + 1_ shares.

> the current design is optimal already

Yes, it's fine, given its intended usage inside hbbft.
I just opened #286 because I find it a bit inconsistent how the API of `ThresholdDecryption` differs from the `ThresholdSign` one. I'm not sure about it, but I feel like it would make sense to bring the two in line.

alyjak:
This approach was rejected, closing the pull request. See #316 for the folloup.

PR #314, opened 2018-10-31
Remove some obsolete TODOs and comments.

afck:
Specifically, we don't have plans at the moment to move serialization into `Subset` and `Broadcast`: It doesn't make sense to force serde on the user, and often there's e.g. additional encryption involved. Our algorithms shouldn't care about that, and just take a `Vec<u8>`.

### DIFF
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -20,7 +20,7 @@ pub struct Broadcast<N> {
     netinfo: Arc<NetworkInfo<N>>,
     /// The ID of the sending node.
     proposer_id: N,
-    data_shard_num: usize,
+    /// The Reed-Solomon erasure coding configuration.
     coding: Coding,
     /// If we are the proposer: whether we have already sent the `Value` messages with the shards.
     value_sent: bool,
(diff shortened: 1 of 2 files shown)


src/broadcast/broadcast.rs
```diff
@@ -462,15 +454,13 @@ fn decode_from_shards(
...
         // Reconstruct the value from the data shards.
-        glue_shards(mtree, data_shard_num)
+        glue_shards(mtree, coding.data_shard_count())
     }
 }
 
-/// Concatenates the first `n` leaf values of a Merkle tree `m` in one value of
-/// type `T`. This is useful for reconstructing the data value held in the tree
-/// and forgetting the leaves that contain parity information.
+/// Concatenates the first `n` leaf values of a Merkle tree. The first four bytes are interpreted
```
vkomenda: "a Merkle tree" without the name may sound like an arbitrary Merkle tree.

afck: You mean it should be "the Merkle tree"

afck: I amended:

```diff
-/// Concatenates the first `n` leaf values of a Merkle tree. The first four bytes are interpreted
-/// as the payload size, and the padding beyond that size is dropped.
+/// Concatenates the first `n` leaf values of the Merkle tree `m`. The first four bytes are
+/// interpreted as the payload size, and the padding beyond that size is dropped.
```

vkomenda: Sorry, I think it should be an [indefinite description](https://plato.stanford.edu/entries/descriptions/): "a Merkle tree `m`". It might be a definite description ("the 'm'") but I'm lazy to follow the dictionary article argument. :books: :books: :books:

afck: OK, I amended it, too.

PR #316, opened 2018-11-01
issue286: make `ThresholdSign` and `ThresholdDecryption` APIs consistent

alyjak:
This gives both methods a `set_message(payload)` and a `sign()` method. If `sign` is called prior to
`set_message` or if was already called, then an empty step is returned. Otherwise share messages are
sent out to peers and `try_output` is called.


alyjak:
@afck, I've attempted to incorporate your feedback in order to resolve #286. This development branch makes `ThresholdSign` and `ThresholdDecription` consistent in terms of API. Although its not strictly necessary in terms of how its used in hbbt right now, a consequence of this is that `ThresholdSign` can now handle asynchronous input of both message shares and the message to sign.

I couldn't get `handle_input` to work to pass a `ThresholdSign` message. Note that I was working off an older version of the code and trying to get the Nonce to work. Maybe with the update to use a `bincode::serialize()` value it would work better. Is it a problem that I resorted to calling an internal function (`send_message`) rather than the `DistAlgorithm` function?

src/binary_agreement/binary_agreement.rs
```diff
@@ -324,7 +324,9 @@ impl<N: NodeIdT, S: SessionIdT> BinaryAgreement<N, S> {
             1 => CoinState::Decided(false),
             _ => {
                 let coin_id = bincode::serialize(&(&self.session_id, self.epoch))?;
-                CoinState::InProgress(Box::new(ThresholdSign::new(self.netinfo.clone(), coin_id)))
+                let mut ts = ThresholdSign::new(self.netinfo.clone());
+                ts.set_message(coin_id).unwrap();
```
afck: Instead of unwrapping, let's return an error here (probably using `map_err` and `?`).

src/threshold_decryption.rs
```diff
@@ -115,22 +115,32 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
     /// Sets the ciphertext, sends the decryption share, and tries to decrypt it.
     /// This must be called exactly once, with the same ciphertext in all participating nodes.
     /// If we have enough shares, outputs the plaintext.
-    pub fn set_ciphertext(&mut self, ct: Ciphertext) -> Result<Step<N>> {
+    pub fn set_message(&mut self, ct: Ciphertext) -> Result<()> {
```
afck: I'd keep the name `set_ciphertext`.

src/threshold_decryption.rs
```diff
@@ -115,22 +115,32 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
...
         if self.ciphertext.is_some() {
             return Err(Error::MultipleInputs(Box::new(ct)));
         }
+        self.ciphertext = Some(ct);
+        Ok(())
+    }
+
+    /// Sends our signature shares, and if we have collected enough, returns the decrypted message.
+    /// Returns an empty step if the ciphertext hasn't been received yet.
+    pub fn sign(&mut self) -> Result<Step<N>> {
```
afck: Maybe `decrypt`? (Or `send_shares`? Or `start_decryption`?)
(The doc comment also needs to be updated.)

src/threshold_decryption.rs
```diff
@@ -115,22 +115,32 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
...
+        let ct = match self.ciphertext {
+            None => return Ok(Step::default()),
```
afck: I think we should return an error here: The user is not supposed to call this method before `set_ciphertext`.

src/threshold_decryption.rs
```diff
@@ -115,22 +115,32 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
...
+        step.fault_log.extend(self.remove_invalid_shares());
         if !self.netinfo.is_validator() {
-            self.ciphertext = Some(ct);
-            return Ok(self.try_output()?);
+            step.extend(self.try_output()?);
+            return Ok(step);
         }
         let share = match self.netinfo.secret_key_share().decrypt_share(&ct) {
-            None => return Err(Error::InvalidCiphertext(Box::new(ct))),
+            None => return Err(Error::InvalidCiphertext(Box::new(ct.clone()))),
```
afck: I'd actually [`verify`](https://docs.rs/threshold_crypto/0.2.1/threshold_crypto/struct.Ciphertext.html#method.verify) this in `set_ciphertext` and use [`decrypt_share_no_verify`](https://docs.rs/threshold_crypto/0.2.1/threshold_crypto/struct.SecretKeyShare.html#method.decrypt_share_no_verify) here.

src/threshold_sign.rs
```diff
@@ -99,26 +101,42 @@ impl<N: NodeIdT> DistAlgorithm for ThresholdSign<N> {
...
+        }
+        self.msg_hash = Some(hash_g2(msg));
+        Ok(())
+    }
+
     /// Sends our signature shares, and if we have collected enough, returns the full signature.
+    /// Returns an empty step if the message to sign hasn't been received yet.
     pub fn sign(&mut self) -> Result<Step<N>> {
-        if self.had_input {
+        if self.had_input || self.msg_hash.is_none() {
```
afck: Calling `sign` before `set_message` should be an error.

src/threshold_sign.rs
```diff
@@ -99,26 +101,42 @@ impl<N: NodeIdT> DistAlgorithm for ThresholdSign<N> {
...
+        let mut step = Step::default();
+        step.fault_log.extend(self.remove_invalid_shares());
         if !self.netinfo.is_validator() {
             return self.try_output();
         }
-        let msg = Message(self.netinfo.secret_key_share().sign_g2(self.msg_hash));
+        let msg = Message(
+            self.netinfo
+                .secret_key_share()
+                .sign_g2(self.msg_hash.unwrap()),
```
afck: Let's take the hash above, at the same point where we're returning an error if it's `None`, to avoid the `unwrap`.

src/threshold_sign.rs
```diff
@@ -99,26 +101,42 @@ impl<N: NodeIdT> DistAlgorithm for ThresholdSign<N> {
...
-            msg_hash: hash_g2(msg),
+            msg_hash: None,
             received_shares: BTreeMap::new(),
             had_input: false,
             terminated: false,
         }
     }
 
+    /// Sets msg_hash and resets internal state accordingly.
+    pub fn set_message<M: AsRef<[u8]>>(&mut self, msg: M) -> Result<()> {
```
afck: Now that I think of it, "message" was a bad choice, since it's easy to confuse with the network messages being handled in `handle_message`. My bad! :grimacing: 
Maybe `set_document`?

src/threshold_sign.rs
```diff
@@ -135,19 +153,43 @@ impl<N: NodeIdT> ThresholdSign<N> {
...
-        if let Some(pk_i) = self.netinfo.public_key_share(sender_id) {
-            if !pk_i.verify_g2(&share, self.msg_hash) {
-                // Log the faulty node and ignore the invalid share.
-                let fault_kind = FaultKind::UnverifiedSignatureShareSender;
-                return Ok(Fault::new(sender_id.clone(), fault_kind).into());
-            }
-            self.received_shares.insert(sender_id.clone(), share);
+        if !self.is_share_valid(sender_id, &share) {
+            let fault_kind = FaultKind::UnverifiedSignatureShareSender;
+            return Ok(Fault::new(sender_id.clone(), fault_kind).into());
```
afck: I think we should keep the old behavior here: If the sender is unknown, return an error.
Actually, we should change `ThresholdDecryption` to do the same.

src/threshold_sign.rs
```diff
@@ -179,7 +221,7 @@ impl<N: NodeIdT> ThresholdSign<N> {
             .netinfo
             .public_key_set()
             .public_key()
-            .verify_g2(&sig, self.msg_hash)
+            .verify_g2(&sig, self.msg_hash.unwrap())
```
afck: I'd like to avoid `unwrap` calls in production code: We could pass in the hash from `try_output`, which should check it anyway and return an empty step if `None`.

tests/binary_agreement_mitm.rs
```diff
@@ -270,7 +270,9 @@ impl AbaCommonCoinAdversary {
                         .expect("Adversary netinfo mutex not populated");
                     let coin_id = bincode::serialize(&(0 as SessionId, epoch))
                         .expect("Failed to serialize coin_id");
-                    let mut coin = ThresholdSign::new(netinfo, coin_id);
+                    let mut coin = ThresholdSign::new(netinfo);
+                    coin.set_message(coin_id)
+                        .expect("Failed to set coin's message");
```
afck: I'd say "`the coin's ID`" here (even if we rename it to `set_document`).

tests/threshold_sign.rs
```diff
@@ -106,7 +106,11 @@ where
             let adversary = |_| new_adversary(num_good_nodes, num_faulty_nodes);
             let nonce = format!("My very unique nonce {:x}:{}", unique_id, i);
             info!("Nonce: {}", nonce);
-            let new_coin = |netinfo: _| ThresholdSign::new(netinfo, nonce.clone());
+            let new_coin = |netinfo: _| {
+                let mut newcoin = ThresholdSign::new(netinfo);
+                newcoin.set_message(nonce.clone()).unwrap();
```
afck:
```rust
expect("failed to set the new coin's ID")
```

src/threshold_sign.rs
```diff
@@ -100,26 +104,39 @@ impl<N: NodeIdT> DistAlgorithm for ThresholdSign<N> {
...
-            msg_hash: hash_g2(msg),
+            doc_hash: None,
             received_shares: BTreeMap::new(),
             had_input: false,
             terminated: false,
         }
     }
 
+    /// Sets doc_hash. Signature shares can only be sent after this function is completed.
+    pub fn set_document<M: AsRef<[u8]>>(&mut self, msg: M) -> Result<()> {
```
afck: The parameter should also be `doc` instead of `msg`.

src/threshold_sign.rs
```diff
@@ -100,26 +104,39 @@ impl<N: NodeIdT> DistAlgorithm for ThresholdSign<N> {
...
-        if self.had_input {
-            return Ok(Step::default());
-        }
+        let hash = match self.doc_hash {
+            Some(hash) => hash,
+            None => return Err(Error::DocumentHashIsNone),
+        };
         self.had_input = true;
+        let mut step = Step::default();
+        step.fault_log.extend(self.remove_invalid_shares());
```
afck: We need to make sure this step isn't dropped. It needs to say

```rust
return step.join(self.try_output());
```

in line 137 and

```rust
step.messages.push(Target::All.message(msg.clone()));
```

in line 140 now.

src/threshold_sign.rs
```diff
@@ -136,23 +153,53 @@ impl<N: NodeIdT> ThresholdSign<N> {
...
+        for id in faulty_senders {
+            self.received_shares.remove(&id);
+            fault_log.append(id, FaultKind::UnverifiedSignatureShareSender);
+        }
+        fault_log
+    }
+
+    /// Returns `true` if the share is valid, or if we don't have the message data yet.
+    fn is_share_valid(&self, id: &N, share: &SignatureShare) -> bool {
+        let msg = match self.doc_hash {
```
afck: `msg` → `doc_hash`

src/threshold_decryption.rs
```diff
@@ -115,22 +117,32 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
...
     /// If we have enough shares, outputs the plaintext.
-    pub fn set_ciphertext(&mut self, ct: Ciphertext) -> Result<Step<N>> {
+    pub fn set_ciphertext(&mut self, ct: Ciphertext) -> Result<()> {
         if self.ciphertext.is_some() {
             return Err(Error::MultipleInputs(Box::new(ct)));
         }
-        if !self.netinfo.is_validator() {
-            self.ciphertext = Some(ct);
-            return Ok(self.try_output()?);
+        if self.netinfo.secret_key_share().decrypt_share(&ct).is_none() {
```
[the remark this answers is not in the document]
alyjak: ah yeah, I should have looked at the docs here ... or read your comment all the way through!

src/threshold_decryption.rs
```diff
@@ -115,22 +117,32 @@ impl<N: NodeIdT> ThresholdDecryption<N> {
...
+        Ok(())
+    }
+
+    /// Sends our decryption shares to peers, and if we have collected enough, returns the decrypted
+    /// message. Returns an error if the ciphertext hasn't been received yet.
+    pub fn start_decryption(&mut self) -> Result<Step<N>> {
+        let ct = match self.ciphertext {
+            None => return Err(Error::CiphertextIsNone),
+            Some(ref ct) => ct.clone(),
+        };
```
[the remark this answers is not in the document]
alyjak: much cleaner! I wish I could do the same with the other matches in these changes that return a default step if doc_has is None...

src/threshold_sign.rs
alyjak: ah ouch, I presume if the tests weren't failing when I committed they would have caught this? If not, maybe I should write a test to try and catch something like this?

alyjak:
@afck Thank you for your patience here. I feel silly about that stack overflow, it seems obvious once you said it.
By the way, I tried to use gdb to find the stack overflow and didn't have much success, but I need practice. Do you guys rely on the debugger at all? It looks like gdb support for rust is fairly new ([Rust Debugging Quest](https://internals.rust-lang.org/t/rust-debugging-quest/6753))

alyjak:
Another comment: in addition to fixing the stack overflow and responding to your feedback, I incorporated your insight into the sign <-> try_output interaction into both `ThresholdSign` as well as `ThresholdDecrypt`.

src/threshold_sign.rs
afck: I know, our tests still leave a lot to be desired. (We currently also have two "network test frameworks", and need to port everything to the new one.)
They don't test that faults are actually reported at all yet. See #120—it would be great to make a start on that one! :+1:

c0gent:
> Do you guys rely on the debugger at all?

Rust drastically reduces the need to fire up `gdb` but it's still a very useful tool, particularly when dealing with problematic unsafe code and low level stuff (which `hbbft` has very little of).

src/threshold_sign.rs
```diff
@@ -99,28 +103,42 @@ impl<N: NodeIdT> DistAlgorithm for ThresholdSign<N> {
...
-            msg_hash: hash_g2(msg),
+            doc_hash: None,
             received_shares: BTreeMap::new(),
             had_input: false,
             terminated: false,
         }
     }
 
+    /// Sets doc_hash. Signature shares can only be sent after this function is completed.
+    pub fn set_document<M: AsRef<[u8]>>(&mut self, doc: M) -> Result<()> {
```
vkomenda: Every call to `set_document` follows a call to `ThresholdSign::new`. Can you call it from `new` instead or explain why you chose to keep them separate? It would look better if there were no separate post-initialization step.

src/threshold_sign.rs
```diff
@@ -60,7 +64,7 @@ impl Message {
 pub struct ThresholdSign<N> {
     netinfo: Arc<NetworkInfo<N>>,
     /// The hash of the data to be signed.
```
vkomenda: Maybe update the comment? "the document to be signed"?

src/threshold_sign.rs
afck: But in our use of `ThresholdDecryption`, it's different. And this issue is about bringing the two APIs in line with each other. So even if in our code, `new` and `set_document` are always called together, having them separate might be useful for external users.

vkomenda: I'm not entirely convinced though that you couldn't have either `new_with_document` or a builder which makes setting the document a requirement.

alyjak: True, we could add a second constructor that sets the doc on creation. To keep the API consistent we would want to do it for both modules. Is that extra function desired?

afck: That's true, that would be convenient. I'd add it to both modules, yes.

alyjak: Change implemented, both modules now have a `new_with_{document,ciphertext}` method.

vkomenda: Nice one. Thanks!

PR #317, opened 2018-11-01
Build reed-solomon-erasure on android

KORuL:
This fix to build reed-solomon-erasure on android

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ init_with = "1.1.0"
 log = "0.4.1"
 rand = "0.4.2"
 rand_derive = "0.3.1"
-reed-solomon-erasure = "3.1.0"
+reed-solomon-erasure = { version = "3.1", features = ["pure-rust"] }
 serde = "1.0.55"
 serde_derive = "1.0.55"
 threshold_crypto = "0.2.1"


afck:
That's definitely something we want to support! Thanks for also filing https://github.com/darrenldl/reed-solomon-erasure/issues/28!

For now, this change would disable SIMD on all platforms, would it? Running
```
time cargo run --example simulation --all-features --release -- --tx-size 10000 --bw 100000
```
goes from ~12.7s to ~13.4s on my machine, so it looks like there would be a small but measurable performance hit under some extreme (unrealistic?) workloads.
But still: Should we put this behind a feature flag in hbbft, too? Maybe `no-simd`?
And let's also add a TODO to remove it once https://github.com/darrenldl/reed-solomon-erasure/issues/28 is resolved.

darrenldl:
Hi, I'm the author of the crate.

So I have asked on Rust forum a few days ago, and it seems I have done the conditional compilation bit correctly, so I have no idea why it doesn't work for Android's case.

And yes, the "pure-rust" flag would disable SIMD at the current version, and it'll be the case until SIMD lands on stable Rust.

Note that "pure-rust" version is roughly 4x slower than normal version (sitting at roughly 1GB/s for my laptop), depending on threading etc, but shouldn't be a bottleneck for networked environment (even at 10gbit/s) from my understanding.

afck:
Thank you!
I think from our side it's okay for now to just put it behind a feature flag.
If we figure out how to properly conditionally compile it for Android, we'll let you know!

PR #319, opened 2018-11-05
Rewrite Subset.

### LINKED ISSUE #313: Subset: Drop Broadcast as soon as BinaryAgreement outputs false.
As soon as a `BinaryAgreement` instance in `Subset` has output `false`, all messages in the corresponding `Broadcast` instance are a waste of bandwidth.

afck:
This merges the four maps into a single map, each entry of which tracks
the progress of both the `Broadcast` and the `BinaryAgreement`
subalgorithm for an individual proposer. Two advantages:
* A slight optimization, because the `Broadcast` instance is dropped as
soon as it is not needed anymore.
* The static guarantee that certain impossible situations (inserting a
broadcast value twice) cannot happen.

The module is also split up into smaller files.

Closes #313.


vkomenda:
Is it possible to move `subset.rs` to `subset/subset.rs` instead of deleting?

afck:
I actually used `git mv`. (I had moved it to `subset/mod.rs` anyway, which doesn't really make sense.) But the files probably don't have much in common anymore. I don't think git actually uses these kinds of hints: If the files are too different, it doesn't consider it a move-and-edit anymore, but a delete-and-create.

c0gent:
You have to use two commits (move + changes) to get git to preserve history, etc.

afck:
Yes, `pop`ping the outputs is another issue where I'd prefer if we could statically ensure that there can be only one.
I feel like it was a mistake to make `Step` depend on `DistAlgorithm` (I know, I was the one who pushed for it :grimacing:). Maybe it should be generic in not what's currently the output type, but what's currently `Vec<Output>`… anyway, that's an issue for a separate PR, I guess.

PR #321, opened 2018-11-06
Move sender queue impls out of Honey Badger algorithms

vkomenda:
This leaves the HB, DHB and QHB modules completely free of sender queue methods.

Can me merged after #320.


afck:
Sorry, I should probably have closed #320 instead of clicking "Rebase and merge". :grimacing:

Issue #214, opened 2018-08-27
Support the Android target


vkomenda:
Ensure that `hbbft` works on Android. The tests should pass and the examples should run OK.

afck:
Also see #317 and https://github.com/darrenldl/reed-solomon-erasure/issues/28.

PR #323, opened 2018-11-06
Add a no-simd feature.

afck:
This disables SIMD in reed-solomon-erasure to work around problems on Android.

Closes #317.
See #214.


afck:
@KORuL: Please let us know whether this feature would work for you, too, instead of #317.

KORuL:
Yes, this feature work for me! Thanks!

PR #328, opened 2018-11-07
Simplify the sender queue.

afck:
Closes #325.

### DIFF
--- a/src/sender_queue/dynamic_honey_badger.rs
+++ b/src/sender_queue/dynamic_honey_badger.rs
@@ -9,7 +9,7 @@ use serde::{de::DeserializeOwned, Serialize};
 use super::{
     SenderQueue, SenderQueueableDistAlgorithm, SenderQueueableMessage, SenderQueueableOutput,
 };
-use {Contribution, DaStep, Epoched, NodeIdT};
+use {Contribution, DaStep, NodeIdT};
 
 use dynamic_honey_badger::{
     Batch, Change, ChangeState, DynamicHoneyBadger, Error as DhbError, Message, NodeChange,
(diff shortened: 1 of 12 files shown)


src/sender_queue/dynamic_honey_badger.rs
```diff
@@ -31,14 +30,16 @@ where
...
-            (era, epoch - era + 1)
-        } else {
-            (epoch + 1, 0)
+impl<N: Rand> Epoched for Message<N> {
+    type Epoch = (u64, u64);
+
+    fn epoch(&self) -> (u64, u64) {
+        match *self {
+            Message::HoneyBadger(era, ref msg) => (era, msg.epoch()),
+            Message::KeyGen(era, _, _) => (era, 0),
```
vkomenda: It's not a big thing but it would have been better to avoid assigning HB epoch 0 to non-HB messages. Assigning 0 does simplify a few things in a way that there are no type constraints to satisfy now.

afck: Yes, it's not ideal. But if I think of a message's epoch as the first epoch in which that message would make sense, then 0 is kind of correct.
I think it's only used to determine which key to use in the `outgoing_queue` map when deferring this message, so its meaning is kind of: "Once you arrive at this epoch, you can send out this message."
To make things less ambiguous, we could just not require that messages are `Epoched`, and instead make this a function of `SenderQueueableMessage` itself? Something like `first_epoch`?

vkomenda: Making DHB messages `Epoched` with `Epoch = (u64, u64)` drops the information about the messages being sent from the HB or not. Epochs `(x, 0)` are now ambiguous. In my opinion, it's better either to have the `(u64, Option<u64>)` epoch or derive epoch information from messages separately for each kind of message content within the message, possibly in `SenderQueueableMessage`.

I'm likely bikeshedding here. @c0gent or @mbr might have a better solution.

afck: Yes, I think messages just shouldn't be `Epoched`, because a message doesn't necessarily belong to a single epoch. Instead, I added `fn first_epoch` to `SenderQueueableMessage`.

PR #329, opened 2018-11-09
Make the incoming message queue finite in Binary Agreement

### LINKED ISSUE #43: Limit message caches and future epochs?

vkomenda:
Closes #43 

This is the second and last part of #43. In includes

- limitation on the number of future Binary Agreement epochs for which we accept and queue messages from remote nodes;

- finite representation of incoming messages in a future epoch.

### DIFF
--- a/src/binary_agreement/binary_agreement.rs
+++ b/src/binary_agreement/binary_agreement.rs
@@ -3,13 +3,15 @@ use std::sync::Arc;
 use std::{fmt, result};
 
 use bincode;
+use crypto::SignatureShare;
 use log::debug;
 
 use super::bool_multimap::BoolMultimap;
-use super::bool_set::BoolSet;
-use super::sbv_broadcast::{self, SbvBroadcast};
+use super::bool_set::{self, BoolSet};
+use super::sbv_broadcast::{self, Message as SbvMessage, SbvBroadcast};
 use super::{Error, Message, MessageContent, Result, Step};
-use threshold_sign::{self, ThresholdSign};
+use fault_log::{Fault, FaultKind};
+use threshold_sign::{self, Message as TsMessage, ThresholdSign};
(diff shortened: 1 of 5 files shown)


src/binary_agreement/binary_agreement.rs
```diff
@@ -38,6 +40,103 @@ impl<N> From<bool> for CoinState<N> {
...
+                } else {
+                    return Some(FaultKind::AgreementEpoch);
+                }
+            }
+        }
+        None
+    }
+
+    /// Creates message content from `ReceivedMessages`. That message content, together with a
+    /// target node ID, forms messages that can be sent to the target node.
```
afck: That comment is a bit confusing. It reads as if it were an outgoing queue. Maybe:
"~`together with a target node ID`~"
"`sent to the target node`" → "`handled`"

vkomenda: It's confusing indeed! Maybe we can reduce the last sentence even further to avoid speaking about conversion from `MessageContent` to `Message`? For example: "That message content can then be handled."

Issue #183, opened 2018-08-02
Remove the randomness feature from Honey Badger.


afck:
A randomness primitive is desirable for a lot of smart contract use cases. Basically it suffices to initialize a (possibly on-chain) PRNG with a new seed in each epoch that cannot be known before that epoch.

This is already achievable with hbbft, by using a separate Coin instance's output value as the pseudo-random seed. However, the coin must not be flipped before the epoch's batch has been determined, otherwise a validator would be able to make their choice of proposed transactions based on the random seed. So currently, the coin would have to be flipped after Honey Badger has output, which adds another round of messages, and therefore decreases block rate/increases lag.

We could make Honey Badger output twice for each epoch: first to notify the user that the new batch has been decided but not yet decrypted, and then, a bit later, to produce the actual batch. That would allow us to initiate the coin flip after the first output, in parallel with the decryption round.

Alternatively, we could make the randomness an optional feature of Honey Badger itself, so it would do the above internally, and just output a random value together with each batch. That might be a slightly more convenient API, but it would be less flexible: there could be use cases where one wants to do something other than flip a coin at the earliest possible moment where a batch is decided.

afck:
As a zero-overhead alternative:
[This value](https://github.com/poanetwork/threshold_crypto/blob/4c533e687331aa849b122a7bb7537b08321d6f07/src/lib.rs#L393) looks like it should be unknowable in advance and independent of the set of validators that provided a share. Maybe we could xor all those values from all accepted contributions, and use _that_ as the random seed?
(It's just going to be a bit ugly to modify the API so that we can extract it without overhead.)

afck:
I'm pretty confident that this would be suitable: For every individual contribution, we'd get the product of the secret master key with the proposer's chosen random value. But we'd sum up (or xor) _N - f_ of those, from different nodes. Some of those are honest and chose a genuinely random value.

With #209, however, we would either not produce a random value in every epochs, or we'd have to resort to another method (use the common coin after all) in those epochs.

afck:
#170 would probably make it insecure, however, since we'd learn about the other nodes' values before Subset terminates…

afck:
Let's implement this step-by-step:
* [ ] Add a "random value" option to Honey Badger. If it's enabled, a coin is flipped in every epoch, and its result is added as a random value to the batch.
[...]

PR #332, opened 2018-11-12
Add a random_value option to HB and DHB.

afck:
This includes a threshold signature in each batch that can be used as a pseudorandom value.

Also moves `EncryptionSchedule` from `threshold_decrypt` to `honey_badger`, and adds both encryption schedule and random value to the test.

See #183.

### DIFF
--- a/src/dynamic_honey_badger/batch.rs
+++ b/src/dynamic_honey_badger/batch.rs
@@ -1,6 +1,8 @@
 use std::collections::BTreeMap;
 use std::sync::Arc;
 
+use crypto::Signature;
+
 use super::EncryptionSchedule;
 use super::{ChangeState, JoinPlan};
 use {NetworkInfo, NodeIdT};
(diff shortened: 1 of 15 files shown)


src/dynamic_honey_badger/batch.rs
```diff
@@ -119,4 +124,12 @@ impl<C, N: NodeIdT> Batch<C, N> {
             && self.netinfo.public_key_map() == other.netinfo.public_key_map()
             && self.encryption_schedule == other.encryption_schedule
     }
+
+    /// Returns the signature that can be used as a pseudorandom value: None of the validators knew
```
vkomenda: Please correct the last sentence to make it more readable. Did you mean "None of the validators of the batch knew this signature before other validators decided."?

It reads like the first sentence might end with a period instead of the colon.

afck: Ah, I was meaning to write "before the other _fields_ of the batch were decided". I'll try to come up with a better formulation anyway…

Issue #335, opened 2018-11-12
Add an EventLog for collecting metrics on hbbft performance


alyjak:
In order to propagate performance information into an hbbft caller, add an `event_log` to `Step`s definition. Events will implement `failure::Fault`, simply because it is a convenient interface for defining event types as well as annotating them formatted Display messages.

By including event_log in `Step`, logs will be naturally extended across calls and provided to the initial caller.

EventLog has a similar signature to `FaultLog`

    pub struct `EventLog(pub Vec<Event>);

Events are similar to the `Fault` struct, but instead of node_id, they record a timestamp:

    pub struct Event<L: Fail> {
        logtime: std::time::Instant,
        content: L
    };

This way, the caller can record duration between two events using `std::time::Instant::duration_since`. `Instants` are monotonically increasing. They can be mapped to system time by logging a message who's contents is current OS system time. This way both clocks can be synchronized.

Besides measuring performance, event logs can be used by hbbft calling applications to detect soft failures, like a validator going off line, by collecting and monitoring logs over time.

If EventLog is added, a nice use case would be to add event logs to DistAlgorithm calls. So, the first time `foo.handle_message` is called it issues a log event, and when `foo.handle_output` produces an output a log event is also issued. Comparing this data between runs can be used to benchmark performance.

Issue #322, opened 2018-11-06
Port all tests to the new framework.


afck:
We have two integration test frameworks at the moment, in `tests/net` and `tests/network`. We need to port all the tests to the former, and remove the latter:

* [x] `binary_agreement_mitm.rs`
* [ ] `binary_agreement.rs`
* [ ] `binary_agreement.rs`
* [ ] `broadcast.rs`
* [ ] `dynamic_honey_badger.rs` (Or just delete—it's superseded by `net_dynamic_hb.rs`?)
* [ ] `honey_badger.rs`
* [x] `net_dynamic_hb.rs`
* [ ] `queueing_honey_badger.rs`
* [ ] `subset.rs`
* [ ] `sync_key_gen.rs` (if applicable?)
* [ ] `threshold_sign.rs`

vkomenda:
I'm not sure why there are two `binary_agreement.rs` tests in the list. I'll work on porting BA tests.

afck:
Thanks, I removed the duplicate.

vkomenda:
I recorded in the checklist that the test of Binary Agreement has been updated in #336.

vkomenda:
`test_threshold_sign_first_silent_50_samples` fails sometimes which may indicate a bias but it's currently non-reproducible because it instantiates the RNG with `thread_rng` instead of using the proptest `TestRng`. This bug must be fixed as part of this bigger issue.

dforsten:
test_threshold_sign_first_silent_50_samples is now using TestRng, tolerance increased to 33% as suggested by @vkomenda in the discussions in PR #387.

dforsten:
sync_key_gen.rs does not use the old network simulator, should probably be taken off the list in this issue.

dforsten:
dynamic_honey_badger.rs is not present in the test folder any more, should probably be taken off the list in this issue as well.

dforsten:
PR #389, once merged, will close this issue.

PR #336, opened 2018-11-13
Binary Agreement test updated to the proptest framework

vkomenda:
In the process of updating the BA test I had to remove the observer check temporarily from that test until we implement observers in the test framework. See #322.

I also removed logging from the test in advance of `EventLog`, see #335.

### DIFF
--- a/tests/binary_agreement.rs
+++ b/tests/binary_agreement.rs
@@ -1,9 +1,10 @@
 #![deny(unused_must_use)]
 //! Tests of the Binary Agreement protocol
 //!
-//! Only one proposer instance is tested. Each of the nodes in the simulated network run only one
-//! instance of Binary Agreement. This way we only test correctness of the protocol and not message
-//! dispatch between multiple proposers.
+//! Each of the nodes in the simulated network runs only one instance of Binary Agreement for node
+//! `0`, the proposer. In real applications there would be multiple proposers in the
+//! network. However, having only one proposer allows to test correctness of the protocol rather
+//! than message dispatch between multiple proposers.
 //!
 //! There are three properties that are tested:
 //!
@@ -23,7 +24,7 @@ extern crate proptest;
 extern crate rand;
 extern crate threshold_crypto;
 
-mod net;
+pub mod net;
 
 use std::iter::once;
 use std::sync::Arc;
@@ -40,15 +41,13 @@ use net::proptest::TestRng;
 use net::{NetBuilder, NewNodeInfo, VirtualNet};
 
 type NodeId = usize;
-type SessionId = u8;
-type Algo = BinaryAgreement<NodeId, SessionId>;
 
-impl VirtualNet<Algo> {
+impl VirtualNet<BinaryAgreement<NodeId, u8>> {
     fn test_binary_agreement<R>(&mut self, input: Option<bool>, mut rng: R)
     where
         R: Rng + 'static,
     {
-        let ids: Vec<NodeId> = self.nodes().map(|n| n.id().clone()).collect();
+        let ids: Vec<NodeId> = self.nodes().map(|n| *n.id()).collect();
         for id in ids {
             let _ = self.send_input(id, input.unwrap_or_else(|| rng.gen::<bool>()));
         }
@@ -86,6 +85,7 @@ fn test_binary_agreement_different_sizes() {
                 "Test start: {} good nodes and {} faulty nodes, input: {:?}",
                 num_good_nodes, num_faulty_nodes, input
             );
+            // Create a network with `size` validators and one observer.
             let mut net: VirtualNet<_> = NetBuilder::new(0..size)
                 .num_faulty(num_faulty_nodes)
                 .message_limit(10_000 * size as usize)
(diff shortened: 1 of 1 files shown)


tests/binary_agreement.rs
```diff
@@ -1,108 +1,137 @@
 #![deny(unused_must_use)]
-//! Tests of the Binary Agreement protocol. Only one proposer instance
-//! is tested. Each of the nodes in the simulated network run only one instance
-//! of Binary Agreement. This way we only test correctness of the protocol and not
-//! message dispatch between multiple proposers.
+//! Tests of the Binary Agreement protocol
+//!
+//! Each of the nodes in the simulated network runs only one instance of Binary Agreement for node
+//! `0`, the proposer. In real applications there would be multiple proposers in the
```
afck: I don't think that's generally the case: In _Honey Badger_, binary agreement happens to be about whether a particular proposer's contribution is accepted, but there may well be other uses. E.g. there are consensus algorithms where (something similar to) BA is used to determine whether a particular _transaction_ gets included in the next batch or not. Or imagine someone extends Honey Badger to make it "finite": Each epoch, the validators run a single additional BA instance, and if that return `true`, they all terminate.
So I'd vote for not mentioning the concept of a "proposer" here at all.

tests/binary_agreement.rs
```diff
@@ -1,108 +1,137 @@
...
+extern crate integer_sqrt;
+extern crate proptest;
 extern crate rand;
-extern crate rand_derive;
-extern crate serde_derive;
-extern crate threshold_crypto as crypto;
+extern crate threshold_crypto;
 
-mod network;
+pub mod net;
```
afck: That shouldn't be `pub`.

tests/binary_agreement.rs
```diff
@@ -1,108 +1,137 @@
...
-                BinaryAgreement::new(netinfo, 0).expect("Binary Agreement instance")
-            };
-            let network = TestNetwork::new(num_good_nodes, num_faulty_nodes, adversary, new_ba);
-            test_binary_agreement(network, input);
-            info!(
-                "Test success: {} good nodes and {} faulty nodes, input: {:?}",
-                num_good_nodes, num_faulty_nodes, input
-            );
+impl VirtualNet<Algo> {
+    fn test_binary_agreement<R>(&mut self, input: Option<bool>, mut rng: R)
```
afck: Should that really be a method of `VirtualNet`, instead of a separate function? (@mbr?)

tests/binary_agreement.rs
```diff
@@ -1,108 +1,137 @@
...
-    test_binary_agreement_different_sizes(new_adversary);
+/// Tests Binary Agreement on a given configuration with random inputs, with all `false` inputs and
+/// with all `true` inputs.
+#[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
+fn binary_agreement(cfg: TestConfig) {
+    let mut rng: TestRng = TestRng::from_seed(cfg.seed);
+    let size = cfg.dimension.size();
+    let num_faulty_nodes = cfg.dimension.faulty();
+    let num_good_nodes = size - num_faulty_nodes;
+    for &input in &[None, Some(false), Some(true)] {
```
afck: Should this be handled by proptest, too?

vkomenda: OK, will do.

tests/binary_agreement.rs
vkomenda: As a private method, it makes a perfect sense to me.

tests/binary_agreement.rs
vkomenda: There are lots of "method is never used" lints otherwise in the `net` module.

afck: Right; some of them are already marked as `#[allow(unused)]`—maybe we should apply that attribute to the whole `net` module instead (either inside it, or at the `use` statements)? That would make the reason clearer than the `pub`.

mbr: One option would also be to make the testing framework a proper crate and just import that; I've done this on another project.

It's a bit more work, but it also does not require importing a million different crates on each test cases - right now, whenever I add a dependency to the `net` testing framework, I also need to add one in every test case.

vkomenda:
@mbr, please check the usage of `VirtualNet`.

tests/binary_agreement.rs
mbr: I'm skeptical as well. I don't have a great retort at this point though =).

tests/binary_agreement.rs
```diff
@@ -1,108 +1,140 @@
...
+}
 
-    // Handle messages in random order until all nodes have output the proposed value.
-    while !network.nodes.values().all(TestNode::terminated) {
-        network.step();
+prop_compose! {
+    /// Strategy to generate a test configuration.
+    fn arb_config()
+        (
+            dimension in NetworkDimension::range(1, 50),
```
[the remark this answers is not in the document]
afck: I'd say we definitely want BA to work with a size of 1.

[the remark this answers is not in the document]
vkomenda: If `proptest` reduces it to 1 in case of failure anyway, maybe it's worth staring from above 1.

PR #338, opened 2018-11-13
Added fault checking on every step in the net framework

vkomenda:
See #120.

Tests will fail if there is a step with a fault report referring to a correct node. This PR also adds storage for the fault log in the test node struct alongside output storage.

### DIFF
--- a/tests/net/mod.rs
+++ b/tests/net/mod.rs
@@ -270,7 +270,7 @@ where
         .store_step(step);
     // Verify that no correct node is reported as faulty.
     for fault in &step.fault_log.0 {
-        if nodes.contains_key(&fault.node_id) {
+        if nodes.get(&fault.node_id).map(|n| !n.is_faulty()) == Some(true) {
             panic!("Unexpected fault: {:?}", fault);
         }
     }
--- a/tests/net/mod.rs
+++ b/tests/net/mod.rs
@@ -26,7 +26,7 @@ use threshold_crypto as crypto;
 
 use hbbft::dynamic_honey_badger::Batch;
 use hbbft::util::SubRng;
-use hbbft::{self, Contribution, DaStep, DistAlgorithm, NetworkInfo, NodeIdT, Step};
+use hbbft::{self, Contribution, DaStep, DistAlgorithm, Fault, NetworkInfo, NodeIdT, Step};
 
 use try_some;
 
@@ -73,6 +73,8 @@ pub struct Node<D: DistAlgorithm> {
     is_faulty: bool,
     /// Captured algorithm outputs, in order.
     outputs: Vec<D::Output>,
+    /// Collected fault log.
+    faults: Vec<Fault<D::NodeId>>,
 }
 
 impl<D> fmt::Debug for Node<D>
@@ -96,6 +98,7 @@ impl<D: DistAlgorithm> Node<D> {
             algorithm,
             is_faulty,
             outputs: Vec::new(),
+            faults: Vec::new(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


vkomenda:
> Not sure whether the store_step functionality should better be left to the individual test?

`VirtualNet::new` produces a `Step`s internally but doesn't outputs those. It calls `process_step` from where the step can be stored.

We could output steps from `new` of course if it's better.

afck:
I see; returning them from `VirtualNet::new` sounds like the right thing to do then, but maybe there's a reason why that's not the case? (@mbr)

vkomenda:
@mbr, your input would be welcome on [this](https://github.com/poanetwork/hbbft/pull/338#issuecomment-438597983).

mbr:
The basic design philosophy behind `VirtualNet` is that it makes _all_ the steps available and everything is passed back to the test (which may choose to ignore the faults, for example, when checking for the presence of them).

However, all `Steps` are passed through processing functions to copy out all messages, which are handled automatically by the network. The fact that outputs are collected is just a convenience bonus.

So, you're right, the fact that `new` does not pass out the `Steps` is probably a mistake. Furthermore, collecting faults like outputs is definitely a possibility, however we should not `panic` on these. At most, we could add an optional `panic_on_fault` convenience method. Ideally though, there'd be no panic, and we'd introduce a new error for the crank function and others.

mbr:
So, to summarize, we actually need to solve three issues?

1. Make `new` (and hopefully all other functions) return the generated `Steps` for inspection by the actual test.
2. Collect `FaultLog`s as a convenience.
3. Add a newoptional failure condition (might even be on by default) that causes an error if a fault is encountered.

vkomenda:
Thanks, @mbr. I'll get following through your points.

vkomenda:
@mbr, please check my response to your review comments.

tests/net/mod.rs
```diff
@@ -73,6 +72,8 @@ pub struct Node<D: DistAlgorithm> {
     is_faulty: bool,
     /// Captured algorithm outputs, in order.
     outputs: Vec<D::Output>,
+    /// Collected fault log.
```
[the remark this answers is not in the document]
vkomenda: Can you explain what you mean by collecting in order?

tests/net/mod.rs
```diff
@@ -244,14 +264,19 @@ where
...
-    // Collect all outputs (not required for network operation) as a convenience for the user.
     nodes
         .get_mut(&sender)
         .expect("Trying to process a step with non-existing node ID")
-        .outputs
-        .extend(step.output.iter().cloned());
-
-    message_count
+        .store_step(step);
+    if error_on_fault {
```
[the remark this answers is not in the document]
vkomenda: I want to check for faults whenever an algorithm makes a step. This goes outside the scope of `crank`. In fact, it matches calls to `process_step` one-to-one.

tests/net/mod.rs
```diff
@@ -318,6 +343,9 @@ where
     message_limit: Option<usize>,
     /// Optional time limit.
     time_limit: Option<time::Duration>,
+    /// Property to cause an error if a `Fault` is output from a correct node. By default,
+    /// encountering a fault leads to an error.
+    error_on_fault: bool,
```
afck: Do we ever want to turn that off?
I imagine that we'll always want any tests to fail if a correct node blames a correct node?

tests/net/mod.rs
```diff
@@ -456,6 +486,16 @@ where
         self
     }
 
+    /// Property to cause an error if a `Fault` is output from a correct node. By default,
+    /// encountering a fault leads to an error.
+    ///
+    /// The deault setting `true` can be changed using this function.
```
afck: Typo: `default`

tests/net/mod.rs
vkomenda: There was a comment asking to have this option but I cannot find it now.

vkomenda: By disabling errors on faults you would be able to see more faults and get stalling or divergence.

afck: Not sure if that's really useful, but I'm fine with it, too.

vkomenda:
Now we have a maintenance task: the `crossbeam` dependency of the old example has to be updated or the example should be removed.

PR #339, opened 2018-11-14
Allow arbitrary validator set changes in DHB.

### LINKED ISSUE #330: Allow arbitrary validator set changes in DHB.
Instead of only adding or removing a node, `DynamicHoneyBadger` should allow voting for arbitrary new sets of validators that could even (but don't have to) be completely disjoint from the current one. This generalizes the current approach, and allows changing _n_ validators with a single key generation step, instead of _n_.

afck:
This replaces `NodeChange` with a full list of IDs and public keys, instead of just a single to-be-added or to-be-removed node, to allow completely replacing the set of validators by any arbitrary new set in a single key generation step.

Closes #330.

### DIFF
--- a/src/sender_queue/mod.rs
+++ b/src/sender_queue/mod.rs
@@ -41,9 +41,9 @@ pub trait SenderQueueableOutput<N, M>
 where
     N: NodeIdT,
 {
-    /// Returns the new set of validator that this batch is starting key generation for, including
-    /// the existing ones. These should be added to the set of all nodes.
-    fn new_nodes(&self) -> Vec<N>;
+    /// Returns the new set of validator that this batch is starting or completing key generation
+    /// for, including the existing ones. These should be added to the set of all nodes.
+    fn added_peers(&self) -> Vec<N>;
 }
 
 pub trait SenderQueueableDistAlgorithm: Epoched + DistAlgorithm {
@@ -206,7 +206,7 @@ where
             return Step::<D>::default();
         }
         // Look up `DynamicHoneyBadger` epoch updates and collect any added peers.
-        for node in step.output.iter().flat_map(|batch| batch.new_nodes()) {
+        for node in step.output.iter().flat_map(|batch| batch.added_peers()) {
             if &node != self.our_id() {
                 self.peer_epochs.entry(node).or_default();
             }
--- a/src/sender_queue/mod.rs
+++ b/src/sender_queue/mod.rs
@@ -41,9 +41,9 @@ pub trait SenderQueueableOutput<N, M>
 where
     N: NodeIdT,
 {
-    /// Returns an optional new node added with the batch. This node should be added to the set of
-    /// all nodes.
-    fn added_node(&self) -> Option<N>;
+    /// Returns the new set of validator that this batch is starting key generation for, including
+    /// the existing ones. These should be added to the set of all nodes.
+    fn new_nodes(&self) -> Vec<N>;
 }
 
 pub trait SenderQueueableDistAlgorithm: Epoched + DistAlgorithm {
@@ -206,7 +206,7 @@ where
             return Step::<D>::default();
         }
         // Look up `DynamicHoneyBadger` epoch updates and collect any added peers.
-        for node in step.output.iter().filter_map(|batch| batch.added_node()) {
+        for node in step.output.iter().flat_map(|batch| batch.new_nodes()) {
             if &node != self.our_id() {
                 self.peer_epochs.entry(node).or_default();
             }
--- a/tests/network/mod.rs
+++ b/tests/network/mod.rs
@@ -105,24 +105,20 @@ impl MessageScheduler {
         &self,
         nodes: &BTreeMap<D::NodeId, TestNode<D>>,
     ) -> D::NodeId {
-        match *self {
-            MessageScheduler::First => nodes
-                .iter()
-                .find(|(_, node)| !node.queue.is_empty())
-                .map(|(id, _)| id.clone())
-                .expect("no more messages in queue"),
-            MessageScheduler::Random => {
-                let ids: Vec<D::NodeId> = nodes
-                    .iter()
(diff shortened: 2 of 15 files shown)


afck:
I prepared the hydrabadger update here: https://github.com/poanetwork/hydrabadger/tree/afck-change

src/sender_queue/mod.rs
```diff
@@ -41,9 +41,9 @@ pub trait SenderQueueableOutput<N, M>
 where
     N: NodeIdT,
 {
-    /// Returns an optional new node added with the batch. This node should be added to the set of
-    /// all nodes.
-    fn added_node(&self) -> Option<N>;
+    /// Returns the new set of validator that this batch is starting key generation for, including
+    /// the existing ones. These should be added to the set of all nodes.
```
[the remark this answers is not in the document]
afck: Good point… that's a more general issue that I'm still unclear about:
Do we want to have the sender queue automatically manage the set of observer nodes or should the user also have some control over that (to e.g. add an observer that isn't about to join as a validator)?
* In the first case, we'd probably also want to _remove_ peers on `ChangeState::Complete`, in addition to adding them on `ChangeState::InProgress`, so that the set of peers is always exactly the union of the current and the next set of validators.
* In the second case, I'm not sure: Either we'd remove this feature completely and _require_ the user to add and remove observers, or we'd maintain an additional list of observers in `SenderQueue` (those would be the ones that DHB will never know about; they only observe, without ever becoming a validator).

For now, maybe I should actually implement the first point right away? It probably wouldn't be much work and I think it's mainly the intended behavior of the current code?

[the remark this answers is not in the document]
vkomenda: Our sender queue is a reference implementation which, I think, should be as simple as possible for that reason. I tend to choose the second option. In principle, the user can fully control the peers and possibly whether or not new peers join as observers only. Removing old peers however should be totally fine in our implementation. The user can change that in their implementation.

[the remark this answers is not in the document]
afck: I just realized that removing peers is actually more complicated than I thought:
If we remove a peer in epoch 42, we'll still need to deliver their messages from epochs up to 41. I think that'll require some changes to `update_epoch`? I'm not entirely sure. Maybe we should do that as a separate PR after all?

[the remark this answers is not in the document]
afck: Yes, I think that would work. I created https://github.com/poanetwork/hbbft/issues/341 for it.

afck:
Looks like I broke `test test_dynamic_honey_badger_first_delivery_silent`… investigating… :confused:

afck:
OK, I found the problem. It's just the "first delivery" thing again, and the fact that I effectively increased the required threshold for key generation (which is the right thing to do, I think): In the test, nodes 0 to 5 just keep producing batches and the others never get to send a message.
I'll try to fix the test tomorrow.

afck:
Please take another look; the "`first`" delivery tests now choose a random node in every 10th step.

tests/network/mod.rs
```diff
@@ -105,24 +105,20 @@ impl MessageScheduler {
...
-                    .expect("no more messages in queue")
-                    .clone()
-            }
-        }
+        let mut ids = nodes
+            .iter()
+            .filter(|(_, node)| !node.queue.is_empty())
+            .map(|(id, _)| id.clone());
+        let rand_node = match *self {
+            MessageScheduler::First => rand::thread_rng().gen_weighted_bool(10),
```
vkomenda: What if the user implements first non-randomized delivery? Wouldn't DHB fail then?

vkomenda: My point is that previously `First` used to work. Is there a fix of the algorithm without changing the test?

afck: Yes: What broke it was that in `is_ready` I'm now requiring _N - f_ signatures, which is what we should always expect to be able to reach. I could revert that, but I think it's safer this way. And I think it just exposed what was wrong with the test in the first place.

afck: The reason why I changed it in this PR is that before, we had the criterion: _f + 1_ complete `Part`s _and_ the candidate's `Part` must be complete.

Now that the candidates could potentially be _all_ of the new nodes, that criterion would actually be too strong. So I changed it to the strongest criterion that we can always expect to be satisfied, under the assumptions the algorithms are making anyway.

vkomenda: OK. This makes the workaround a requirement for validator set changes. How would this map to the networking layer? Would you expect randomization of message delivery there too? This makes it impossible to use `hbbft` in with deterministic delivery such as the previous `First`.

afck: In the real world the "first delivery" scenario is unrealistic, and corresponds to a network that would completely block the connection between C and D as long as there are messages sent from A to B. The requirement has not changed:
> The adversary is allowed to control the network as long as every message from a correct node to a correct node eventually arrives.

This requirement was not satisfied by the "first" schedule. That's why the test was wrong, and we were just "lucky" that it passed so far.

vkomenda: I agree in general. Now you essentially have two randomized delivery strategies. I think we still need a deterministic one, round-robin for example. I'm not sure what priority that would have given that we move all tests to `net`.

afck: I agree, in the long run it would be good to have several different "extreme" delivery strategies (while still satisfying the above requirement). We could easily make this one deterministic, too, by making it do round-robin instead of randomizing in every 10th round.
But I didn't want to spend too much time on implementing delivery strategies for the old test network.

vkomenda: I'm getting more convinced about this change but still think that the new validator set feature is not explored enough to be deemed safe and moreover the old test failed on it. This is too big for a single-person review. @c0gent, @mbr, please check if you are happy with @afck's ["real world" argument](https://github.com/poanetwork/hbbft/pull/339/files#r233843400).

Buyer beware: adversaries would stretch the real world boundaries.

PR #340, opened 2018-11-14
An optional observer node in the net framework

### LINKED ISSUE #337: Add observer nodes in the proptest framework
`VirtualNet` doesn't have a notion of observer built in, [unlike](https://github.com/poanetwork/hbbft/blob/767944c0f6a69ef362d96a812a5282e2dbf3ac80/tests/network/mod.rs#L372) `TestNetwork` of the old test framework, see #81. For porting old tests to the `proptest` framework - see #322 - it is therefore desired that we integrate observers into `VirtualNet`.

vkomenda:
An observer node in `VirtualNet` serves the purpose of porting old tests, cf. #322.

Closes #337.

### DIFF
--- a/tests/net/mod.rs
+++ b/tests/net/mod.rs
@@ -22,7 +22,7 @@ use std::{cmp, collections, env, fmt, fs, io, ops, process, time};
 
 use rand;
 use rand::{Rand, Rng};
-use threshold_crypto as crypto;
+use threshold_crypto::{self as crypto, SecretKeyShare};
 
 use hbbft::dynamic_honey_badger::Batch;
 use hbbft::util::SubRng;
@@ -302,7 +302,7 @@ pub struct NetBuilder<D, I>
 where
     D: DistAlgorithm,
 {
-    /// Iterator used to create node ids.
+    /// Iterator used to create validator node ids.
     node_ids: I,
     /// Number of faulty nodes in the network.
     num_faulty: usize,
@@ -318,6 +318,9 @@ where
     message_limit: Option<usize>,
     /// Optional time limit.
     time_limit: Option<time::Duration>,
+    /// Optional observer node ID. It must not be a member of `node_ids` for the latter are
+    /// validator node IDs.
+    observer_id: Option<D::NodeId>,
     /// Random number generator used to generate keys.
     rng: Option<Box<dyn Rng>>,
 }
@@ -336,6 +339,7 @@ where
             .field("crank_limit", &self.crank_limit)
             .field("message_limit", &self.message_limit)
             .field("time_limit", &self.time_limit)
+            .field("observer_id", &self.observer_id)
             .field("rng", &"<RNG>")
             .finish()
     }
@@ -368,6 +372,7 @@ where
             message_limit: None,
             time_limit: DEFAULT_TIME_LIMIT,
             rng: None,
+            observer_id: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


tests/net/mod.rs
```diff
@@ -513,9 +527,13 @@ where
...
         #[cfg_attr(feature = "cargo-clippy", allow(redundant_closure))]
-        let mut net = VirtualNet::new(self.node_ids, self.num_faulty as usize, rng, move |node| {
-            cons(node)
-        })?;
+        let mut net = VirtualNet::new(
+            self.node_ids,
+            self.num_faulty as usize,
+            rng,
+            self.observer_id,
+            move |node| cons(node),
```
afck: Isn't this equivalent to just `cons`?

vkomenda: No, it doesn't compile without the closure.

mbr:
I am not happy with this yet, for several reasons:

1. We are still missing a good definition of _observer_, here's the latest from @afck 

   > A node without a threshold key share, that doesn't _send_ messages as part of the algorithms (except DHB while joining) and only _receives_ `Target::All` messages.

   One huge issue here naturally is "except DHB", since it is very counter-intuitive to have observer nodes send anything at all. We may need a better definition, or just reduce it to something like

   > A node without a threshold key share, that does not count toward the network size `N` or faulty node count `f`.

   It may be advantageous to separate the concept of observers from joining nodes in DHB entirely though.

1. Making observers an overlay by tracking them, but otherwise treating them as normal nodes (they themselves do not know they are observers with the implementation in this PR) might cause issues down the road. I believe it's much better to add an `is_observer` to `Node`, update the relevant iterators and message dispatching logic.

1. With that said, why only have one observer? We could have multiple, sure it _should_ not make a difference for most algorithms, but testing that it *actually* does not do so is very much in the scope of things to test for?

   Additionally, if we are to implement a new, more generic DHB that allows swapping out a validator set completely instead of just adding or dropping one node, this functionality might be worthwhile anyway. While we do not need to make affordances for these hypotheticals right now, I think the multiple-observers implementation is more natural and not more complicated than a single one anyway!

1. It might be a good idea to allow enabling selective, automatic filtering of messages from observers (by panic'ing if they ever send some). In this model, we are shifting the burden of making sure observers do not interfere to the outer layer, which is already in charge of making sure messages coming in are authenticated.

   This functionality must come with an off-switch for DHB. Alternative, we could approach this from the other way around and ensure algorithms handle these issues by introducing adversaries that inject message from observers.

1. Documentation must be updated, especially regarding network sizes. We are not counting observer nodes in the network sizes usually, as they are not relevant and should be ignored from a security perspective (and therefore should also not be able to affect the network, weirdness of DHB aside).

1. The interface then probably needs a separate construction function for observer nodes and a way to specify how many should be constructed.

These are just some points to take into account. All-in-all, I think this is a bit of a bigger issue, which is why I put it off.

afck:
I agree that we need to properly document the observer concept (if we keep it).

Regarding the "weirdness" of DHB: I don't think it's _that_ weird. It's just that we're in a situation with an _old_ and a _new_ set of validators, where the new set of validators needs to perform DKG before it can fully take over. Since we do DKG "on-chain", using the HB that is still run by the old set of validators, the two sets temporarily overlap: During DKG, the new nodes need to _use_ the old nodes' network, kind of as "clients", because they need to use an existing consensus process for their DKG. They send their DKG messages as their input to the old network, and they _observe_ it so they can see the output.

Anyway, maybe the definition should just be that then: doesn't have a key share but receives `Target::All` messages.

> these hypotheticals

Uh… these hypotheticals will be merged soon: #339. :grimacing:

vkomenda:
@mbr, to sum up your comments, would this list of changes be sufficient?

- [ ] Allow for multiple observers (while still keeping them in `nodes` as at present).

- [ ] Add a getter for the subset of nodes which are observers (in the same way as the getters fo correct and faulty nodes; that would cover your `is_observer` requirement).

- [ ] Add a separate constructor in `Node` for an observer node.

- [ ] Update the test documentation with the definition of observer.

PR #342, opened 2018-11-15
Add epoch getters to JoinPlan, HB, DHB and QHB.

afck:


afck:
This will make https://github.com/poanetwork/hydrabadger/pull/18 less ugly.

PR #343, opened 2018-11-16
Splits up FaultKind into module specific enums

### LINKED ISSUE #285: Split up `FaultKind`.
`FaultKind`s should form a hierarchy similar (and parallel) to errors, and live in the respective modules.

Let's create a `FaultKind` enum for each algorithm, with `SubAlgorithm(sub_algorithm::FaultKind)` variants where needed (like currently with `AckMessageFault`), and possibly with corresponding `From` impls. I think even `Fault` and `FaultLog` should be generic in the fault kind, so that if you're using `BinaryAgreement`, the compiler doesn't force you to consider `HoneyBadger` faults.

They should all implement `Display`, and print an explanation (basically what's currently the doc comment). Maybe we should derive `Fail` to make this easier?

alyjak:
this will close #285

### DIFF
--- a/src/dynamic_honey_badger/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger/dynamic_honey_badger.rs
@@ -12,10 +12,10 @@ use serde::{de::DeserializeOwned, Serialize};
 use super::votes::{SignedVote, VoteCounter};
 use super::{
     Batch, Change, ChangeState, DynamicHoneyBadgerBuilder, EncryptionSchedule, Error, ErrorKind,
-    Input, InternalContrib, KeyGenMessage, KeyGenState, Message, NodeChange, Result,
+    FaultKind, Input, InternalContrib, KeyGenMessage, KeyGenState, Message, NodeChange, Result,
     SignedKeyGenMsg, Step,
 };
-use fault_log::{Fault, FaultKind, FaultLog};
+use fault_log::{Fault, FaultLog};
 use honey_badger::{self, HoneyBadger, Message as HbMessage};
 
 use sync_key_gen::{Ack, AckOutcome, Part, PartOutcome, SyncKeyGen};
@@ -56,6 +56,7 @@ where
     type Output = Batch<C, N>;
     type Message = Message<N>;
     type Error = Error;
+    type FaultKind = FaultKind;
 
     fn handle_input(&mut self, input: Self::Input) -> Result<Step<C, N>> {
         // User contributions are forwarded to `HoneyBadger` right away. Votes are signed and
@@ -228,7 +229,7 @@ where
         sender_id: &N,
         kg_msg: KeyGenMessage,
         sig: Signature,
-    ) -> Result<FaultLog<N>> {
+    ) -> Result<FaultLog<N, FaultKind>> {
         if !self.verify_signature(sender_id, &sig, &kg_msg)? {
             let fault_kind = FaultKind::InvalidKeyGenMessageSignature;
             return Ok(Fault::new(sender_id.clone(), fault_kind).into());
@@ -258,7 +259,12 @@ where
         hb_step: honey_badger::Step<InternalContrib<C, N>, N>,
     ) -> Result<Step<C, N>> {
         let mut step: Step<C, N> = Step::default();
-        let output = step.extend_with(hb_step, |hb_msg| Message::HoneyBadger(self.era, hb_msg));
+        let to_fault = |hb_fault: Fault<N, honey_badger::FaultKind>| {
+            Fault::new(hb_fault.node_id, FaultKind::HbFault(hb_fault.kind))
+        };
+        let output = step.extend_with(hb_step, to_fault, |hb_msg| {
+            Message::HoneyBadger(self.era, hb_msg)
+        });
         for hb_batch in output {
             let batch_era = self.era;
             let batch_epoch = hb_batch.epoch + batch_era;
--- a/src/dynamic_honey_badger/error.rs
+++ b/src/dynamic_honey_badger/error.rs
@@ -92,3 +92,51 @@ impl Display for Error {
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
+
+/// Represents each way an an incoming message can be considered faulty.
+#[derive(Debug, Fail, PartialEq)]
+pub enum FaultKind {
+    #[fail(
+        display = "`DynamicHoneyBadger` received a key generation message with an invalid signature."
+    )]
+    InvalidKeyGenMessageSignature,
+    #[fail(
+        display = "`DynamicHoneyBadger` received a key generation message with an invalid era."
+    )]
+    InvalidKeyGenMessageEra,
+    #[fail(
+        display = "`DynamicHoneyBadger` received a key generation message when there was no key
+                    generation in progress."
+    )]
+    UnexpectedKeyGenMessage,
+    #[fail(
+        display = "`DynamicHoneyBadger` received a signed `Ack` when no key generation in progress."
+    )]
+    UnexpectedKeyGenAck,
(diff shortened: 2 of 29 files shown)


src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -258,7 +259,12 @@ where
         hb_step: honey_badger::Step<InternalContrib<C, N>, N>,
     ) -> Result<Step<C, N>> {
         let mut step: Step<C, N> = Step::default();
-        let output = step.extend_with(hb_step, |hb_msg| Message::HoneyBadger(self.era, hb_msg));
+        let to_fault = |hb_fault: Fault<N, honey_badger::FaultKind>| {
+            Fault::new(hb_fault.node_id, FaultKind::HbFault(hb_fault.kind))
```
afck: Maybe this functionality should be extracted into a `Fault::map` convenience method, so this line would be easier on the eye:

```rust
hb_fault.map(FaultKind::HbFault)
```

src/dynamic_honey_badger/error.rs
```diff
@@ -92,3 +92,51 @@ impl Display for Error {
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
+
+/// Represents each way an an incoming message can be considered faulty.
+#[derive(Debug, Fail, PartialEq)]
+pub enum FaultKind {
```
afck: Not sure whether it should go into `error` or a separate `fault` submodule (same in the other algorithms). Let's see what the other reviewers say.

src/honey_badger/epoch_state.rs
```diff
@@ -38,8 +38,10 @@ impl<N: NodeIdT> SigningState<N> {
     fn handle_message(&mut self, sender_id: &N, msg: ts::Message) -> ts::Result<ts::Step<N>> {
         match self {
             SigningState::None => {
-                let fault_kind = FaultKind::UnexpectedSignatureShare;
-                Ok(Fault::new(sender_id.clone(), fault_kind).into())
+                let fault_kind = ts::FaultKind::UnexpectedSignatureShare;
+                let mut step = ts::Step::default();
+                step.fault_log.append(sender_id.clone(), fault_kind);
+                Ok(step)
```
afck: Why was this change necessary? Shouldn't `step` still implement `From<Fault<…>>`?

src/honey_badger/epoch_state.rs
```diff
@@ -370,7 +372,10 @@ where
     /// Checks whether the subset has output, and if it does, sends out our decryption shares.
     fn process_subset(&mut self, cs_step: CsStep<N>) -> Result<Step<C, N>> {
         let mut step = Step::default();
-        let cs_outputs = step.extend_with(cs_step, |cs_msg| {
+        let to_fault = |cs_fault: Fault<N, cs::FaultKind>| {
+            Fault::new(cs_fault.node_id, FaultKind::SubsetFault(cs_fault.kind))
+        };
+        let cs_outputs = step.extend_with(cs_step, to_fault, |cs_msg| {
```
afck: Actually, we'll never want to change the node ID when converting fault logs, so `extend_with` should only take a map from a fault _kind_ to fault kind:

```rust
step.extend_with(cs_step, FaultKind::SubsetFault, |cs_msg| { … })
```

src/traits.rs
```diff
@@ -92,26 +98,41 @@ impl<M, O, N> Step<M, O, N> {
...
     where
+        F2: Fail,
         FO: Fn(O) -> O2,
+        FF: Fn(Fault<N, F>) -> Fault<N, F2>,
         FM: Fn(M) -> M2,
     {
         Step {
             output: self.output.into_iter().map(f_out).collect(),
-            fault_log: self.fault_log,
+            fault_log: self.fault_log.into_iter().map(f_fail2fail).collect(),
```
afck: (See comment above.) This should probably convert only the fault kind, and could use the `Fault::map` method:

```rust
fault_log: self.fault_log.into_iter().map(|fault| fault.map(f_fault)).collect(),
```

(Same for `extend_with`.)

src/traits.rs
```diff
@@ -92,26 +98,41 @@ impl<M, O, N> Step<M, O, N> {
...
 
-    /// Converts `self` into a step of another type, given conversion methods for output and
-    /// messages.
-    pub fn map<M2, O2, FO, FM>(self, f_out: FO, f_msg: FM) -> Step<M2, O2, N>
+    /// Converts `self` into a step of another type, given conversion methods for output, faults,
+    /// and messages.
+    pub fn map<M2, O2, F2, FO, FF, FM>(
+        self,
+        f_out: FO,
+        f_fail2fail: FF,
```
afck: Maybe just `f_fault`?

src/dynamic_honey_badger/error.rs
c0gent: I don't think it should be in the `error` module, as faults are not errors. I'd just put it in `mod.rs`.

src/fault_log.rs
```diff
@@ -1,134 +1,74 @@
...
+//! `DistAlgorihm`'s `.handle_input()` and `.handle_message()` trait methods.
+
+pub use failure::Fail;
 
 /// A structure representing the context of a faulty node. This structure
 /// describes which node is faulty (`node_id`) and which faulty behavior
 /// that the node exhibited ('kind').
 #[derive(Debug, PartialEq)]
-pub struct Fault<N> {
+pub struct Fault<N, F: Fail> {
```
c0gent: What is the reasoning for replacing the `FaultKind` enum with a generic field? I don't see any clear benefits. 

This method has several drawbacks including the fact that `Fault`s of different kinds would have to be boxed when stored in a collection, etc., not to mention having an extra type parameter everywhere.

What's wrong with nested enums?

alyjak: The intent was to have each module's `FaultKind` defined within the module itself.

If we do want `FaultKind`'s defined in their modules, AND if we want to keep hbbft submodules fairly independent from one another, then I think we have to keep `FaultKind` generic. With nested enums, we still have to identify what type gets nested, which means that we have to import each nested `FaultKind` into the fault_log module, which would mean that every module that uses fault_log indirectly uses every other module. I'm pretty sure that is worth avoiding, right?

I agree that having another generic within `Step` is awkward, so in practice this pull request may not be worthwhile.

A more minor improvement  could be defining submodule `FaultKind`s within the fault_log module and making `FaultKind` enums which nest those values, or we could keep the current behavior where all faultkinds are flat listed. There's no performance improvement with this pull request, it only tries to clarify ownership.

c0gent: I'd just import all of the fault kinds into `fault_log`, renaming each of them to `{module}Fault`, and restore the original `FaultKind` enum. I realize that there will be circular dependencies but it's typical for error handling and isn't a problem.

It would also be fine to move all of the sub-fault kinds into the `fault_log` module. It really doesn't matter.

afck: I'd actually prefer the approach with the generic: It's just exactly parallel to the hierarchy of errors. And I don't think there's a situation where we'd ever have to `Box` them: Depending on which algorithm you use, you know which `FaultKind` you get, just like with errors. It avoids unreachable branches in `match` statements and better decouples the modules: There are no `HoneyBadger` variants in the faults returned from `BinaryAgreement`.
Also, we have a type alias for `Step` in each of the algorithms' modules which fixes the additional generic argument, so it doesn't inconvenience the user.

c0gent: Fair enough.

mbr:
Should I volunteer to tackle the rebasing required here?

afck:
@phahulin: Can this be closed in favor of #371?

phahulin:
I think so, yes.

PR #344, opened 2018-11-19
Random value optimization

vkomenda:
See #183, especially the [comment](https://github.com/poanetwork/hbbft/issues/183#issuecomment-423962634).

Coins are derived from bits of `random_value`.

There is a pair PR 62 in `threshold_crypto`.


vkomenda:
This has been [obsoleted](https://github.com/poanetwork/hbbft/issues/183#issuecomment-439947528).

Issue #333, opened 2018-11-12
Reorganize DHB and HB builders.


afck:
There is a lot of duplication between HB, DHB and their builders. They each have an `encryption_schedule`, `subset_handling_strategy`, `max_future_epochs`, `rng` and, with #332, `random_value`.

We need to find a solution that avoids the duplication but still makes usage of the builders convenient. (E.g. maybe the above fields—probably excluding `rng`—should be grouped in a `Params` struct?)

afck:
Unassigning this for now, but possibly the builders' individual setters should be removed, and the user should create a `Params` instead?

PR #345, opened 2018-11-19
Add a `Params` struct with HB parameters.

afck:
This removes some duplication between DHB, HB and their builders.

See #333.


afck:
Possibly we should include the full `Params` object in the `JoinPlan`; not sure.

PR #351, opened 2018-11-26
Extend the documentation, simplify errors.

afck:
This changes and adds several doc comments, adds the `missing_docs` lint and simplifies some of the error types.

### DIFF
--- a/src/binary_agreement/mod.rs
+++ b/src/binary_agreement/mod.rs
@@ -83,28 +83,31 @@ pub use self::sbv_broadcast::Message as SbvMessage;
 /// An Binary Agreement error.
 #[derive(Clone, Eq, PartialEq, Debug, Fail)]
 pub enum Error {
-    #[fail(display = "Error handling threshold sign message: {}", _0)]
+    /// Error handling a `ThresholdSign` message.
+    #[fail(display = "Error handling ThresholdSign message: {}", _0)]
     HandleThresholdSign(threshold_sign::Error),
+    /// Error invoking the common coin.
     #[fail(display = "Error invoking the common coin: {}", _0)]
     InvokeCoin(threshold_sign::Error),
-    // Strings because `io` and `bincode` errors lack `Eq` and `Clone`.
-    #[fail(display = "Error writing epoch for nonce: {}", _0)]
-    Io(String),
-    #[fail(display = "Error serializing session ID for nonce: {}", _0)]
+    // String because `io` and `bincode` errors lack `Eq` and `Clone`.
+    /// Error serializing the session ID for the common coin.
+    #[fail(display = "Error serializing session ID for coin: {}", _0)]
     Serialize(String),
 }
 
 impl From<bincode::Error> for Error {
     fn from(err: bincode::Error) -> Error {
-        Error::Io(format!("{:?}", err))
+        Error::Serialize(format!("{:?}", err))
     }
 }
 
 /// An Binary Agreement result.
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+/// A `BinaryAgreement` step, containing at most one output.
 pub type Step<N> = ::Step<Message, bool, N>;
 
+/// The content of a message belonging to a particular `BinaryAgreement` epoch.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub enum MessageContent {
     /// Synchronized Binary Value Broadcast message.
@@ -138,7 +141,9 @@ impl MessageContent {
 /// Messages sent during the Binary Agreement stage.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Rand)]
 pub struct Message {
+    /// The `BinaryAgreement` epoch this message belongs to.
     pub epoch: u64,
+    /// The message content for the `epoch`.
     pub content: MessageContent,
 }
 
--- a/src/broadcast/error.rs
+++ b/src/broadcast/error.rs
@@ -4,29 +4,32 @@ use reed_solomon_erasure as rse;
 /// A broadcast error.
 #[derive(Clone, PartialEq, Debug, Fail)]
 pub enum Error {
+    /// Failed to create a `ReedSolomon` instance.
     #[fail(display = "CodingNewReedSolomon error: {}", _0)]
     CodingNewReedSolomon(#[cause] rse::Error),
+    /// Failed to encode the value.
     #[fail(display = "CodingEncodeReedSolomon error: {}", _0)]
     CodingEncodeReedSolomon(#[cause] rse::Error),
+    /// Failed to reconstruct the value.
     #[fail(display = "CodingReconstructShardsReedSolomon error: {}", _0)]
     CodingReconstructShardsReedSolomon(#[cause] rse::Error),
+    /// Failed to reconstruct the value.
+    // TODO: This should be unreachable.
     #[fail(
         display = "CodingReconstructShardsTrivialReedSolomon error: {}",
         _0
     )]
     CodingReconstructShardsTrivialReedSolomon(#[cause] rse::Error),
+    /// Observers cannot propose a value.
     #[fail(display = "Instance cannot propose")]
     InstanceCannotPropose,
+    /// Multiple inputs received. Only a single value can be proposed.
     #[fail(display = "Multiple inputs received")]
     MultipleInputs,
-    #[fail(display = "Not implemented")]
-    NotImplemented,
+    /// Failed to construct a Merkle tree proof.
     #[fail(display = "Proof construction failed")]
     ProofConstructionFailed,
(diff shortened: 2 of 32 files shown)


src/binary_agreement/mod.rs
```diff
@@ -83,28 +83,31 @@ pub use self::sbv_broadcast::Message as SbvMessage;
...
 }
 
 impl From<bincode::Error> for Error {
     fn from(err: bincode::Error) -> Error {
-        Error::Io(format!("{:?}", err))
+        Error::Serialize(format!("{:?}", err))
     }
 }
 
 /// An Binary Agreement result.
```
andogro: ~~An~~ A

src/sync_key_gen.rs
```diff
@@ -190,14 +190,16 @@ use {NetworkInfo, NodeIdT};
...
 pub enum Error {
+    /// Error creating `SyncKeyGen`.
     #[fail(display = "Error creating SyncKeyGen: {}", _0)]
     Creation(CryptoError),
+    /// Error generating keys.
     #[fail(display = "Error generating keys: {}", _0)]
     Generation(CryptoError),
-    #[fail(display = "Error acknowledging part: {}", _0)]
-    Ack(CryptoError),
+    /// Unknown sender
```
vkomenda: Missing period.

src/threshold_decrypt.rs
```diff
@@ -25,15 +25,20 @@ use {DistAlgorithm, NetworkInfo, NodeIdT, Target};
 /// A threshold decryption error.
 #[derive(Clone, Eq, PartialEq, Debug, Fail)]
 pub enum Error {
+    /// Redundant input provided.
     #[fail(display = "Redundant input provided: {:?}", _0)]
     MultipleInputs(Box<Ciphertext>),
+    /// Invalid ciphertext
```
vkomenda: Missing period.

src/threshold_decrypt.rs
```diff
@@ -25,15 +25,20 @@ use {DistAlgorithm, NetworkInfo, NodeIdT, Target};
...
     #[fail(display = "Invalid ciphertext: {:?}", _0)]
     InvalidCiphertext(Box<Ciphertext>),
+    /// Unknown sender
```
vkomenda: Missing period.

src/threshold_decrypt.rs
```diff
@@ -25,15 +25,20 @@ use {DistAlgorithm, NetworkInfo, NodeIdT, Target};
...
     #[fail(display = "Unknown sender")]
     UnknownSender,
+    /// Decryption failed
```
vkomenda: Missing period.

src/threshold_sign.rs
```diff
@@ -31,33 +31,32 @@ use {DistAlgorithm, NetworkInfo, NodeIdT, Target};
...
     #[fail(display = "Redundant input provided")]
     MultipleMessagesToSign,
-    #[fail(display = "CombineAndVerifySigCrypto error: {}", _0)]
+    /// Error combining and verifying signature shares.
+    #[fail(
+        display = "Error combining and verifying signature shares: {}",
+        _0
+    )]
     CombineAndVerifySigCrypto(crypto::error::Error),
+    /// Unknown sender
```
vkomenda: Missing period.

src/broadcast/error.rs
```diff
@@ -4,29 +4,32 @@ use reed_solomon_erasure as rse;
...
-    #[fail(display = "Not implemented")]
-    NotImplemented,
+    /// Failed to construct a Merkle tree proof.
     #[fail(display = "Proof construction failed")]
     ProofConstructionFailed,
-    #[fail(display = "Root hash mismatch")]
-    RootHashMismatch,
-    #[fail(display = "Threading")]
-    Threading,
+    /// Unknown sender
```
vkomenda: Missing period.

src/lib.rs
```diff
@@ -6,34 +6,42 @@
...
+//! problem arises. The application will usually run on different nodes, connected to each other
+//! via a network. The nodes give an input to the algorithm, exchange several messages, and
+//! eventually the algorithm returns an output, which is guaranteed to be the same in each correct
+//! node.
 //!
-//! * Messages received from other nodes must be passed into the instance, and messages produced by
-//! the instance sent to corresponding nodes.
+//! However, the `hbbft` crate only implements the abstract protocols, not the networking. It is
+//! the application's responsibility to serialize, sign and send the messages to the other nodes.
+//! That is why in addition to methods for giving input, the algorithms also have a
```
andogro: 'methods for giving input' is a little confusing, are you referring to the above 'nodes give an input' statement. Are you talking about methods for providing input to the nodes, or receiving info from the nodes - or both?  Is this referencing the`handle_input` method?

src/lib.rs
```diff
@@ -6,34 +6,42 @@
...
+//! `handle_message` method. The application is required to call this for every (validly signed,
```
andogro: call this method for every valid (signed and deserialized) message received from a peer.

src/lib.rs
```diff
@@ -6,34 +6,42 @@
...
-//! of faulty nodes _f_ (including complete takeover by an attacker), as long as the total number
-//! _N_ of nodes is greater than _3 f_.
-//!
-//! The Honey Badger consensus algorithm is both Byzantine fault tolerant and asynchronous. It does
-//! not make timing assumptions about message delivery. An adversary can control network scheduling
-//! and delay messages without impacting consensus, and progress can be made in adverse networking
-//! conditions.
+//! A consensus algorithm is a protocol that helps a number of nodes agree on some data value.
+//! Byzantine fault tolerant systems can tolerate a number of faulty nodes _f_ (broken, or even
+//! controlled by an attacker), as long as the total number _N_ of nodes is greater than _3 f_.
```
c0gent: "the total number of nodes _N_"?

src/lib.rs
```diff
@@ -6,34 +6,42 @@
...
+//! node.
 //!
-//! * Messages received from other nodes must be passed into the instance, and messages produced by
-//! the instance sent to corresponding nodes.
+//! However, the `hbbft` crate only implements the abstract protocols, not the networking. It is
+//! the application's responsibility to serialize, sign and send the messages to the other nodes.
+//! That is why in addition to methods for giving input, the algorithms also have a
+//! `handle_message` method. The application is required to call this for every (validly signed,
+//! deserialized) message that was received from a peer.
+//! The methods return a [Step](struct.Step.html) which may contain messages, fault logs and outputs.
```
andogro: Did you mean 'the method returns'? or are you referring to the collective output i.e. 'These methods return'.

src/messaging.rs
```diff
@@ -1,18 +1,18 @@
...
     pub source: N,
+    /// The message's content.
     pub message: M,
 }
 
-/// Message destination can be either of the two:
-///
-/// 1) `All`: all remote nodes.
-///
-/// 2) `Node(id)`: remote node `id`.
```
c0gent: I'd probably simplify this to just: "The intended message recipient(s)." It's pretty clear by looking at the variants what the possibilities are and it's a bit of a duplication.

c0gent: This comment applies to the entire doc comments.

c0gent: I might comment...

src/lib.rs
afck: Yes, I'm very unsure about this part. The problem is that a lot of it depends on what the user actually wants to do with the hbbft library, and what we mean by "node"—only the algorithm instance or the whole running program. (Probably the latter, in general.) I'm also trying to talk about all our algorithms and not a specific ones, which adds to the vagueness. :grimacing: 
By "giving input" I mean that the higher-level application logic has somehow obtained a value that it now passes to the algorithm instance.
For example:
* The application is Parity, and it has picked some of the pending transactions. It passes these as "input" to its instance of the `HoneyBadger` algorithm.
* A distributed chat application uses `Broadcast` to make sure all messages reach everyone. The user just typed a line, and it passes that text as input to the `Broadcast` algorithm.

But I wouldn't give an example at that point in the docs yet; they should start with a general formulation, I think.

src/lib.rs
afck: I'm referring to both `handle_message` and `handle_input`, but also to algorithm-specific methods like `propose` and others.

src/lib.rs
afck: I'm a bit reluctant to refer to the whole message as "valid", since a faulty peer could have sent us a message with invalid content that has a valid signature and can be successfully deserialized. The validation of the message content happens inside our algorithms themselves, after the message has already been passed to `handle_message`.

afck:
I updated the crate root docs and used a much shorter text, with @mbr's help.
@andogro: I'm happy to make changes to that in another PR, so please let me know if the new version doesn't address your concerns.

PR #352, opened 2018-11-26
Remove peers from sender queue

### LINKED ISSUE #341: Remove peers from sender queue after validator change.
The sender queue already automatically adds peers based on `StateChange::InProgress`, so that joining nodes are guaranteed to receive their messages. It should also remove peers that have left the network:

`SenderQueue` probably needs a map that assigns to each peer the epoch after which they left, which would be updated in `update_epoch`. Then `defer_messages` would drop all messages from later epochs. When all outgoing messages to a removed peer have been sent, the peer can be removed from both maps.

vkomenda:
Closes #341 

Prior to this PR we had only addition of new peers in the sender queue. This PR allows removal of those peers which are validators that become superseded by other validators in the next epoch.


src/sender_queue/mod.rs
```diff
@@ -237,6 +283,20 @@ where
...
+                    removed_ids.push(id.clone());
+                }
+            } else {
+                removed_ids.push(id.clone());
+            }
+        }
+        for id in removed_ids {
+            self.peer_epochs.remove(&id);
+            self.last_epochs.remove(&id);
+        }
```
afck: Is `defer_messages` the right place to check that? I see two triggers for removing an entry from `peer_epochs`:
* Either we handle a batch in which the peer is removed and their epoch is already at least that. Then we can remove it immediately. (In `update_epoch`.)
* Or we receive an `EpochStarted` message with the epoch in `last_epochs`, send them their remaining messages and then remove them. (In `process_new_epoch`.)

src/sender_queue/mod.rs
```diff
@@ -284,13 +346,19 @@ where
         self
     }
 
+    pub fn last_epochs(mut self, last_epochs: BTreeMap<D::NodeId, D::Epoch>) -> Self {
+        self.last_epochs = last_epochs;
+        self
+    }
```
afck: Do we really need that in the builder? A joining node should receive all the required information in one of the following batches, shouldn't it? (Also unsure about `peer_epochs` and `outgoing_queue`.)

vkomenda: To be honest, I'm not sure we need the builder at all. But if we keep it in case the user wishes to replace the managed algorithm or change peer epochs (questionable), we could leave only `peer_epochs` out of the three - `outgoing_queue`, `peer_epochs` and `last_epochs` in the builder.

src/sender_queue/mod.rs
```diff
@@ -199,7 +199,11 @@ where
                     *e = epoch;
                 }
             }).or_insert(epoch);
-        self.process_new_epoch(sender_id, epoch)
+        if !self.remove_validator_if_old(sender_id) {
+            self.process_new_epoch(sender_id, epoch)
```
afck: I think we always need to call `process_new_epoch` first. Otherwise `remove_validator_if_old` might remove nonempty entries from `outgoing_queue`.

vkomenda: Or we can remove here when the received epoch is greater than the last epoch.

src/sender_queue/dynamic_honey_badger.rs
```diff
@@ -7,27 +7,40 @@ use rand::Rand;
...
 };
 
-impl<C, N> SenderQueueableOutput<N, Message<N>> for Batch<C, N>
+impl<C, N> SenderQueueableOutput<N, (u64, u64)> for Batch<C, N>
 where
     C: Contribution,
     N: NodeIdT + Rand,
 {
-    fn added_peers(&self) -> Vec<N> {
+    fn new_validators(&self) -> NewValidators<N> {
```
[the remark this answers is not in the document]
vkomenda: I've been trying to satisfy this but I think it's impossible for a managed algorithm to provide all participants because it does not know about observers. The new set of validators in `NewValidators::CompletedKeyGen` does not replace the old set of peers but only its subset of old validators which the managed algorithm has to specify because, in turn, the sender queue is not aware of the distinction between validators and observers.

I'd leave `NewValidators` for the reason above. If you know another way, let's continue trying to optimise this.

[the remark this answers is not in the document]
vkomenda: I'll do it a bit differently. Instead of one set, I'll return two: one containing the current participants and one containing the next participants, with an optimisation for the common case when both are the same.

src/sender_queue/mod.rs
vkomenda: Would you be happy with my current solution, that's waiting for `EpochStarted` from a first epoch after the last epoch of the old validator? Or would you like me to remove the old validator after all  locally produced messages for the last epoch have been sent to it?

Another possibility would be to remove in either of these two cases, whichever happens first.

vkomenda: Implemented the latter.

src/sender_queue/mod.rs
afck: I'd find it a bit more natural and safer to do it the other way round. Waiting for an `EpochStarted(43)` from a validator that was removed in epoch 42 kind of relies on the fact that we currently don't detect our own removal, and pretend we can continue as an observer. (Not sure, though; I'm fine with keeping it this way for now.)

src/sender_queue/mod.rs
```diff
@@ -220,9 +238,20 @@ where
...
+            if let Some((current_participants, next_participants)) = batch.participant_transition()
+            {
+                // Insert candidates.
+                for id in &next_participants {
+                    if id != self.our_id() {
+                        self.peer_epochs.entry(id.clone()).or_default();
+                    }
+                }
+                // Remove obsolete participants.
+                for id in current_participants.difference(&next_participants) {
```
afck: Could we instead take the difference of `peer_epochs.keys()` and `next_participants` (as sets, or using `filter`)? Then we wouldn't need `current_participants` at all, and would correctly handle the case of two subsequent `InProgress` after a failed key generation attempt.

src/sender_queue/mod.rs
```diff
@@ -248,6 +277,55 @@ where
...
+    ///
+    /// Returns `true` if the participant has been removed and `false` otherwise.
+    fn remove_participant(&mut self, id: &D::NodeId, last_epoch: &D::Epoch) -> bool {
+        let remove = if let Some(peer_epoch) = self.peer_epochs.get(id) {
+            last_epoch < peer_epoch
+                || (*last_epoch < self.algo.epoch() && self.outgoing_queue.get(id).map_or(
+                    false,
+                    |q| {
+                        q.iter()
+                            .take_while(|(epoch, _)| *epoch <= last_epoch)
```
afck: A bit shorter:

```rust
q.keys().take_while(|epoch| *epoch <= last_epoch)…
```

src/sender_queue/dynamic_honey_badger.rs
```diff
@@ -15,19 +16,38 @@ use dynamic_honey_badger::{
...
 };
 
-impl<C, N> SenderQueueableOutput<N, Message<N>> for Batch<C, N>
+impl<C, N> SenderQueueableOutput<N, (u64, u64)> for Batch<C, N>
 where
     C: Contribution,
     N: NodeIdT + Rand,
 {
-    fn added_peers(&self) -> Vec<N> {
+    fn participant_transition(&self) -> Option<(BTreeSet<N>, BTreeSet<N>)> {
```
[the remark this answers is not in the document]
vkomenda: Yes. The sender queue doesn't know the participants, i.e., non-observer peers.

src/sender_queue/mod.rs
vkomenda: No. See above. We should take the difference of current and next participants.

src/sender_queue/dynamic_honey_badger.rs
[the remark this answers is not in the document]
afck: Doesn't `peer_epochs` contain exactly the participants as its keys?

src/sender_queue/mod.rs
afck: Sorry, I might be a bit confused by all the sets… :dizzy_face: 
What I mean is: If `participant_transition` simply returned a single list, namely the IDs of _all_ future participants, i.e. all nodes that will be either a validator or a candidate in the next epoch, then I imagine that would suffice? Because then the difference of `peer_epochs.keys()` and `next_participants` would be exactly the peers that won't participate (either as candidates or as validators) in the next epoch anymore?

src/sender_queue/dynamic_honey_badger.rs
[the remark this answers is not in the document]
vkomenda: No, it also includes observers.

vkomenda:
> Although… this still doesn't handle the case of two subsequent InProgress changes, does it? Those would leave the failed candidates as zombie entries in the peer_epochs map?

I'm putting together a workaround. Let's see if we can optimise it.

tests/dynamic_honey_badger.rs
```diff
@@ -99,11 +102,65 @@ where
         if !input_add && network.nodes.values().all(has_remove) {
             network.input_all(Input::Change(Change::NodeChange(pub_keys_add.clone())));
             input_add = true;
+        } else if !rejoined_node0 {
+            if let Some(join_plan) = network
+                .nodes
+                .values()
+                .flat_map(|node| node.outputs())
+                .filter_map(|batch| match batch.change() {
```
afck: There seems to be an `Iterator::find_map` now. That would make the `next()` call unnecessary.

tests/dynamic_honey_badger.rs
```diff
@@ -99,11 +102,65 @@ where
...
+                    ChangeState::Complete(Change::NodeChange(pub_keys))
```
afck: Shouldn't we add the joining node already on `InProgress`?

vkomenda: Not sure. It's still failing...

vkomenda: Fixed it. There was an issue with a message from node 0 not being sent. Also I updated the batch verification function to allow the test pass when node 0 is missing some batches in the middle.

afck:
Still failling; but it makes sense that the outputs differ, of course, since the re-joined node didn't receive all the batches. So I guess it's just the test?

vkomenda:
I've just ran `drop_and_readd` 1000 times and it never failed!

I expected the QHB test to fail though because I'm still updating it. In addition to missed batches, node 0 does not send all of its user transactions, and the unmodified `verify_batches` will fail.

tests/net/mod.rs
```diff
@@ -994,20 +988,53 @@ where
     C: Contribution + Clone,
     N: NodeIdT,
 {
-    /// Verifies that all nodes' outputs agree, and returns the output.
+    /// Verifies that all nodes' outputs agree, and returns the output of node 1. Requirements
+    /// towards the output from node 0 are weaker compared to other nodes: it can be removed - hence
```
afck: Let's not hard-code into the test frameworks that node 0 was removed. I hope to add more complicated tests at some point, that add and remove more nodes.
Maybe we should instead pass in one node's index (or ID?) that is expected to have received all the batches. (A test can always make sure such a node exists, even if it's just an observer.)
We'd then compute the expected output of each node by filtering the complete output, and omitting all the batches where that node wasn't present, according to `SenderQueueableOutput::added_peers`?

vkomenda: That's right! I forgot to pass the node ID here...

tests/net/mod.rs
```diff
@@ -984,57 +984,40 @@ where
...
-                assert!(
-                    expected.iter().zip(node.outputs()).all(pub_eq),
-                    "Outputs of nodes {} and {} differ: {:?} != {:?}",
-                    EXPECTED_ID,
-                    i,
-                    expected,
-                    node.outputs()
-                );
+    /// Verifies that all nodes' outputs agree, given a correct "full" node that received all
+    /// batches. Validity of other nodes' output is judged on the grounds of their contributions
```
afck: That's filtering out too many batches: A node may have been too slow, or it may have been an observer or a candidate; then it won't have a contribution in the batch, but is still expected to output it.

tests/net/mod.rs
vkomenda: > Maybe we should instead pass in one node's index (or ID?) that is expected to have received all the batches. (A test can always make sure such a node exists, even if it's just an observer.) We'd then compute the expected output of each node by filtering the complete output, and omitting all the batches where that node wasn't present,

I implemented this, with a small difference that I pass the whole node by reference instead of by ID or index in the vector of correct nodes. I interpreted "node's presence" as whether the full node has a contribution for the node in question in a given epoch.

>  according to SenderQueueableOutput::added_peers?

I'm still thinking how better to add this to `verify_batches`. Or maybe in a separate check?

afck: Yes, I think it can't be done in a simple `filter` call. You'd need to essentially mimick what `SenderQueue` is doing, for every node:
* The first batch is expected. (This will also have to change at some point, for nodes that are added later.)
* Whenever `SenderQueueableOutput::added_peers` returns a new set that doesn't contain the node, all _following_ batches are not expected in the output.
* Whenever it returns a new set that _does_ contain the node, all following batches are expected again.

Not sure whether there's an easier way to do this. :confused:

vkomenda: My case for a separate check is this: if we check every batch against `SenderQueueableOutput::added_peers`, we would get early stage errors instead of divergence and an uninformative `MessageLimitExceeded`.

Waiting until all the batches are output is not helpful if the test diverges due to invalid batches.

afck: Right, that makes sense.
We should still test for both, though, and also assert that at the end, everyone got all the expected batches.

tests/queueing_honey_badger.rs
```diff
@@ -65,27 +65,91 @@ where
         if !has_remove(node) || !has_add(node) {
             return true;
         }
-        if node.outputs().iter().flat_map(Batch::iter).unique().count() < num_txs {
+        if node.id != NodeId(0)
+            && node.outputs().iter().flat_map(Batch::iter).unique().count() < num_txs
+        {
```
afck:
```rust
if A {
    return true;
}
false
```

Not your fault, but that can be simplified to just `A`.

tests/network/mod.rs
```diff
@@ -596,21 +601,43 @@ where
...
+                assert!(expected.len() >= node.outputs().len());
+                assert!(
+                    expected
+                        .iter()
+                        .zip(node.outputs())
+                        .take_while(pub_eq_ref)
+                        .next()
+                        .is_some(),
+                    "Outputs of nodes 0 and 1 don't start the same",
+                );
```
afck: That only asserts that the first batch is the same.

vkomenda: I know. That's not ideal. This check should resemble `VirtualNet::vefify_batches` more.

tests/net/mod.rs
afck: I think the right thing to do is something like this:

```rust
let mut participants = unimplemented!("all initial nodes?");
let mut expected = BTreeMap::new();
for batch in &full_node.outputs {
    for id in participants {
        expected.entry(*id).or_default().push(batch);
    }
    unimplemented!("update participants according to batch.change()");
}
for node in self.correct_nodes().filter(|n| n.id() != full_node.id()) {
    assert_eq!(node.outputs, expected[node.id()]);
}
```

But in the long run I guess `VirtualNet` should support actually adding and removing nodes to/from the network? Not sure. Adding @mbr for ideas.

tests/network/mod.rs
vkomenda: It turns out that the removed node still outputs a batch! Found after modifying `TestNetwork::verify_batches` to compute the same expectation as in `VirtualNet::verify_batches`. So, the test needs to be corrected again.

afck: I'd expect that it outputs all the blocks up to and including the one with the `Complete(…)` that removed it, and then again all blocks starting with the `InProgress(…)` that added it?

vkomenda: Yes, with the random scheduler. But with the first scheduler it outputs all batches, including the one made when it wasn't a participant.

afck: That sounds like a bug. :thinking:

vkomenda: > I'd expect that it outputs all the blocks up to and including the one with the Complete(…) that removed it, and then again all blocks starting with the InProgress(…) that added it?

All up to `Complete`, then skipping until and including `InProgress`, then all the rest? It isn't a participant in the epoch which produced the batch with `InProgress`.

afck: Ah, sorry, you're right, of course! The `InProgress` that adds it back is the last batch that it _doesn't_ receive. :+1:

src/sender_queue/mod.rs
```diff
@@ -248,10 +293,64 @@ where
...
+                    |q| {
+                        q.keys()
+                            .take_while(|epoch| *epoch <= last_epoch)
+                            .next()
+                            .is_none()
+                    },
+                ))
+        } else {
+            true
+        };
```
afck: That's reeeally hard to read, but I also can't come up with anything much better right now.
It _might_ help (_or_ make it worse) to `return false` in all the places where we'd assign `false` to `remove`, and get rid of that variable. The last line of the method would then just be `true`. Not sure.
But I'm also fine with merging it like that for now. Maybe we can reformulate it later.

tests/dynamic_honey_badger.rs
```diff
@@ -96,14 +99,70 @@ where
...
+    let secret_key = node0.instance().algo().netinfo().secret_key().clone();
+    let (dhb, dhb_step) = DynamicHoneyBadger::new_joining(
+        NodeId(0),
+        secret_key,
+        join_plan,
+        rand::thread_rng().gen::<Isaac64Rng>(),
+    ).expect("failed to reconstruct node 0");
+    let (sq, mut sq_step) = SenderQueue::builder(dhb, peer_ids.into_iter()).build(our_id);
+    *node0.instance_mut() = sq;
+    sq_step.extend(dhb_step.map(|output| output, Message::from));
```
afck: Is the `map` call needed at all? I think the two step types are now identical?

tests/net/mod.rs
```diff
@@ -990,24 +986,42 @@ where
...
+    {
+        let mut participants: BTreeSet<N> = self.nodes().map(Node::id).cloned().collect();
+        let mut expected: BTreeMap<N, Vec<_>> = BTreeMap::new();
+        for batch in &full_node.outputs {
+            for id in &participants {
+                expected.entry(id.clone()).or_default().push(batch);
+            }
+            if let Some(new_participants) = batch.participant_change() {
+                participants = new_participants;
+            }
```
afck: This wouldn't work with a pure observer node, I think?
If we don't need that right now, I'd add a `TODO` here: Ideally we'd need the same logic as in the `SenderQueue` implementation itself, where we remove the _difference_ of the previous batch's and the new participants, etc.

tests/net/mod.rs
```diff
@@ -990,24 +986,42 @@ where
...
+        }
+        for node in self.correct_nodes().filter(|n| n.id() != full_node.id()) {
```
afck: Actually, maybe let's not even filter here, so we get alerted it we passed in a non-full node by accident?

tests/net_dynamic_hb.rs
```diff
@@ -165,11 +165,51 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
+                } else if awaiting_addition.contains(&node_id) {
+                    // The node has removed the pivot node but hasn't added it back yet.
+                    pub_keys_rm.keys()
+                } else {
+                    // The node has added the pivot node back.
+                    pub_keys_add.keys()
+                }.collect();
+                assert!(
+                    batch.contributions().count() * 3 > expected_participants.len() * 2,
+                    "The batch contains less contributions than there are correct validators: {:?}",
```
afck: "less than N - f contributions"?

tests/net_dynamic_hb.rs
```diff
@@ -260,8 +319,43 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
+        .correct_nodes_mut()
+        .find(|node| *node.id() == pivot_node_id)
+        .expect("failed to get the pivot node");
+    let secret_key = pivot_node.algorithm().algo().netinfo().secret_key().clone();
+    let (dhb, dhb_step) =
+        DynamicHoneyBadger::new_joining(pivot_node_id, secret_key, join_plan, rng)
+            .expect("failed to reconstruct the pivot node");
+    let (sq, mut sq_step) = SenderQueue::builder(dhb, peer_ids.into_iter()).build(pivot_node_id);
+    *pivot_node.algorithm_mut() = sq;
+    sq_step.extend(dhb_step.map(|output| output, Message::from));
```
afck: As above, maybe the `map` isn't needed?

tests/network/mod.rs
```diff
@@ -59,8 +59,13 @@ impl<D: DistAlgorithm> TestNode<D> {
...
     }
 
+    /// Returns the internal algorithm's mutable instance.
+    #[allow(unused)] // Not used in all tests.
+    pub fn instance_mut(&mut self) -> &mut D {
+        &mut self.algo
+    }
     /// Creates a new test node with the given broadcast instance.
-    fn new((algo, step): (D, DaStep<D>)) -> TestNode<D> {
+    pub fn new((algo, step): (D, DaStep<D>)) -> TestNode<D> {
```
afck: Do `new` and `dispatch_messages` really need to be public now? They don't seem to be used when restarting node 0.

tests/network/mod.rs
```diff
@@ -596,21 +601,43 @@ where
     D: DistAlgorithm<Output = Batch<C, NodeId>, NodeId = NodeId>,
     C: Contribution + Clone,
 {
-    /// Verifies that all nodes' outputs agree.
+    /// Verifies that all nodes' outputs agree, noting that node 0 was removed - hence could have
+    /// missed some batches - and then rejoined the network.
```
afck: That's very specific to our tests. Maybe add a comment that this framework is deprecated and that's why we're not making it more general.

tests/network/mod.rs
afck: That assertion is still weaker than even this:
```rust
assert_eq!(expected[0], node.outputs()[0]);
```

tests/network/mod.rs
vkomenda: OK, unmaking `new` as `pub`.
`dispatch_messages` is used to dispatch the messages in the step after restarting node 0.

src/queueing_honey_badger/mod.rs
```diff
@@ -127,7 +143,10 @@ where
             queue: self.queue,
             rng: Box::new(rng),
         };
-        let step = qhb.propose()?;
+        let mut step = qhb.propose()?;
+        if let Some(dhb_step) = self.step {
+            step.extend_with(dhb_step, Message::from);
```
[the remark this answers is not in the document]
vkomenda: The types are subtly different: `DaStep<SenderQueue<QHB>>` and `DaStep<QHB>`. So, conversion is still required. (And I have to remind myself not to use `Step::extend_with` because it discards the output...)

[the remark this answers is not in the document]
afck: Right, we should probably make `extend_with` `#[must_use]`.

tests/network/mod.rs
```diff
@@ -631,12 +631,14 @@ where
...
+                node.outputs.iter().map(|batch| batch.epoch()).collect();
+            let expected_epochs: BTreeSet<_> =
+                expected[id].iter().map(|batch| batch.epoch()).collect();
             assert_eq!(
-                node.outputs.len(),
-                expected[id].len(),
-                "The output length of {:?} is incorrect: {:?}",
-                id,
-                node.outputs
+                expected_epochs, actual_epochs,
```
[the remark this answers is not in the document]
vkomenda: I made that because it looks better on screen. :)

tests/queueing_honey_badger.rs
```diff
@@ -69,11 +71,7 @@ where
     // Returns `true` if the node has not output all transactions yet.
     // If it has, and has advanced another epoch, it clears all messages for later epochs.
     let node_busy = |node: &mut TestNode<QHB>| {
-        if !has_remove(node) || !has_add(node) {
-            return true;
-        }
-        node.id != NodeId(0)
-            && node.outputs().iter().flat_map(Batch::iter).unique().count() < num_txs
+        !has_remove(node) || !has_add(node) || !node.instance().algo().queue().is_empty()
```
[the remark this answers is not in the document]
vkomenda: That's true. My idea was that the order in which checks are applied would make this situation impossible. I'll double-check. The rebase after the edition switch somehow introduced a new fault in `net_dynamic_hb`.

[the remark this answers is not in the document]
vkomenda: Also, the previous inequality is incorrect when the node is truly removed. It misses all batches while it is removed.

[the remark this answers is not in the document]
vkomenda: The second half of transactions is input into `node` before the `has_add(node)` becomes true. So, if the first half of transactions are accepted before `has_add(node)` becomes true, when it does become true, `node` will either have output the second half of transactions or will have some of those transactions left in the queue. Both cases are correct and the test should finish when all nodes output the second half of transactions.

tests/net_dynamic_hb.rs
vkomenda: It is needed as well.

vkomenda:
Rebased and ready for review. Comments welcome!

tests/net/mod.rs
vkomenda: Please check if this comment still applies. I'm not sure about the "pure observer". The same procedure works with a usual observer in `TestNetwork::verify_batches`.

tests/net/mod.rs
vkomenda: If there is another correct node with more batches, we will be alerted with the filter too! It happened to me a few times.

src/dynamic_honey_badger/mod.rs
```diff
@@ -144,7 +144,8 @@ pub struct JoinPlan<N: Ord> {
     change: ChangeState<N>,
     /// The current public key set for threshold cryptography.
     pub_key_set: PublicKeySet,
-    /// The public keys of the nodes taking part in key generation.
+    /// The public keys of the nodes taking part in key generation. Those nodes are the current
+    /// validators.
```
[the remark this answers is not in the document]
afck: Let's remove the first sentence. I think it's wrong, isn't it?

tests/net/mod.rs
afck: But does `TestNetwork::verify_batches` still work if you add comparing the _length_ of the observer's output to the expected one?

Here I'm just thinking of the case where one of `self.nodes()` is an observer from the beginning. It would get removed from `participants` here with the first batch that contains a change.
I guess we don't have that in the tests at the moment, but I'd maybe add a `TODO` here to account for observers.

tests/network/mod.rs
```diff
@@ -601,21 +608,59 @@ where
...
-                node.outputs()
+                node.outputs
+                    .iter()
+                    .zip(expected.get(id).expect("node is not expected"))
+                    .all(|(a, b)| a.public_eq(b)),
+                "Outputs of {:?} don't match the expectation",
+                id
             );
         }
+        assert!(
```
afck: Since `zip` doesn't fail if the lengths differ, we should assert that `self.observer.outputs.len()` equals `full_node.outputs.len()`.

tests/net/mod.rs
vkomenda: That observer would never be a participant. So it won't be removed here. The way `TestNetwork` handles observers is by explicitly adding those to the peer list of the sender queue.

PR #355, opened 2018-12-03
Added a test for expected participants when receiving a batch from a correct node

vkomenda:
We discussed a possibility of an immediate check of participants of a batch in https://github.com/poanetwork/hbbft/pull/352#discussion_r238253318. This patch adds such a check. It helps a lot with debugging `VitrualNet` tests updated to the latest sender queue, #352. Instead of divergence it gives an "incorrect batch" error.


tests/net_dynamic_hb.rs
```diff
@@ -169,7 +169,27 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
+                } else if awaiting_addition.contains(&node_id) {
+                    // The node has removed the pivot node but hasn't added it back yet.
+                    pub_keys_rm.keys()
+                } else {
+                    // The node has added the pivot node back.
+                    pub_keys_add.keys()
+                }.collect();
+                let batch_participants: Vec<_> = batch.contributions().map(|(id, _)| id).collect();
+                assert_eq!(
+                    batch_participants, expected_participants,
```
afck: That's not something we can expect. (If it's always true in the test, that's because the test doesn't simulate any difficult network conditions.)
The guarantee is just that at least _N - f_ participants managed to get their contribution accepted.

vkomenda: Should we expect only correct nodes to be batch participants in this case?

vkomenda: > But at this point in the code, some of them may still be waiting for it.

Right, this place in the code is better for making sanity checks on batches, possibly checking for participants that _shouldn't_ be in the batch.

The fact that all expected participants in a batch receive it is better checked at the end.

tests/net_dynamic_hb.rs
```diff
@@ -170,7 +170,7 @@ fn do_drop_and_readd(cfg: TestConfig) {
     loop {
         let (node_id, step) = net.crank_expect();
         if !net[node_id].is_faulty() {
-            // Verify that contributions from all participants of the batch are present.
+            // Verify that contributions from all correct participants of the batch are present.
```
afck: But we can't expect that either: Even correct participants may fail to get their contribution accepted, if their messages are delayed.

afck:
The check is correct now, but kind of weak. I think the original idea was to fail as soon as two nodes' outputs diverged, i.e. they output two different batches with the same number? More generally, the following points could be checked while the test is still running. Not sure if we want all of them, or which of them are sufficient to check at the end:

* [x] All accepted contributions are from validators (_not_ participants: candidates can't propose a contribution).
* [ ] The batch contains at least _N - f_ contributions.
* [ ] All nodes that have already output a batch with the same number have output the exact same batch.
* [ ] If this batch has number _k_ and the previous one had number _l_ and _k - l > 1_, then batch _l_ had a `Complete(NodeChange(…))` without this node, and batch _k_ has an `InProgress(NodeChange(…))` _with_ this node.

(If it keeps the code much simpler, I'd also be happy with keeping all of those checks at the end only, but you're right, of course, that that doesn't help if the test doesn't terminate.)

vkomenda:
Thanks for the checklist!

> I think the original idea was to fail as soon as two nodes' outputs diverged, i.e. they output two different batches with the same number?

My purpose initially was to fail if the removed node is still a participant in a batch of a correct node or if the added (same) node is not a participant. As you pointed out earlier, the latter check would not be correct because the added node may be lagging. So what remains of my original idea is the check that when a node is removed, it is no longer a participant in a batch.

afck:
Makes sense. And I also prefer to avoid unnecessarily complicating the test code; not sure how complicated the third point would be, but the fourth one is probably not worth it.
The second, however, should be easy: Just assert that the batch has the required number of contributions.

vkomenda:
> The batch contains at least N - f contributions.

Note that `N` is a moving target. When the test removes a node, `N` decreases from the point of view of all correct nodes.

vkomenda:
> The batch contains at least N - f contributions.

Could we change this to `N * 2 / 3` instead? Tests fail with `N - f`.

afck:
Yes, instead of "at least _N - f_" you can use "strictly more than ⅔ of the current validators"—excluding candidates.

tests/net_dynamic_hb.rs
```diff
@@ -169,7 +169,38 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
+                    pub_keys_add.keys()
+                } else if awaiting_addition.contains(&node_id) {
+                    // The node has removed the pivot node but hasn't added it back yet.
+                    pub_keys_rm.keys()
+                } else {
+                    // The node has added the pivot node back.
+                    pub_keys_add.keys()
+                }.collect();
+                // FIXME: Update the validators in VirtualNet when a validator is removed or added.
+                let correct_validator_num = expected_participants.len() * 2 / 3;
```
afck: …so it should probably be `expected_participants.len() * 2 / 3 + 1`.
Or `batch.contributions().count() * 3 > correct_validator_num * 2` below.

tests/net_dynamic_hb.rs
```diff
@@ -165,11 +165,50 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
+                } else if awaiting_addition.contains(&node_id) {
+                    // The node has removed the pivot node but hasn't added it back yet.
+                    pub_keys_rm.keys()
+                } else {
+                    // The node has added the pivot node back.
+                    pub_keys_add.keys()
+                }.collect();
+                assert!(
+                    batch.contributions().count() * 3 > expected_participants.len() * 2,
+                    "The batch contains less contributions than there are correct validators: {:?}",
```
afck: "less than N - f contributions"

Issue #356, opened 2018-12-05
The default RNG of Builders isn't cryptographically secure


sgeisler:
`DynamicHoneyBadgerBuilder` and `HoneyBadgerBuilder` use `ThreadRng` as their default RNG which isn't cryptographically secure according to the [`rand 0.4.2` docs](https://docs.rs/rand/0.4.2/rand/#cryptographic-security). Updating to a more recent version of `rand` would probably fix this problem.

The scariest part of the `0.4.2` `ThreadRng` is:
```
/// After generating a certain amount of randomness, the RNG will reseed itself
/// from the operating system or, if the operating system RNG returns an error,
/// a seed based on the current system time.
```

mbr:
This bit is a bit tricky; I believe we used to use `OsRng` before, the `thread_rng`s crept back in when adding in builders. Checking the impact, `rand` 0.4.5 (the highest version we can use that is not 0.5), any `thread_rng` is a [reseeding](https://github.com/rust-random/rand/blob/0.4/src/lib.rs#L879) RNG, using a [`StdRng`](https://github.com/rust-random/rand/blob/0.4/src/lib.rs#L882) as the actual RNG, which uses an [IsaacWord](https://github.com/rust-random/rand/blob/0.4/src/lib.rs#L772) RNG on every platform. Certainly [not ideal](https://en.wikipedia.org/wiki/ISAAC_(cipher)#Cryptanalysis).

The scary part is indeed scary, luckily on Linux, I am not aware of any actual circumstance where `/dev/urandom` stops working.

Historically, all cryptography (e.g. `threshold_crypto`) used their own `OsRng` instances, which defaulted to `OsRng`. In production use we usually explicitly set the RNG though. As opposed to the externally audited `threshold_crypto` crate, `hbbft` has not been published on crates.io yet.

Regarding newer versions of `rand`, we did try to migrate to `rand` 0.5 when that was the latest version, specifically to be able to use the [`CryptoRng`](https://docs.rs/rand/0.5.5/rand/trait.CryptoRng.html) trait, but this failed because the underlying [`pairing`](https://github.com/zkcrypto/pairing) crate has not kept up yet and we have not been able to get anything merged into it yet (@afck has an [outstanding PR](https://github.com/zkcrypto/pairing/pull/87), for example). In their defense, I don't know any crate that has breaking changes at a rate as high as the `rand` crate though, evidenced by the fact that as of this writing, it's already on 0.6 =).

[...]

PR #357, opened 2018-12-05
OsRng / external RNG Refactoring

mbr:
This is a WIP PR, as the tests won't pass (I have not updated them yet). I want feedback on these changes first, before I undertake updating and refactoring all the tests.

Here's the issue: Someone noticed that the default RNG for both `HoneyBadger` and `DynamicHoneyBadger` was the suboptimal `thread_rng()` (see Issue #356).

We cannot simply put in an `OsRng` and call it a day, since Hydrabadger uses tokio, which requires all of our algorithms to be send and sync. Our boxed RNG model has to go (I mentioned this a month back, but we kinda put off this refactoring until now, since it was not apparent what the gain was). There already was a plan to make all RNGs to be passed in from the outside, now we just had to accelerate that schedule.

The first commit shows what does not work, namely just plopping in `OsRng` - which is not `Send+Sync`. It still changes all the tests to use `OsRng`, because why not (makes grepping for `thread_rng` really convenient).

The second commit takes care of adding an extra `rng: &mut R` parameter to all relevant methods but cuts off at the boundary to `DistAlgorithm`.

The third commit takes the final step and changes the `DistAlgorithm` interface.

One open question is whether the change in `epoch_state.rs:78` is allowed; the function used to call `handle_input`, but now calls `propose` instead, which works out the same, but removes a layer of abstraction. The rationale behind this is to avoid introducing a chain of completely unnecessary `R: Rng` parameters in that area. Any feedback @afck ?

The PR is rather long and messy, it's a fairly unpleasant refactoring. Still, I am grateful for early feedback.

### DIFF
--- a/src/dynamic_honey_badger/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger/dynamic_honey_badger.rs
@@ -53,22 +53,22 @@ where
     type Message = Message<N>;
     type Error = Error;
 
-    fn handle_input<R: Rng>(&mut self, rng: &mut R, input: Self::Input) -> Result<Step<C, N>> {
+    fn handle_input<R: Rng>(&mut self, input: Self::Input, rng: &mut R) -> Result<Step<C, N>> {
         // User contributions are forwarded to `HoneyBadger` right away. Votes are signed and
         // broadcast.
         match input {
-            Input::User(contrib) => self.propose(rng, contrib),
+            Input::User(contrib) => self.propose(contrib, rng),
             Input::Change(change) => self.vote_for(change),
         }
     }
 
     fn handle_message<R: Rng>(
         &mut self,
-        rng: &mut R,
         sender_id: &Self::NodeId,
         msg: Self::Message,
+        rng: &mut R,
     ) -> Result<Step<C, N>> {
-        self.handle_message(rng, sender_id, msg)
+        self.handle_message(sender_id, msg, rng)
     }
 
     fn terminated(&self) -> bool {
@@ -92,10 +92,10 @@ where
 
     /// Creates a new `DynamicHoneyBadger` ready to join the network specified in the `JoinPlan`.
     pub fn new_joining<R: Rng>(
-        rng: &mut R,
         our_id: N,
         secret_key: SecretKey,
         join_plan: JoinPlan<N>,
+        rng: &mut R,
     ) -> Result<(Self, Step<C, N>)> {
         let netinfo = NetworkInfo::new(
             our_id,
@@ -121,7 +121,7 @@ where
         };
         let step = match join_plan.change {
             ChangeState::InProgress(ref change) => match change {
-                Change::NodeChange(change) => dhb.update_key_gen(rng, join_plan.era, change)?,
+                Change::NodeChange(change) => dhb.update_key_gen(join_plan.era, change, rng)?,
                 _ => Step::default(),
             },
             ChangeState::None | ChangeState::Complete(..) => Step::default(),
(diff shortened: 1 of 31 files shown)


src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -150,12 +148,15 @@ where
...
-                votes: self.vote_counter.pending_votes().cloned().collect(),
-            }).map_err(Error::ProposeHoneyBadger)?;
-        self.process_output(step)
+            .propose(
+                rng,
+                &InternalContrib {
+                    contrib,
+                    key_gen_messages,
+                    votes: self.vote_counter.pending_votes().cloned().collect(),
+                },
```
afck: I'd construct the `InternalContrib` in a separate step now, to avoid too much nesting.

src/sync_key_gen.rs
```diff
@@ -65,8 +65,8 @@
 //! use threshold_crypto::{PublicKey, SecretKey, SignatureShare};
 //! use hbbft::sync_key_gen::{AckOutcome, PartOutcome, SyncKeyGen};
 //!
-//! // Use a default random number generator for any randomness:
-//! let mut rng = rand::thread_rng();
+//! // Use the os random number generator for any randomness:
```
vkomenda: os -> OS

src/traits.rs
```diff
@@ -249,6 +250,10 @@ where
 }
 
 /// A distributed algorithm that defines a message flow.
+///
+/// Many algorithms require a source of random data, which must be supplied on each call. It is up
```
vkomenda: "source of random data" could be just "RNG" for definiteness.

mbr:
> I think I'd prefer making rng the last parameter everywhere instead of the first. 

What did I ever do to you? :P

Jokes aside, I made it the first, because it's the "correct" order from a functional programming point of view - imagine if you were able to curry these functions, I believe you'd find more use in being able to define a new function with a fixed RNG and calling propose multiple times than the other way around.

A good indicator is also that the `rng` argument used to be inside `self`, now it's as close as possible to `self`, but not quite contained!

afck:
> What did I ever do to you? :P

Yes, I only dared to make that suggestion because I'm 500 km away from you. :grin: 
I tend to think that readability (having the most "important" arguments first) weighs more than the currying argument, but I'm fine with either. :+1:

vkomenda:
> if you were able to curry these functions, I believe you'd find more use in being able to define a new function with a fixed RNG

That would have been an absolute hit with me. Indeed, passing the same RNG is now a routine operation. It's not like passing a message since the message gets constructed or deconstructed in the process.

src/traits.rs
mbr: :+1:

mbr:
> Looks great so far. It takes some time getting used to QueueingHoneyBadger::apply with the added rng argument. But after considering inferior alternatives such as cloning the RNG, your version is probably the best.

I briefly considered having two `apply` functions, or just not using apply when I need the RNG. Both are valid options as well, this one was just the fastest to implement whilst not seeming worse than the other two.

mbr:
I'll change the argument order due to majority opinion, but I'd like to formally note that I much prefer the other option. I have also not forgotten that one time you made me paint the bike shed yellow, when red was *obviously* a much better color.

mbr:
I've amend this for the hopefully final version. The test cases are now fixed, and every instance of `thread_rng` is removed from new-style tests (I had to amend the `binary_agreement_*` tests, which now leverage `TestRng::from_seed` and `proptest`, as a result we are testing 40 more instances each run.

Some other cleanup resulted in `fmt::Debug` implementations now going away (due to boxed `Rng` fields being removed, they can now be `#[derive]`d again.)

mbr:
Some `impl` blocks can now be rearranged and merged, as the trait requirements suddenly got stricter. I'm omitting this, it makes the PR much harder to review.

src/traits.rs
```diff
@@ -249,6 +250,9 @@ where
 }
 
 /// A distributed algorithm that defines a message flow.
+///
+/// Many algorithms require an RNG which must be supplied on each call. It is up to the caller to
+/// ensure that these source supply random numbers suitable for cryptographic applications.
```
afck: "this source supplies" or "these sources supply"?

vkomenda: I believe "to ensure that this RNG is cryptographically secure".

PR #361, opened 2018-12-10
Update the crossbeam dependency to 0.5

vkomenda:
This solves the vulnerability lint which affects the previously used version 0.3.2 of `crossbeam`. The changes only affect the old test.

### DIFF
--- a/examples/network/commst.rs
+++ b/examples/network/commst.rs
@@ -12,17 +12,6 @@ use serde::{de::DeserializeOwned, Serialize};
 
 use hbbft::SourcedMessage;
 
-#[derive(Debug)]
-pub enum Error {
-    IoError(io::Error),
-}
-
-impl From<io::Error> for Error {
-    fn from(err: io::Error) -> Error {
-        Error::IoError(err)
-    }
-}
-
 /// A communication task connects a remote node to the thread that manages the
 /// consensus algorithm.
 pub struct CommsTask<'a, M: 'a> {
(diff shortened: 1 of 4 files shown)


examples/network/commst.rs
```diff
@@ -59,16 +48,19 @@ impl<'a, M: Serialize + DeserializeOwned + Send + 'a> CommsTask<'a, M> {
...
-    pub fn run(mut self) -> Result<(), Error> {
+    pub fn run(mut self) -> Result<(), Box<dyn std::any::Any + Send + 'static>> {
         // Borrow parts of `self` before entering the thread binding scope.
         let tx = self.tx;
         let rx = self.rx;
-        let mut stream1 = self.stream.try_clone()?;
+        let mut stream1 = match self.stream.try_clone() {
+            Ok(stream) => stream,
+            Err(e) => return Err(Box::new(e)),
+        };
```
afck: Would this work, too? (Or even without the `map_err` because `Box` implements `From`?)
```rust
let mut stream1 = match self.stream.try_clone().map_err(Box::new)?;
```

examples/network/messaging.rs
```diff
@@ -85,9 +85,12 @@ impl<M: Send> Messaging<M> {
     }
 
     /// Spawns the message delivery thread in a given thread scope.
-    pub fn spawn<'a>(&self, scope: &Scope<'a>) -> ScopedJoinHandle<Result<(), Error>>
+    pub fn spawn<'a, 'scope>(
```
afck: If those lifetimes are the same (or even elided) it doesn't work?

examples/network/messaging.rs
```diff
@@ -97,46 +100,50 @@ impl<M: Send> Messaging<M> {
         let stop_rx = self.stop_rx.to_owned();
         let mut stop = false;
 
-        // TODO: `select_loop!` seems to really confuse Clippy.
+        // TODO: `select!` seems to really confuse Clippy.
```
afck: Does it still? Or can these `allow`s be removed now?

examples/network/messaging.rs
```diff
@@ -97,46 +100,50 @@ impl<M: Send> Messaging<M> {
...
-                            },
-                            Target::Node(i) => {
-                                result = if i < txs_to_comms.len() {
-                                    txs_to_comms[i].send(tm.message)
-                                        .map_err(Error::from)
-                                } else {
-                                    Err(Error::NoSuchTarget)
-                                };
+                select! {
+                    recv(rx_from_algo) -> tm => {
```
afck: Does the macro allow you to pattern match here? I.e. could you say `recv(rx_from_algo) -> Ok(tm)`?

examples/network/commst.rs
vkomenda: I tried those before and it didn't work. Pattern-matching is the best I could come up with.

examples/network/messaging.rs
vkomenda: Let's see. `Scope` has a lifetime parameter, so `'a` is required. Then `ScopedJoinHandle` also now has a lifetime parameter which cannot be the same as `'a` or it won't compile. I think both parameters are unavoidable as long as the function takes a `Scope` and returns a `ScopedJoinHandle`.

examples/network/messaging.rs
vkomenda: Ingenious! But that doesn't work.

PR #362, opened 2018-12-10
Remove unwrap and expect calls from production code.

afck:
Also make `Step::extend_with` `#[must_use]`, so the user doesn't forget to handle the output.

### DIFF
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -157,9 +157,7 @@ impl<N: NodeIdT> Broadcast<N> {
         let mut shards: Vec<&mut [u8]> = shards_iter.collect();
 
         // Construct the parity chunks/shards
-        self.coding
-            .encode(&mut shards)
-            .expect("incorrect shard size or number");
+        self.coding.encode(&mut shards)?;
 
         debug!(
             "{}: Value: {} bytes, {} per shard. Shards: {:0.10}",
--- a/src/dynamic_honey_badger/votes.rs
+++ b/src/dynamic_honey_badger/votes.rs
@@ -55,8 +55,8 @@ where
             voter: voter.clone(),
             sig: self.netinfo.secret_key().sign(ser_vote),
         };
-        self.pending.insert(voter.clone(), signed_vote);
-        Ok(self.pending.get(&voter).expect("entry was just inserted"))
+        self.pending.remove(&voter);
+        Ok(self.pending.entry(voter).or_insert(signed_vote))
     }
 
     /// Inserts a pending vote into the buffer, if it has a higher number than the existing one.
--- a/src/queueing_honey_badger/mod.rs
+++ b/src/queueing_honey_badger/mod.rs
@@ -117,10 +117,7 @@ where
     }
 
     /// Creates a new Queueing Honey Badger instance with an empty buffer.
-    pub fn build<R: Rng>(self, rng: &mut R) -> Result<QueueingHoneyBadgerWithStep<T, N, Q>>
-    where
-        R: 'static + Rng + Send + Sync,
-    {
+    pub fn build<R: Rng>(self, rng: &mut R) -> Result<QueueingHoneyBadgerWithStep<T, N, Q>> {
         self.build_with_transactions(None, rng)
     }
 
--- a/src/queueing_honey_badger/mod.rs
+++ b/src/queueing_honey_badger/mod.rs
@@ -117,9 +117,11 @@ where
     }
 
     /// Creates a new Queueing Honey Badger instance with an empty buffer.
-    pub fn build<R: Rng>(self, rng: &mut R) -> QueueingHoneyBadgerWithStep<T, N, Q> {
+    pub fn build<R: Rng>(self, rng: &mut R) -> Result<QueueingHoneyBadgerWithStep<T, N, Q>>
+    where
+        R: 'static + Rng + Send + Sync,
+    {
         self.build_with_transactions(None, rng)
-            .expect("building without transactions cannot fail")
     }
 
     /// Returns a new Queueing Honey Badger instance that starts with the given transactions in its
--- a/src/threshold_decrypt.rs
+++ b/src/threshold_decrypt.rs
@@ -58,7 +58,7 @@ pub struct ThresholdDecrypt<N> {
     /// The encrypted data.
     ciphertext: Option<Ciphertext>,
     /// All received threshold decryption shares.
-    shares: BTreeMap<N, DecryptionShare>,
+    shares: BTreeMap<N, (usize, DecryptionShare)>,
     /// Whether we already sent our shares.
     had_input: bool,
     /// Whether we have already returned the output.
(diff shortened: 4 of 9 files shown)


src/queueing_honey_badger/mod.rs
```diff
@@ -101,12 +101,11 @@ where
     }
 
     /// Creates a new Queueing Honey Badger instance with an empty buffer.
-    pub fn build<R>(self, rng: R) -> QueueingHoneyBadgerWithStep<T, N, Q>
+    pub fn build<R>(self, rng: R) -> Result<QueueingHoneyBadgerWithStep<T, N, Q>>
     where
         R: 'static + Rng + Send + Sync,
     {
         self.build_with_transactions(None, rng)
-            .expect("building without transactions cannot fail")
```
c0gent: If it's true that this cannot fail it would be better to leave this `expect` here. There's no point in having a runtime error for something that cannot occur at runtime (and is not a consequence of programmer error).

I'm guessing there's more to the story though.

src/threshold_decrypt.rs
```diff
@@ -182,7 +187,7 @@ impl<N: NodeIdT> ThresholdDecrypt<N> {
         let faulty_senders: Vec<N> = self
             .shares
             .iter()
-            .filter(|(id, share)| !self.is_share_valid(id, share))
+            .filter(|(id, (_, share))| !self.is_share_valid(id, share))
```
vkomenda: This would pass the same pair of ID and share with different node indices. Would that still be correct? Shouldn't we extend `is_share_valid` to handle node indices?

src/threshold_sign.rs
```diff
@@ -181,7 +182,7 @@ impl<N: NodeIdT> ThresholdSign<N> {
         let faulty_senders: Vec<N> = self
             .received_shares
             .iter()
-            .filter(|(id, share)| !self.is_share_valid(id, share))
+            .filter(|(id, (_, ref share))| !self.is_share_valid(id, share))
```
vkomenda: Same remark about `is_share_valid` as above.

src/threshold_decrypt.rs
afck: Would it? Maybe I missed something, but my intention was to just compute the index that matches the ID for each entry at the point where we put them into the map instead of here, to avoid the `unwrap`.
For each entry, the `idx` should be exactly the result of `netinfo.node_index(id)`, and `share` should be the one we received from `id`.

vkomenda: For example, if node 0 is removed, all other nodes' indices decrement by 1. If it's impossible to get the same share both with the old (before node 0 removal) and new (after removal) node indices, please ignore my remark.

afck: I'd say `ThresholdDecrypt` (and in general anything other than DHB and QHB) doesn't support changing sets of validators. The `NetworkInfo` is constant, and the node indices stay the same.

afck:
I rebased it, and thanks to non-lexical lifetimes and the RNG refactoring was finally able to reduce [proposing in the current epoch](https://github.com/poanetwork/hbbft/pull/362/files#diff-b665ce5dd9e05c0c0d24ef0dfff6f119R93) to a single line, like it should be. :tada:

src/broadcast/broadcast.rs
```diff
@@ -157,9 +157,7 @@ impl<N: NodeIdT> Broadcast<N> {
         let mut shards: Vec<&mut [u8]> = shards_iter.collect();
 
         // Construct the parity chunks/shards
-        self.coding
-            .encode(&mut shards)
-            .expect("incorrect shard size or number");
+        self.coding.encode(&mut shards)?;
```
mbr: Is this an error that can realistically occur, that we want to give the user ability to handle? Or is this just to satisfy the audit?

src/dynamic_honey_badger/votes.rs
```diff
@@ -55,8 +55,8 @@ where
             voter: voter.clone(),
             sig: self.netinfo.secret_key().sign(ser_vote),
         };
-        self.pending.insert(voter.clone(), signed_vote);
-        Ok(self.pending.get(&voter).expect("entry was just inserted"))
+        self.pending.remove(&voter);
+        Ok(self.pending.entry(voter).or_insert(signed_vote))
```
mbr: I think this is correct, unless someone can think of a use-case where it would be really helpful to notify the caller that he did try to insert things twice?

src/queueing_honey_badger/mod.rs
```diff
@@ -101,9 +101,11 @@ where
     }
 
     /// Creates a new Queueing Honey Badger instance with an empty buffer.
-    pub fn build<R: Rng>(self, rng: &mut R) -> QueueingHoneyBadgerWithStep<T, N, Q> {
+    pub fn build<R: Rng>(self, rng: &mut R) -> Result<QueueingHoneyBadgerWithStep<T, N, Q>>
+    where
+        R: 'static + Rng + Send + Sync,
```
mbr: This `where` clause is probably not necessary, right (looks like a merge-artifact). The `Rng` trait bound is already present, and `'static + Send + Sync` have been carefully removed.

tests/queueing_honey_badger.rs
```diff
@@ -93,9 +93,10 @@ fn new_queueing_hb(
     let mut rng = rand::thread_rng().gen::<Isaac64Rng>();
     let (qhb, qhb_step) = QueueingHoneyBadger::builder(dhb)
         .batch_size(3)
-        .build(&mut rng);
+        .build(&mut rng)
+        .expect("failed to build QueueingHoneyBadger");
```
mbr: Technically, we could use a result here, but it's a legacy test, so... =)

src/broadcast/broadcast.rs
afck: I _think_ it shouldn't be possible, looking at [the docs](https://docs.rs/reed-solomon-erasure/3.1.1/reed_solomon_erasure/enum.Error.html), but they're not 100% clear about this.
If we keep this as `expect`, we should probably remove all Reed-Solomon-related error variants.

src/queueing_honey_badger/mod.rs
afck: I amended the commit. (Then I force-pushed _again_ to rebase.)

tests/queueing_honey_badger.rs
mbr: Yeah, as if the test was a (traditional) main function.

src/dynamic_honey_badger/votes.rs
mbr: I guess this is fine then =).

src/broadcast/broadcast.rs
mbr: Well, since the user expects no Reed-Solomon-Errors for any arbitrary input they put in, we should not expose errors there to the user, right?

We should probably use `expect`s and `assert`s there that indicate any time they are a triggered, it is 100% certain to be a bug in either of the two libraries.

tests/queueing_honey_badger.rs
afck: Right; I don't think it's worth updating all the legacy tests for now?

mbr: Nope =).

mbr:
It's fine with me now, the remaining issues are not really blocking ones. I'll leave those up to you!

afck:
I agree with your issue with the Reed-Solomon errors, but I want to make sure I'm not introducing a new bug with that, so I'm now checking the reed_solomon_erasure code and adding a few more doc comments to explain the `expect` calls. I'll make a separate PR for those.

PR #364, opened 2018-12-12
Add docs and assert that 3 * f < N.

afck:


src/network_info.rs
```diff
@@ -8,15 +8,27 @@ use crate::{util, NodeIdT};
...
 #[derive(Debug, Clone)]
 pub struct NetworkInfo<N> {
+    /// This node's ID.
     our_id: N,
+    /// The number _N_ of nodes in the network. Equal to the size of `public_keys`.
     num_nodes: usize,
+    /// The number _f_ of faulty nodes that can be tolerated. Less than a third of _N_.
     num_faulty: usize,
+    /// The minimum number _N - f_ of correct nodes that are required.
+    num_correct: usize,
```
mbr: I am not a fan of this duplication of information, at all.

We should be saving the minimum, ensure things like `num_nodes` and `num_faulty` are immutable throughout, verify all assumptions through asserts in the `new` function, then add comments above every use that could possibly overflow in either direction, explaining/proving that this will never occur due to our checks made in `new`!

afck: I amended the commit:

```diff
@@ -14,8 +14,6 @@ pub struct NetworkInfo<N> {
     num_nodes: usize,
     /// The number _f_ of faulty nodes that can be tolerated. Less than a third of _N_.
     num_faulty: usize,
-    /// The minimum number _N - f_ of correct nodes that are required.
-    num_correct: usize,
     /// Whether this node is a validator. This is true if `public_keys` contains our own ID.
     is_validator: bool,
     /// This node's secret key share. Only validators have one.
@@ -52,7 +50,6 @@ impl<N: NodeIdT> NetworkInfo<N> {
         let num_nodes = public_keys.len();
         let num_faulty = util::max_faulty(num_nodes);
         assert!(3 * num_faulty < num_nodes, " 3 f >= N. This is a bug!");
-        let num_correct = num_nodes - num_faulty;
         let is_validator = public_keys.contains_key(&our_id);
         let node_indices: BTreeMap<N, usize> = public_keys
             .keys()
@@ -67,7 +64,6 @@ impl<N: NodeIdT> NetworkInfo<N> {
             our_id,
             num_nodes,
             num_faulty,
-            num_correct,
             is_validator,
             secret_key_share: secret_key_share.into(),
             secret_key,
@@ -107,7 +103,8 @@ impl<N: NodeIdT> NetworkInfo<N> {
     /// correct.
     #[inline]
     pub fn num_correct(&self) -> usize {
-        self.num_correct
+        // As asserted in `new`, `num_faulty` is never greater than `num_nodes`.
+        self.num_nodes - self.num_faulty
     }
```

PR #368, opened 2018-12-18
Migrate to rand 0.6.

afck:


examples/simulation.rs
```diff
@@ -299,13 +299,13 @@ where
             .values()
             .filter_map(TestNode::next_event_time)
             .min()?;
-        let min_ids: Vec<NodeId> = self
+        let next_id = self
             .nodes
             .iter()
             .filter(|(_, node)| node.next_event_time() == Some(min_time))
+            .choose(rng)
```
mbr: If I am reading the [rand source](https://docs.rs/rand/0.6.1/src/rand/seq/mod.rs.html#173-308) correctly, this will consume `n` random floats. This is only an example, but I assume that usually random numbers are more expensive than the space to store a few node ids and calling choose on the slice.

It reads much nicer though, so I am conflicted! Still worth it though to find out about the *P = 1/n* trick for successively selecting a random element.

src/binary_agreement/mod.rs
```diff
@@ -148,11 +149,11 @@ pub struct Message {
 }
 
 // NOTE: Extending rand_derive to correctly generate random values from boxes would make this
-// implementation obsolete; however at the time of this writing, `rand::Rand` is already deprecated
+// implementation obsolete; however at the time of this writing, `rand_derive` is already deprecated
 // with no replacement in sight.
-impl rand::Rand for MessageContent {
-    fn rand<R: rand::Rng>(rng: &mut R) -> Self {
-        let message_type = *rng.choose(&["sbvb", "conf", "term", "coin"]).unwrap();
+impl Distribution<MessageContent> for Standard {
```
mbr: These are slated for removal anyway, as the previous `rand::Rand` implementation was only used for random adversary testing. We could port the little bit of code that uses this (if it's still present) and remove these altogether. Maybe this is for a separate PR though.

src/dynamic_honey_badger/builder.rs
```diff
@@ -112,7 +108,7 @@ where
         let sk_set = SecretKeySet::random(0, rng);
         let pk_set = sk_set.public_keys();
         let sks = sk_set.secret_key_share(0);
-        let sk: SecretKey = rng.gen();
+        let sk = rng.gen::<SecretKey>();
```
mbr: With the new `distribution`s, should be make it explicit here, which distribution we are using, e.g. something like `Uniform`?

src/honey_badger/message.rs
```diff
@@ -1,16 +1,16 @@
 // `threshold_sign::Message` triggers this Clippy lint, but `Box<T>` doesn't implement `Rand`.
 #![allow(clippy::large_enum_variant)]
 
-use rand::Rand;
-use rand_derive::Rand;
+use rand::distributions::{Distribution, Standard};
```
mbr: Same question as above, should we be more explicit without distributions, i.e. preferring `Uniform` over `Standard`, because the former is definitely correct for choosing values, while the latter is not very descriptive?

tests/honey_badger.rs
```diff
@@ -133,16 +133,16 @@ where
...
             .nodes
             .iter()
             .filter(|(_, node)| !node.instance().algo().has_input())
             .map(|(id, _)| *id)
-            .collect();
-        if let Some(id) = rng.choose(&input_ids) {
-            let queue = queues.get_mut(id).unwrap();
-            queue.remove_multiple(network.nodes[id].outputs().iter().flat_map(Batch::iter));
-            network.input(*id, queue.choose(&mut rng, 3, 10));
+            .choose(&mut rng)
```
mbr: Same as above.

tests/net_util.proptest-regressions
```diff
@@ -1,7 +0,0 @@
-# Seeds for failure cases proptest has generated in the past. It is
```
mbr: Why did we kill the regressions file?

afck: By mistake. :grimacing: I'll restore it.
I'm not sure what this one is for, though: Was there a bug in `network_dimensions_shrink_and_grow`?

examples/simulation.rs
afck: Good point! Since it's only the simulation example I'm not too worried about the performance or entropy here, so I'd go with the nicer code?

src/binary_agreement/mod.rs
afck: I'm not sure what you mean: Don't we want to keep the random adversaries?

I guess it would be cleaner to make the random adversary put together the messages by hand, so these `impl`s don't pollute the production code. In a separate PR, though, yes.

src/dynamic_honey_badger/builder.rs
afck: I'd say no: `SecretKey` is not a type where any distributions other than a uniform one make any sense.

tests/honey_badger.rs
afck: But as above, it's just a test, and I don't think this is a performance bottleneck.

examples/simulation.rs
mbr: The issue with examples is that they, at some point, end up as production code, by virtue of someone being unfamiliar with the specifics using them as, well, an example worth emulating.

At worst, this is `n` additional syscalls. I am somewhat thinking that maybe the `.choose` method on iterators is intended for cases where iterator result are costly to store only.

I'll leave this up to you, I am fine with either.

src/binary_agreement/mod.rs
mbr: Yes, we do want the random adversaries, but we would use `proptest`'s facilities for this in the future, generating these messages outside the core library.

src/dynamic_honey_badger/builder.rs
mbr: So why not make it easy on auditors by making this explicit?

tests/net_util.proptest-regressions
mbr: Frankly, I don't know how to read these =). I could look it up, but for now, it probably won't hurt to leave it in?

src/dynamic_honey_badger/builder.rs
afck: It looks like `Uniform` is only meant for intervals:
[`pub fn new<B1, B2>(low: B1, high: B2) -> Uniform<X>`](https://docs.rs/rand/0.6.1/rand/distributions/struct.Uniform.html#method.new)
And `Fr` (the field of `SecretKey`) doesn't even implement `Ord`.

afck: Also, the rand documentation says:
> [Both `Standard` and `Uniform` are in some sense uniform distributions.](https://docs.rs/rand/0.6.1/rand/distributions/index.html#distribution-to-sample-from-a-uniform-range)

mbr: It's not even inclusive, so we cannot use some sort of `::MAX` here. Alright, I take back the mean things I said!

PR #371, opened 2018-12-27
Split FaultKind

### LINKED ISSUE #285: Split up `FaultKind`.

phahulin:
Largely based on https://github.com/poanetwork/hbbft/pull/343
Closes #285

### DIFF
--- a/examples/simulation.rs
+++ b/examples/simulation.rs
@@ -439,7 +439,8 @@ fn main() {
             .build_with_transactions(txs.clone(), rng)
             .expect("instantiate QueueingHoneyBadger");
         let (sq, mut step) = SenderQueue::builder(qhb, peer_ids.into_iter()).build(our_id);
-        assert!(step.extend_with(qhb_step, Message::from).is_empty());
+        let output = step.extend_with(qhb_step, |fault| fault, Message::from);
+        assert!(output.is_empty());
         (sq, step)
     };
 
--- a/src/dynamic_honey_badger/dynamic_honey_badger.rs
+++ b/src/dynamic_honey_badger/dynamic_honey_badger.rs
@@ -11,11 +11,11 @@ use serde::{de::DeserializeOwned, Serialize};
 
 use super::votes::{SignedVote, VoteCounter};
 use super::{
-    Batch, Change, ChangeState, DynamicHoneyBadgerBuilder, EncryptionSchedule, Error, Input,
-    InternalContrib, JoinPlan, KeyGenMessage, KeyGenState, Message, Params, Result,
+    Batch, Change, ChangeState, DynamicHoneyBadgerBuilder, EncryptionSchedule, Error, FaultKind,
+    Input, InternalContrib, JoinPlan, KeyGenMessage, KeyGenState, Message, Params, Result,
     SignedKeyGenMsg, Step,
 };
-use crate::fault_log::{Fault, FaultKind, FaultLog};
+use crate::fault_log::{Fault, FaultLog};
 use crate::honey_badger::{self, HoneyBadger, Message as HbMessage};
 
 use crate::sync_key_gen::{Ack, AckOutcome, Part, PartOutcome, SyncKeyGen};
@@ -52,6 +52,7 @@ where
     type Output = Batch<C, N>;
     type Message = Message<N>;
     type Error = Error;
+    type FaultKind = FaultKind;
 
     fn handle_input<R: Rng>(&mut self, input: Self::Input, rng: &mut R) -> Result<Step<C, N>> {
         // User contributions are forwarded to `HoneyBadger` right away. Votes are signed and
@@ -280,7 +281,7 @@ where
         sender_id: &N,
         kg_msg: KeyGenMessage,
         sig: Signature,
-    ) -> Result<FaultLog<N>> {
+    ) -> Result<FaultLog<N, FaultKind>> {
         if !self.verify_signature(sender_id, &sig, &kg_msg)? {
             let fault_kind = FaultKind::InvalidKeyGenMessageSignature;
             return Ok(Fault::new(sender_id.clone(), fault_kind).into());
@@ -311,7 +312,9 @@ where
         rng: &mut R,
     ) -> Result<Step<C, N>> {
         let mut step: Step<C, N> = Step::default();
-        let output = step.extend_with(hb_step, |hb_msg| Message::HoneyBadger(self.era, hb_msg));
+        let output = step.extend_with(hb_step, FaultKind::HbFault, |hb_msg| {
+            Message::HoneyBadger(self.era, hb_msg)
+        });
         for hb_batch in output {
             let batch_era = self.era;
             let batch_epoch = hb_batch.epoch + batch_era;
--- a/src/fault_log.rs
+++ b/src/fault_log.rs
@@ -1,135 +1,90 @@
 //! Functionality for logging faulty node behavior encountered by each
 //! algorithm.
 //!
-//! Each algorithm can propogate their faulty node logs upwards to a
-//! calling algorithm via `DistAlgorihm`'s `.handle_input()` and
-//! `.handle_message()` trait methods.
+//! Each algorithm can propogate their faulty node logs upwards to a calling algorithm via
+//! `DistAlgorihm`'s `.handle_input()` and `.handle_message()` trait methods.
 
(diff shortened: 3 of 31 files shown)


examples/simulation.rs
```diff
@@ -439,7 +439,9 @@ fn main() {
             .build_with_transactions(txs.clone(), rng)
             .expect("instantiate QueueingHoneyBadger");
         let (sq, mut step) = SenderQueue::builder(qhb, peer_ids.into_iter()).build(our_id);
-        assert!(step.extend_with(qhb_step, Message::from).is_empty());
+        assert!(step
+            .extend_with(qhb_step, |fault| fault, Message::from)
+            .is_empty());
```
afck: Might be more readable to just assert that both `step` and `qhb_step` are empty. No need to `extend_with`. (Also below.)

phahulin: `Step.extend_with` returns its argument's `output` field, in this case that would be `qhb_step.output`, so it seems that this assertion is equivalent to `assert!(qhb_step.output.is_empty())` not to `assert!(step.is_empty() && qhb_step.is_empty())`

Is that correct or did you mean something else?

afck: Ah, you're right, of course, sorry! :+1:

src/fault_log.rs
```diff
@@ -1,135 +1,78 @@
...
-    /// `BinaryAgreement` received a duplicate `Aux` message.
-    DuplicateAux,
-    /// `BinaryAgreement` received multiple `Conf` messages.
-    MultipleConf,
-    /// `BinaryAgreement` received multiple `Term` messages.
-    MultipleTerm,
-    /// `BinaryAgreement` received a message with an epoch too far ahead.
-    AgreementEpoch,
-}
+
```
afck: I think empty lines in doc comments should also start with `//!`.

src/subset/error.rs
```diff
@@ -27,3 +27,14 @@ pub enum Error {
 
 /// A subset result.
 pub type Result<T> = result::Result<T, Error>;
+
+/// Subset does not actually have any messages defined, so there's no real FaultKind to define here
```
afck: That documentation is a bit weird. Maybe something like: "Faults that can be detected in `Subset`."
It's clear from the set of variants that these can only originate in `Broadcast` or `BinaryAgreement`.

src/traits.rs
```diff
@@ -99,27 +109,42 @@ impl<M, O, N> Step<M, O, N> {
...
+    pub fn map<M2, O2, F2, FO, FF, FM>(
+        self,
+        f_out: FO,
+        f_fail2fail: FF,
+        f_msg: FM,
+    ) -> Step<M2, O2, N, F2>
     where
+        F2: Fail,
         FO: Fn(O) -> O2,
+        FF: Fn(Fault<N, F>) -> Fault<N, F2>,
```
afck: We should only have to specify an `Fn(F) -> F2` here: The node ID must never change in this conversion. (Also in the other methods.)

examples/simulation.rs
mbr: I think assertions should be side-effect free! If the `assert!` macro were to be turned off, the code would break!

If this was `debug_assert!`, it would definitely be wrong, I am not 100% sure about the `assert!` macro though - but the fact that I even have to think about/research it make me prefer to have all the actual work done outside the assert.

mbr: And yes, I realize that this was the case even before this PR, so the question is whether or not we want to change it now in one fell swoop.

src/fault_log.rs
```diff
@@ -1,135 +1,78 @@
...
 }
 
-impl<N> Fault<N> {
-    /// Creates a new fault, blaming `node_id` for the `kind`.
-    pub fn new(node_id: N, kind: FaultKind) -> Self {
+impl<N, F> Fault<N, F>
+where
+    F: Fail,
+{
+    /// Creates a new fault given a node ID and a fault description.
```
mbr: Minor, but I really liked the old docs because it made it abundantly clear what is going on (assigning blame for `kind` to `node_id`). Is there a reason to change it?

examples/simulation.rs
phahulin: `step` is defined as `mut` in this piece of code, and `step.extend_with(...)` actually changes it's state, then `step` is returned from the closure. So it seems that in order to preserve the current functionality intact, it would be necessary to call `extend_with` anyway. In such a case we can keep the assert as it is now.
If assertion is changed to something like `assert!(step.is_empty() && qhb_step.is_empty())` the ci indeed fails.

phahulin: So I suggest it's not changed (at least in this PR)

afck: I agree, we should assign this to a local `output` variable and _then_ assert that `output.is_empty()`. But I'm fine with postponing this change to another PR, too, if you prefer.

src/traits.rs
phahulin: Sorry, I don't quite know how to do this one. So the general idea is that `FF` should only work on `Fail` and return `Fail` without ability to touch node id, correct?

If I change the line here (and below it in `extend_with`) to `FF: Fn(F) -> F2`, it seems to break things in some other places (because of type mismatch). Is correcting them one by one the way to go or should I do this in a different way?

afck: Yes, I'm afraid they have to be fixed one by one, sorry. :grimacing: 
I imagine something like this:
* Add something like `fn map(self, FF) -> Fault<N, F2> where FF: Fn(F) -> F2` to `Fault<N, F>`, and similarly to `FaultLog`, so that below in line 128 you can just write `fault_log: self.fault_log.map(f_fail2fail)`—and actually I'd rename the latter to `f_fault`.
`Fault::map` would just return `Fault { node_id: self.node_id, kind: f_fault(self.kind) }`.
* Wherever the code currently says something like
```rust
step.extend_with(other_step, |flt| Fault::new(flt.node_id, f(flt.kind)), to_msg);
```
it should be changed to just:
```rust
step.extend_with(other_step, f, to_msg);
```

src/subset/proposal_state.rs
```diff
@@ -5,7 +5,7 @@ use super::subset::BaSessionId;
 use super::{Error, FaultKind, MessageContent, Result};
 use crate::binary_agreement;
 use crate::broadcast::{self, Broadcast};
-use crate::fault_log::Fault;
+//use crate::fault_log::Fault;
```
afck: Please remove this if it's not needed.

src/dynamic_honey_badger/dynamic_honey_badger.rs
```diff
@@ -312,9 +312,7 @@ where
         rng: &mut R,
     ) -> Result<Step<C, N>> {
         let mut step: Step<C, N> = Step::default();
-        let to_fault = |hb_fault: Fault<N, honey_badger::FaultKind>| {
-            Fault::new(hb_fault.node_id, FaultKind::HbFault(hb_fault.kind))
-        };
+        let to_fault = |hb_fault: honey_badger::FaultKind| FaultKind::HbFault(hb_fault);
```
afck: This could also be replaced with `FaultKind::HbFault`.

src/honey_badger/epoch_state.rs
```diff
@@ -306,9 +306,7 @@ where
     /// Checks whether the subset has output, and if it does, sends out our decryption shares.
     fn process_subset(&mut self, cs_step: CsStep<N>) -> Result<Step<C, N>> {
         let mut step = Step::default();
-        let to_fault = |cs_fault: Fault<N, cs::FaultKind>| {
-            Fault::new(cs_fault.node_id, FaultKind::SubsetFault(cs_fault.kind))
-        };
+        let to_fault = |cs_fault: cs::FaultKind| FaultKind::SubsetFault(cs_fault);
```
afck: …and this with `FaultKind::SubsetFault`.

PR #372, opened 2018-12-27
Fixes the net_dynamic_hb test

### LINKED ISSUE #369: Fix sender queue messaging when a validator is removed and then rejoined
The sender queue currently does not handle the edge case of removal and rejoining the same validator in the next epoch correctly. The bug is as follows. The removed validator announces the epoch in which it is no longer a validator. Other nodes start sending messages for that epoch. If we start a ballot for adding back the same node as validator in that next epoch, we can no longer correctly restart the removed validator with the join plan because it has already had processed some of the messages for that epoch.

With the current implementation of the sender queue this use case requires additional processing of messages, e.g., replaying those next epoch messages on the restarted node. There are better options.

Messages to the removed validator should be queued correctly on all other validators starting from the epoch when it is no longer a validator. If the removed validator is going to rejoin in that epoch, it must not send the `EpochStarted` message for the epoch when it is no longer a validator until it receives a `JoinPlan` out of band and restarts. Once restarted, it sends out the `EpochStarted` message and other validators send the queued messages to it.

This solution changes how removed validators are treated. They no longer become observers since they don't send any messages until they receive a `JoinPlan`.

If you think there are better solutions or that we simply have to keep the removed validator as an observer by default, let's discuss it in this issue.

vkomenda:
Fixes #369.

The fix mostly amounts to improving how removed nodes are handled in `VirtualNet`.

The sender queue algorithm now flags itself if it was removed. The user can check the flag `is_removed` to shut down the node.

### DIFF
--- a/src/sender_queue/dynamic_honey_badger.rs
+++ b/src/sender_queue/dynamic_honey_badger.rs
@@ -148,7 +148,7 @@ where
         let id = self.algo().netinfo().our_id().clone();
         let (dhb, dhb_step) =
             DynamicHoneyBadger::new_joining(id.clone(), secret_key, join_plan, rng)?;
-        let (sq, mut sq_step) = SenderQueue::builder(dhb, peer_ids.into_iter()).build(id);
+        let (sq, mut sq_step) = SenderQueue::builder(dhb, peer_ids).build(id);
         sq_step.extend(dhb_step.map(|output| output, Message::from));
         *self = sq;
         Ok(sq_step)
--- a/src/sender_queue/dynamic_honey_badger.rs
+++ b/src/sender_queue/dynamic_honey_badger.rs
@@ -8,12 +8,14 @@ use rand::Rng;
 use serde::{de::DeserializeOwned, Serialize};
 
 use super::{
-    SenderQueue, SenderQueueableDistAlgorithm, SenderQueueableMessage, SenderQueueableOutput,
+    Message, SenderQueue, SenderQueueableDistAlgorithm, SenderQueueableMessage,
+    SenderQueueableOutput, Step,
 };
 use crate::{Contribution, DaStep, NodeIdT};
 
 use crate::dynamic_honey_badger::{
-    Batch, Change, ChangeState, DynamicHoneyBadger, Error as DhbError, Message,
+    Batch, Change, ChangeState, DynamicHoneyBadger, Error as DhbError, JoinPlan,
+    Message as DhbMessage,
 };
 
 impl<C, N> SenderQueueableOutput<N, (u64, u64)> for Batch<C, N>
@@ -42,34 +44,34 @@ where
     }
 }
 
-impl<N: Ord> SenderQueueableMessage for Message<N> {
+impl<N: Ord> SenderQueueableMessage for DhbMessage<N> {
     type Epoch = (u64, u64);
 
     fn is_premature(&self, (them_era, them): (u64, u64), max_future_epochs: u64) -> bool {
         match *self {
-            Message::HoneyBadger(era, ref msg) => {
+            DhbMessage::HoneyBadger(era, ref msg) => {
                 era > them_era || (era == them_era && msg.epoch() > them + max_future_epochs)
             }
-            Message::KeyGen(era, _, _) => era > them_era,
-            Message::SignedVote(ref signed_vote) => signed_vote.era() > them_era,
+            DhbMessage::KeyGen(era, _, _) => era > them_era,
+            DhbMessage::SignedVote(ref signed_vote) => signed_vote.era() > them_era,
         }
     }
 
     fn is_obsolete(&self, (them_era, them): (u64, u64)) -> bool {
         match *self {
-            Message::HoneyBadger(era, ref msg) => {
+            DhbMessage::HoneyBadger(era, ref msg) => {
                 era < them_era || (era == them_era && msg.epoch() < them)
             }
-            Message::KeyGen(era, _, _) => era < them_era,
-            Message::SignedVote(ref signed_vote) => signed_vote.era() < them_era,
+            DhbMessage::KeyGen(era, _, _) => era < them_era,
+            DhbMessage::SignedVote(ref signed_vote) => signed_vote.era() < them_era,
         }
     }
 
     fn first_epoch(&self) -> (u64, u64) {
         match *self {
-            Message::HoneyBadger(era, ref msg) => (era, msg.epoch()),
-            Message::KeyGen(era, _, _) => (era, 0),
-            Message::SignedVote(ref signed_vote) => (signed_vote.era(), 0),
+            DhbMessage::HoneyBadger(era, ref msg) => (era, msg.epoch()),
+            DhbMessage::KeyGen(era, _, _) => (era, 0),
+            DhbMessage::SignedVote(ref signed_vote) => (signed_vote.era(), 0),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/sender_queue/mod.rs
```diff
@@ -306,49 +330,46 @@ where
...
             return false;
         }
-        if let Some(peer_epoch) = self.peer_epochs.get(id) {
-            if last_epoch >= peer_epoch {
-                return false;
-            }
-            if let Some(q) = self.outgoing_queue.get(id) {
-                if q.keys().any(|epoch| epoch <= last_epoch) {
+        if id == self.our_id() {
+            // TODO: Schedule our own shutdown?
```
afck: I think it's fine the way it is; the user can now drop the object or restart.

tests/net/err.rs
```diff
@@ -41,7 +41,12 @@ where
     /// The execution time limit has been reached or exceeded.
     TimeLimitHit(time::Duration),
     /// A `Fault` is encountered in a step of a `DistAlgorithm`.
-    Fault(Fault<D::NodeId>),
+    Fault {
+        /// The ID of the node that reported the fault.
+        reported_by_id: D::NodeId,
```
mbr: Could we name this `reported_by` and save precious 3 bytes? =)

tests/net/err.rs
```diff
@@ -93,9 +98,14 @@ where
...
             }
-            CrankError::Fault(fault) => {
-                write!(f, "Node {:?} is faulty: {:?}.", fault.node_id, fault.kind)
-            }
+            CrankError::Fault {
+                reported_by_id,
+                fault,
+            } => write!(
+                f,
+                "Node {:?} reported another node {:?} as faulty: {:?}.",
```
mbr: Is "another" always true by the way? Technically, a node could report itself as being faulty, which would be strange, but all the more reason to fail a test!

tests/net/err.rs
```diff
@@ -41,7 +41,12 @@ where
     /// The execution time limit has been reached or exceeded.
     TimeLimitHit(time::Duration),
     /// A `Fault` is encountered in a step of a `DistAlgorithm`.
```
mbr: We should add to the docs here that only faults reported by correct nodes are taken into account, e.g.

```rust
    /// A `Fault` was reported by a correct node in a step of a `DistAlgorithm`.
```

tests/net/err.rs
mbr: Maybe even mention that the reporting node was correct!

tests/net/mod.rs
```diff
@@ -914,7 +923,14 @@ where
         self.adversary = adv;
 
         // Step 1: Pick a message from the queue and deliver it; returns `None` if queue is empty.
-        let msg = self.messages.pop_front()?;
+        // We use a do-while loop to remove the prefix of messages addressed to removed nodes.
+        let mut msg;
+        while {
```
mbr: I appreciate the clever construction, however I do share Steve's concerns here: https://github.com/rust-lang/rfcs/issues/1313#issuecomment-146671501

Especially since we already have a comment explaining the whole construct. I think `loop { ...; cond(); }`-style might be better here?

tests/net_dynamic_hb.rs
```diff
@@ -121,7 +131,8 @@ fn do_drop_and_readd(cfg: TestConfig) {
 
     // We generate a list of transaction we want to propose, for each node. All nodes will propose
     // a number between 0..total_txs, chosen randomly.
-    let mut queues: collections::BTreeMap<_, Vec<usize>> = net
+    let mut queues: BTreeMap<_, Vec<usize>> = state
+        .get_net_mut()
```
afck: Why is `get_net_mut` required? (Also below.)

tests/net_dynamic_hb.rs
```diff
@@ -269,10 +277,25 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
-        if removed_ourselves {
-            // Further operations on the cranked node are not possible. Continue with processing
-            // other nodes.
-            continue;
+
+        let (era, hb_epoch) = state.get_net_mut()[node_id].algorithm().algo().epoch();
+        if node_id != pivot_node_id
+            && awaiting_addition_input.contains(&node_id)
+            && state.shutdown_epoch.is_some()
+            && era + hb_epoch == state.shutdown_epoch.unwrap()
```
afck: That restriction shouldn't be necessary, I think? We can vote for re-adding in any epoch.

src/sender_queue/dynamic_honey_badger.rs
```diff
@@ -124,4 +126,31 @@ where
...
+        &mut self,
+        join_plan: JoinPlan<N>,
+        peer_ids: I,
+        rng: &mut R,
+    ) -> Result<C, N>
+    where
+        I: Iterator<Item = N>,
+    {
+        if !self.is_removed {
+            // TODO: return an error?
```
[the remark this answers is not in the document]
vkomenda: `is_removed` is reset to `false` by this function as a result of calling `SenderQueueBuilder::build`. So, calling it twice in a row is an error.

At the moment the type of `SenderQueue` error is the same as the type of underlying `DistAlgorithm` error. It looks like adding an error variant in `dynamic_honey_badger::error::Error` to be used in the sender queue is not optimal. Possibly I have to introduce a wrapper for errors in `sender_queue`.

tests/net_dynamic_hb.rs
vkomenda: It fails without the validators waiting for this epoch. Is this incorrect, you think?

vkomenda:
I'm finished responding to comments. It appears some new functions in `net_dynamic_hb` weren't required. I refactored those out.

> I am a bit concerned over the filter that filters messages for removed nodes, it seems to remove almost all usefulness from the error that gets thrown when messages are sent to non-existing nodes.

Only messages to removed nodes are filtered out. If the node is not removed using the provided method `remove_node` and yet disappears, a message to it still triggers an error in `crank`.

tests/net_dynamic_hb.rs
```diff
@@ -278,22 +272,21 @@ fn do_drop_and_readd(cfg: TestConfig) {
...
+        let (era, hb_epoch) = state.net[node_id].algorithm().algo().epoch();
         if node_id != pivot_node_id
             && awaiting_addition_input.contains(&node_id)
             && state.shutdown_epoch.is_some()
             && era + hb_epoch == state.shutdown_epoch.unwrap()
         {
             // Now we can add the node again. Public keys will be reused.
-            let step = state
-                .get_net_mut()
+            let _ = state
```
afck: This is still a `Step`, isn't it? Let's `assert!` that its output is empty instead of dropping it right away.
(We don't actually expect the step to be completely empty, because it will contain a `SignedVote` message, right?)

tests/net/mod.rs
vkomenda: Now I started thinking that the messages should be removed earlier, at dispatch time. That way we can guarantee that no messages to the removed node arrive to it when it is finally readded. That would have been an undesirable side-effect.

src/sender_queue/dynamic_honey_badger.rs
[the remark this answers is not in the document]
mbr: That sounds like an unfortunate situation. The way you describe it, a proper error type for the sender queue is probably correct though.

afck:
I agree with all of @mbr's points. :heavy_plus_sign: 
But let's try to find a way to avoid turning this PR into a major `tests/net` extension, in addition to the test fix.

src/sender_queue/error.rs
```diff
@@ -0,0 +1,19 @@
...
+/// Sender queue error variants.
+#[derive(Debug, Fail)]
+pub enum Error<E>
+where
+    E: Debug + Fail,
+{
+    /// Failed to apply a function to the managed algorithm.
+    #[fail(display = "Function application failure: {}", _0)]
+    Apply(E),
+    /// Failed to restart `DynamicHoneyBadger` because it had not been removed.
```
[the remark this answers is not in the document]
afck: Isn't it already generic? I think at the moment, `E` doesn't necessarily have to be a `dynamic_honey_badger::Error`?

tests/net_dynamic_hb.rs
afck: Right, I'd just use `>=` then. :+1:

vkomenda:
These are all very thoughtful comments, thanks @mbr and @afck.

> I think it's problematic to add these automatic removal features to the simulated networking code, as this seems a bit like fixing the tests instead of fixing the code.

Fixing the node removal feature in `VirtualNet` is probably the most important part of this PR. Without it the test was struggling to fight off the ghost messages addressed to the removed node that were still sitting in the `VirtualNet` queue. I would stress that this feature based on a set of removed nodes is more selective than a flag might have been since it only allows `DynamicHoneyBadger` work with messages of the matching "lifetime", and does that without a major overhaul.

src/sender_queue/error.rs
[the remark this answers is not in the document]
vkomenda: Sorry, @afck, I don't understand why I cannot mention `DynamicHoneyBadger` in `Error::DynamicHoneyBadgerNotRemoved`. It is an error that is returned only from a `DynamicHoneyBadger` instance of the `SenderQueue`. This error is not "generic" in the sense that it cannot be used by a `HoneyBadger` sender queue. It's specific to `DynamicHoneyBadger`.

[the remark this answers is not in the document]
afck: Ah, sorry, you're right: I guess all variants except `Apply` are only used if `E` is `dynamic_honey_badger::Error`?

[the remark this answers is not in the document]
vkomenda: That's right!

Issue #347, opened 2018-11-19
Clear up confusion by introducing definitions in hbbft


mbr:
There has been a bit of confusion between observers, nodes, validators and some other terminology, especially since it is not 100% clearly defined. Our main issues occur on the boundary between algorithms and network transport, also the "abuse" of the term "observer" has been an issue.

This is an attempt to define these terms clearly. Any suggestions/improvements are welcome, also we need a place to stick these (ideally, the hbbft docs?).

Finally, these should be checked for compatibility with other PoA projects.

(see end for up-to-date version)

~~## Transport layer~~

~~The transport layer routes messages sent by *nodes*. Every *transport network* message has a destination, which is either a single node ID or `All`. Messages destined to `All` are sent to any known node by the sender, this resolution happens at the latest possible time when the message is sent.~~

~~* **Node**: Any instance of the program running, capable of receiving and sending messages. Every *node* is known to the *transport network* by an ID. Nodes may send and receive messages.~~

~~* **Peer**: A *node* in the *transport network* that is not the current one.~~

~~* **Observer**: A *node* that never sends messages. For this reason, it never influences the network (and is technically "invisible" to upper algorithm layers), but affects performance.~~

~~* **Transport network**: An fully interconnected network of *nodes*.~~

~~## Algorithm layer:~~

~~Any algorithm network runs on top of a *transport network*.~~

~~* **Participant**: A *node* that actively runs an algorithm. It sends and receives messages.~~

~~* **Auditor**: An *observer* that runs an instance of an algorithm passively, e.g. to verify its output. It has no influence on the *network*.~~

~~* **Network**: A complete graph of *participants* connected through a *transport network*.~~

~~* **Network size**: A *network's* number of *participants*.~~

~~* **Faulty participant**: A malicious or malfunctioning ("byzantine") *participant*.~~

~~## Specific to DHB~~

~~Every *node* ever actively involved in DHB is a *participant*. However, any participant can be in one of two states:~~

~~* **Candidate**: A participant seeking to be included in a future set of *validators*.~~

~~* **Validator**: An "incumbent", active validator that actively signs blocks.~~

~~The network sizes of DHB must not be confused with the network size of a contained Honey Badger instance inside DHB, as the size of DHB includes *candidates*.~~

c0gent:
[...]

PR #377, opened 2019-01-08
Rename DistAlgorithm to ConsensusProtocol.

afck:
See #347.


afck:
Good catch, thanks! That wasn't replaced because it had a typo. :sweat_smile:

Issue #374, opened 2019-01-03
Remove a random subset of validators in net_dynamic_hb


vkomenda:
Instead of a single validator (a.k.a. [pivot node](https://github.com/poanetwork/hbbft/blob/742ad7b83a3f613fd9a7b4ff88cfa82b10bd7dd3/tests/net_dynamic_hb.rs#L124)), the test `net_dynamic_hb` should pick a subset of validators at random for removal. This will better reflect the [interface](https://github.com/poanetwork/hbbft/blob/742ad7b83a3f613fd9a7b4ff88cfa82b10bd7dd3/src/dynamic_honey_badger/dynamic_honey_badger.rs#L169) to `DynamicHoneyBadger` that instead of changing one node in fact changes all nodes by substituting one set of validators for another.

vkomenda:
@RicoGit, thanks for taking this issue!

You are right about breaking the correctness invariant. I think that if we remove a subset of validators from a correct network and it becomes incorrect as a result then we can skip checking subsequent assertions because they may not hold with more than a third of validators being malicious.

However, for this test I would suggest always making sure that the network remains correct after removal of a subset of nodes. That is, if there is no such subset that you can remove without making the network incorrect, there is no reason to try readdition because the resulting incorrect network may in general not readd the nodes in that subset.

RicoGit:
Let me clarify. The test has to remove more than 1 validators, but cluster should remain correct.
 
For example is we have total N nodes and f malicious node, the test should remove less than N-(3f+1). But in the case of N=4 f=1, the test should remove 0 nodes - in that case, the test will check nothing.
Perhaps, we should improve `NetworkDimensionStrategy` that allow us to change N/f ratio? We should escape situations like N=4 f=1, N=7, f=2  and so on.

vkomenda:
[...]

PR #385, opened 2019-02-26
Remove a random subset of validators in net_dynamic_hb

RicoGit:
related issue #374 

`do_drop_and_re_add` chooses at random at least 1 node for removing from the network and then re-adding removed nodes back. Cluster always remain correct. In other words before and after removing validators network correctness condition (N=3f+1) is always satisfied.

Removed nodes can be as faulty nodes as well as correct ones. You can see in the log of the test how much is actually nodes will be removed.
```
Max number of nodes for removing is 8, was chosen 2 nodes
Will remove and re-add nodes {1, 2}
```


tests/net_dynamic_hb.rs
```diff
@@ -410,18 +407,24 @@ fn do_drop_and_readd(cfg: TestConfig) {
         }
     }
 
-    // As a final step, we verify that all nodes have arrived at the same conclusion. The pivot node
-    // can miss some batches while it was removed.
-    let full_node = state
+    // As a final step, we verify that all nodes have arrived at the same conclusion.
+    // Removed nodes can miss some batches while there were removed.
```
afck: Typo: "while they were removed"

tests/net_dynamic_hb.rs
```diff
@@ -455,10 +458,15 @@ where
...
     /// The test network.
     net: VirtualNet<DHB, A>,
-    /// The join plan for readding the pivot node.
+    /// The join plan for adding nodes.
     join_plan: Option<JoinPlan<usize>>,
-    /// The epoch in which the pivot node should go offline.
+    /// The epoch in which the removed nodes should go offline.
     shutdown_epoch: Option<u64>,
+    /// The removed nodes which are to be restarted as soon as all remaining
+    /// validators agree to add their back.
```
afck: Typo: "to add them back"

tests/net_dynamic_hb.rs
```diff
@@ -471,6 +479,73 @@ where
...
             join_plan: None,
             shutdown_epoch: None,
+            saved_nodes: Vec::new(),
+            old_subset_applied: false,
         }
     }
+
+    /// Selects random subset of validators which can be safely removed from the network
+    ///
+    /// The cluster always remain correct after removing this subset from the custer.
```
afck: Typo: "from the cluster".

tests/net_dynamic_hb.rs
```diff
@@ -471,6 +479,73 @@ where
...
+    /// This method may select as correct nodes as malicious.
```
afck: "may select correct nodes as well as malicious ones"

tests/net_dynamic_hb.rs
```diff
@@ -471,6 +479,73 @@ where
...
+        assert!(
+            n > f * 3,
+            "the network is already captured by the faulty nodes"
+        );
+
+        let max_number_correct_nodes_for_remove = n - (f * 3 + 1);
+        let remove_from_correct = rng.gen_range(0, max_number_correct_nodes_for_remove + 1);
+        let remove_from_faulty = if remove_from_correct == 0 && f > 0 {
+            // if we can't remove any correct node, because network will become in an
+            // incorrect state we have to remove at least 1 faulty node
```
afck: I think we only actually remove the nodes from the network after all correct nodes got `Complete(NodeChange(new_pub_keys))`, do we?
In that case, we shouldn't have to care about the _old_ `n`: We just need to make sure that less than a third of the _new_ set of nodes is faulty.

E.g. if `n` is 7 and `f` is 2, we should be able to remove two correct and one faulty node, because then `new_n` is 5 and `new_f` is 1, which is still less than a third. Even though 2 faulty out of 5 would be too much.

RicoGit: `do we?` - sure! 
`We just need to make sure that less than a third of the _new_ set of nodes is faulty.` - yes. Old set is always correct.
`E.g. if n is 7 and f is 2...` - I agree. Current implementation selects only 1 or 2 malicious nodes in this case. Do you want to improve `subset_for_remove` that it will be able to select 2 correct and 1 malicious node as well?

tests/net/mod.rs
```diff
@@ -657,6 +657,13 @@ where
         self.nodes.remove(id)
     }
 
+    /// Removes a set of nodes with the given IDs from the network and all messages addressed to
+    /// them. Returns the removed nodes if there were nodes with this IDs at the time of removal.
```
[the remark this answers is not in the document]
RicoGit: thanks!

tests/net_dynamic_hb.rs
afck: I'd say so, yes. Maybe just pick a random `new_n < n` first, and then remove as many correct nodes as possible and as many faulty nodes as necessary to keep the new set >⅔ correct.

RicoGit: done

tests/net_dynamic_hb.rs
```diff
@@ -72,23 +73,21 @@ prop_compose! {
     }
 }
 
-/// Proptest wrapper for `do_drop_and_readd`.
+/// Proptest wrapper for `do_drop_and_re_add`.
```
vkomenda: That's rather unnecessary. No need to reverting, just a note.

tests/net_dynamic_hb.rs
```diff
@@ -503,16 +503,22 @@ where
...
             rng.gen_range(1, f + 1)
         } else {
             rng.gen_range(0, f + 1)
         };
 
+        let new_f = f - remove_from_faulty;
+        let new_n = n - remove_from_faulty;
+        let remove_from_correct =
+            rng.gen_range(0, max_number_correct_nodes_for_remove(new_n, new_f) + 1);
+
```
afck: This all seems unnecessarily complicated to me. If I understand correctly, we want to:
* remove at least one node if `n > 1`, and
* make sure `new_n > 3 * new_f`.

So why not just:
```rust
if n <= 1 {
    return BTreeSet::new(); // Only one node left; nothing to remove.
}
let new_n = rng.gen_range(1, n);
let new_f = rng.gen_range(0, f.min(util::max_faulty(new_n)) + 1);
```

RicoGit: Ok. It really is a bit easier.

tests/net_dynamic_hb.rs
```diff
@@ -533,15 +526,7 @@ where
...
-        assert!(
-            n - result.len() > 0,
-            "can't remove all nodes from the network, at least one node remain"
-        );
-        println!(
-            "Max number of nodes for removing is {}, was chosen {} nodes",
-            max_number_correct_nodes_for_remove(new_n, new_f) + new_f,
-            result.len()
-        );
+        println!("Was chosen {} nodes for removing.", result.len());
```
afck: "{} nodes were chosen for removing" or
"Chose {} nodes for removing."

RicoGit:
Running tests on the CI server caught an error:
```
thread 'drop_and_re_add' panicked at 'crank: node failed to process step: Fault { reported_by: 3, faulty_id: 7, fault_kind: HbFault(SubsetFault(BaFault(DuplicateBVal))) }', libcore/result.rs:1009:5
```
Trying to reproduce the error locally I caught another error:
```
thread 'drop_and_re_add' panicked at 'crank: network queue empty', src/libcore/option.rs:1008:5
```
Second error appears when only one node left in the network after removing nodes. I'm trying to figure out what's wrong.

vkomenda:
@RicoGit, to reproduce locally, add the proptest seed from the failing test run on Travis to your local seeds.

RicoGit:
Should I add a static test case for the failed inputs? like this:
```
#[test]
fn drop_and_re_add_one_node_left() {
    let cfg = TestConfig {
        dimension: NetworkDimension::new(4, 0),
        total_txs: 20,
        batch_size: 10,
        contribution_size: 1,
        seed: [151, 234, 50, 31, 109, 65, 28, 122, 82, 93, 226, 143, 185, 27, 195, 133]
    };
    do_drop_and_re_add(cfg)
}
```

vkomenda:
No, just add the seed to `net_dynamic_hb.proptest-regressions`.

vkomenda:
This one: https://travis-ci.org/poanetwork/hbbft/jobs/499202391#L2980.

RicoGit:
I've got it. Awesome! Thanks! But local test (with the seed from CI) produces another error.

vkomenda:
Have a look at crank error variants if your local failure is `'crank: network queue empty'`. Check that the removed nodes finished sending all their messages before removal.

afck:
Does the second error always happen whenever we're left with only a single node?
That one may be related to the test (or the test net framework) and not the production code: Of course there are no messages in the queue if there's only one node. A single node will always immediately deliver a batch whenever we provide input.
If it's difficult to make the test handle that case, I'd be happy with making that a TODO for a separate PR, and restricting the test to at least two nodes for now.

The first error is strange, though. We _are_ a bit aggressive with our fault reports and sending the same `BVal` twice wouldn't break anything in production; but it still shouldn't happen in theory… :thinking:

RicoGit:
"restricting the test to at least two nodes for now." - Yes! It fixed all the tests. Thanks! I've run a test about ten times for all known seeds.
"I'd be happy with making that a TODO" - I'll make an issue for that. ok?

afck:
Great, thanks!

afck:
Great, thanks!
@vkomenda: Feel free to merge once you're happy with it.

vkomenda:
@afck, are you OK with versioning `tests/net_dynamic_hb.proptest-regressions`? It's reasonable but wasn't done before.

afck:
Ah, right, that file should be removed, since with a single node it always fails anyway. Good catch!

PR #387, opened 2019-03-12
Ported more integration tests to the new net simulator

dforsten:
Substantial progress towards resolving #322,

## Ported tests

- threshold_sign.rs
- subset.rs
- broadcast.rs
- honey_badger.rs

## Left to do:

- queueing_honey_badger.rs
- dynamic_honey_badger.rs (Or just delete—it's superseded by net_dynamic_hb.rs?)
- sync_key_gen.rs (if applicable?)

## Findings

To be reported/fixed as separate issues/pull requests.

In general the new net simulator allows to write simpler and more compact code compared to the old network simulator. There were some issues I encountered while porting which are documented in the following sections.

### NetworkInfo hard to access in Adversaries

I used the same strategy as employed in the already ported binary_agreement_mitm tests, using Arc to populate a shared map of NetworkInfo instances on test node creation to be able to access them in Adversary implementations.

Access to each node's NetworkInfo instance should be provided by the net simulator to eliminate that workaround.

### broadcast test failing with ReorderingAdversary

Initially I replaced the SilentAdversary::new(MessageScheduler::Random) with the new framework's ReorderingAdversary, which worked well for threshold_sign, but not at all for broadcast tests.
When using ReorderingAdversary the message queue would in about 50% of the cases starve, or (in fewer cases) the adversary's broadcast would actually win.

Analyzing the behavior of the "Random" scheduler in the old framework and re-implementing that behavior in a new function in adversary.rs, called "sort_by_random_node" made the tests pass again. This may point to a fragility in the broadcast algorithm implementation, although it could also be argued that the capabilities of ReorderingAdversary go beyond what an attacker would be capable of in the real world.

To reproduce:

In tests/broadcast.rs change the following line:
`            MessageSorting::RandomPick => sort_by_random_node(&mut net, rng),`
to 
`            MessageSorting::RandomPick => swap_random(&mut net, rng),`

then run
`cargo test --release test_broadcast_random_delivery_adv_propose`
a couple of times


tests/broadcast.rs
```diff
@@ -1,164 +1,236 @@
...
+            .message_limit(10_000 * size as usize)
+            .no_time_limit()
+            .adversary(new_adversary())
+            .using(move |info| {
+                let netinfo = Arc::new(info.netinfo);
+                cloned_netinfo_map
+                    .lock()
+                    .unwrap()
+                    .insert(info.id, netinfo.clone());
+                Broadcast::new(netinfo, 0).expect("Failed to create a ThresholdSign instance.")
```
afck: Should that say `Broadcast`? (Also below.)

tests/broadcast.rs
```diff
@@ -1,164 +1,236 @@
...
-        &[b' '; 32],
-    );
+    test_broadcast(net, &[b' '; 32], &mut rng);
 }
 
 #[test]
 fn test_broadcast_random_delivery_silent() {
-    let new_adversary = |_| SilentAdversary::new(MessageScheduler::Random);
-    test_broadcast_different_sizes(new_adversary, b"Foo");
+    let new_adversary = || ReorderingAdversary::new();
```
afck: That closure (and maybe also the variable) seems unnecessary.
```rust
let new_adversary = ReorderingAdversary::new;
```

tests/honey_badger.rs
```diff
@@ -1,11 +1,10 @@
 #![deny(unused_must_use)]
 //! Network tests for Honey Badger.
 
-mod network;
+pub mod net;
```
afck: Is the `pub` to avoid the "unused" warnings?

tests/honey_badger.rs
```diff
@@ -14,149 +13,165 @@ use rand::{seq::SliceRandom, Rng};
...
+        // For each untriggered epoch, send fake shares
         for (epoch, trigger_set) in &mut self.share_triggers {
             if *trigger_set {
                 // Unset the trigger.
                 *trigger_set = false;
+
+                // Get node id vectors up-front to avoid borrow issues
+                let faulty_node_ids: Vec<NodeId> =
+                    net.faulty_nodes_mut().map(|node| *node.id()).collect();
+                let all_node_ids: Vec<NodeId> = net.nodes_mut().map(|node| node.id()).collect();
```
afck: Is that closure needed?
```rust
.map(Node::id)
```

tests/honey_badger.rs
```diff
@@ -175,71 +190,84 @@ fn new_honey_badger(
...
         info!(
             "Network size: {} good nodes, {} faulty nodes",
             num_good_nodes, num_adv_nodes
         );
-        let adversary = |adv_nodes| new_adversary(num_good_nodes, num_adv_nodes, adv_nodes);
-        let network =
-            TestNetwork::new_with_step(num_good_nodes, num_adv_nodes, adversary, new_honey_badger);
-        test_honey_badger(network, num_txs);
+
+        let adversary = || new_adversary();
```
afck: Is that closure needed?

tests/honey_badger.rs
```diff
@@ -175,71 +190,84 @@ fn new_honey_badger(
...
+
+        test_honey_badger(net, num_txs);
     }
 }
 
 #[test]
 fn test_honey_badger_random_delivery_silent() {
-    let new_adversary = |_: usize, _: usize, _| SilentAdversary::new(MessageScheduler::Random);
-    test_honey_badger_different_sizes(new_adversary, 30);
+    let new_adversary = || ReorderingAdversary::new();
```
afck: (Probably unnecessary closure. Also below.)

tests/subset.rs
```diff
@@ -1,39 +1,52 @@
...
+    // Handle messages until all good nodes have terminated.
+    while !net.nodes().all(|node| node.algorithm().terminated()) {
+        let _ = net.crank_expect(&mut rng);
     }
 
-    // Verify that all instances output the same set.
-    let observer: BTreeSet<_> = network.observer.outputs().iter().cloned().collect();
-    for node in network.nodes.values() {
+    // Get reference value
+    let observer: BTreeSet<_> = net
```
afck: Not sure whether node 0 is an observer? Otherwise I'd just call it `value`.

Are all the ported tests dropping the observer nodes?
For now that's fine, I think, but we should make a note to either add them back later or abandon the observer concept.

tests/subset.rs
```diff
@@ -64,30 +77,42 @@ fn new_network<A, F>(
...
+        .build(&mut rng)
+        .expect("Could not construct test network.");
+    net
 }
 
 #[test]
 fn test_subset_3_out_of_4_nodes_propose() {
     let proposed_value = Vec::from("Fake news");
-    let proposing_ids: BTreeSet<NodeId> = (0..3).map(NodeId).collect();
+    let proposing_ids: BTreeSet<NodeId> = (0..3).map(|node_id| node_id).collect();
```
afck: Is that `map` call redundant? (Also below.)

tests/subset.rs
```diff
@@ -64,30 +77,42 @@ fn new_network<A, F>(
...
     let proposals: BTreeMap<NodeId, ProposedValue> = proposing_ids
         .iter()
         .map(|id| (*id, proposed_value.clone()))
         .collect();
-    let adversary = |_| SilentAdversary::new(MessageScheduler::First);
-    let network = new_network(3, 1, adversary);
-    test_subset(network, &proposals);
+    let adversary = || NodeOrderAdversary::new();
```
afck: (Redundant closure. Also below.)

tests/threshold_sign.rs
```diff
@@ -1,46 +1,42 @@
...
 {
-    let mut rng = rand::thread_rng();
-
-    network.input_all(());
+    /// Tests a network of threshold signing instances with an optional expected value. Outputs the
+    /// computed signature if the test is successful.
+    fn test_threshold_sign(&mut self) -> Signature {
+        let mut rng = rand::thread_rng();
+        self.broadcast_input(&(), &mut rng)
+            .expect("broadcast input failed");
```
afck: "threshold sign input failed"?

tests/threshold_sign.rs
```diff
@@ -86,24 +82,29 @@ where
...
+                .message_limit(size * (size - 1))
+                .no_time_limit()
+                .adversary(new_adversary())
+                .using(move |node_info: NewNodeInfo<_>| {
+                    ThresholdSign::new_with_document(Arc::new(node_info.netinfo), nonce.clone())
+                        .expect("Failed to create a ThresholdSign instance.")
+                })
+                .build(&mut rng)
+                .expect("Could not construct test network.");
+            let coin = net.test_threshold_sign().parity();
```
afck: Not your fault, but now that we've replaced the `Coin` (uses threshold signatures as random values) with `ThresholdSign` itself, maybe the test should also verify that the signature is valid, probably inside `test_threshold_sign`. (Feel free to just add a `TODO` for now.)

afck:
I'll try to debug the broadcast issue.
One more thing regarding the new broadcast tests that I'm still a bit confused about: The old `SilentAdversary` used to never send any messages at all (basically just simulating crashed nodes). If I remember correctly, the new test framework by default makes the adversary behave like correct nodes: We should still have a test with an adversary that actually drops all their own messages instead of sending them.

afck:
> or (in fewer cases) the adversary's broadcast would actually win.

Is it still guaranteed that the adversary doesn't actually control the proposer node? Maybe the new framework randomizes their positions. Basically what we need to check is:
* If the proposer is correct, everyone outputs what the proposer sent.
* If the proposer is faulty, everyone outputs either nothing, or everyone outputs the same thing (doesn't matter what).

tests/honey_badger.rs
dforsten: Yes, that seems to be the strategy the already ported tests are following, like the binary_agreement test.

tests/honey_badger.rs
dforsten: It does expect a function taking a &NodeMutHandle as first argument, this gets close:
`.map(NodeMutHandle::id)`

but not quite, it has an extra lifetime parameter, so it boils down to the required:
`fn(net::adversary::NodeMutHandle<'_, hbbft::sender_queue::SenderQueue<hbbft::honey_badger::honey_badger::HoneyBadger<std::vec::Vec<usize>, u16>>>) -> _
`
vs NodeMutHandle's:
`for<'r> fn(&'r net::adversary::NodeMutHandle<'_, hbbft::sender_queue::SenderQueue<hbbft::honey_badger::honey_badger::HoneyBadger<std::vec::Vec<usize>, u16>>>) -> _`

tests/subset.rs
dforsten: Yes, I thought that naming was a bit odd when porting from the old tests, I renamed it to "expected_value" which makes a lot more sense.
The strategy of using the first node's output as reference value I took from the new net simulator's documentation, I make a note in the code to indicate that the value should be taken from an observer as soon as we have such a concept in the new simlator.

tests/broadcast.rs
```diff
@@ -1,164 +1,236 @@
...
+    netinfo_mutex: Arc<Mutex<NetworkInfoMap>>,
 }
 
 impl ProposeAdversary {
-    /// Creates a new replay adversary with the given message scheduler.
-    fn new(
-        scheduler: MessageScheduler,
-        adv_nodes: BTreeMap<NodeId, Arc<NetworkInfo<NodeId>>>,
-    ) -> ProposeAdversary {
+    /// Create a new `ProposeAdversary`.
```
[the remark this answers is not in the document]
dforsten: Fixed

tests/broadcast.rs
dforsten: Fixed

tests/broadcast.rs
dforsten: Eliminated

tests/honey_badger.rs
dforsten: Eliminated

tests/honey_badger.rs
dforsten: Eliminated

tests/subset.rs
dforsten: Indeed - removed!

tests/subset.rs
dforsten: Eliminated

tests/threshold_sign.rs
dforsten: Fixed

tests/threshold_sign.rs
dforsten: Added a TODO

dforsten:
> I'll try to debug the broadcast issue.
> One more thing regarding the new broadcast tests that I'm still a bit confused about: The old `SilentAdversary` used to never send any messages at all (basically just simulating crashed nodes). If I remember correctly, the new test framework by default makes the adversary behave like correct nodes: We should still have a test with an adversary that actually drops all their own messages instead of sending them.



> I'll try to debug the broadcast issue.
> One more thing regarding the new broadcast tests that I'm still a bit confused about: The old `SilentAdversary` used to never send any messages at all (basically just simulating crashed nodes). If I remember correctly, the new test framework by default makes the adversary behave like correct nodes: We should still have a test with an adversary that actually drops all their own messages instead of sending them.

I will make ProposeAdversary optionally drop all messages produced by regular operation and add a test switching that option on.

dforsten:
> I will make ProposeAdversary optionally drop all messages produced by regular operation and add a test switching that option on.
Changes committed, the new test dropping messages similar to the tests in the old framework is called "test_broadcast_random_delivery_adv_propose_and_drop".

afck:
I restarted it, and made a note of the random seed:
```
thread 'drop_and_re_add' panicked at 'Test failed: crank: node failed to process step: Fault { reported_by: 5, faulty_id: 6, fault_kind: SyncKeyGenPart(MultipleParts) }; minimal failing input: cfg = TestConfig { dimension: NetworkDimension { size: 11, faulty: 0 }, total_txs: 35, batch_size: 12, contribution_size: 7, seed: [31, 66, 156, 122, 45, 158, 241, 243, 61, 138, 255, 224, 247, 45, 201, 112] }
```

dforsten:
> > or (in fewer cases) the adversary's broadcast would actually win.
> 
> Is it still guaranteed that the adversary doesn't actually control the proposer node? Maybe the new framework randomizes their positions. Basically what we need to check is:
> 
>     * If the proposer is correct, everyone outputs what the proposer sent.
> 
>     * If the proposer is faulty, everyone outputs either nothing, or everyone outputs the same thing (doesn't matter what).
> 
> 
> (Edit: By "everyone" I only mean the correct nodes, of course.)

Randomly swapping any message in the queue to the top could mean that sometimes the adversary gets its messages sent and processed before the actual proposer node had the chance to send his - is that a viable attack scenario?

afck:
> is that a viable attack scenario?

No, in the Broadcast algorithm, the proposer is a predefined node. Even if a different node poses as the proposer, the others should ignore its messages. I think the algorithm should really work for every message ordering.

dforsten:
> > or (in fewer cases) the adversary's broadcast would actually win.
> 
> Is it still guaranteed that the adversary doesn't actually control the proposer node? Maybe the new 

In the old framework the input was just sent to the node with index 0, that may not be a good idea with the new net simulator framework. I will change that and make sure the index is actually from a correct node, and see if that fixes the issue.

dforsten:
> > > or (in fewer cases) the adversary's broadcast would actually win.
> > 
> > 
> > Is it still guaranteed that the adversary doesn't actually control the proposer node? Maybe the new
> 
> In the old framework the input was just sent to the node with index 0, that may not be a good idea with the new net simulator framework. I will change that and make sure the index is actually from a correct node, and see if that fixes the issue.

Tricky though, on Broadcast Algorithm creation the VirtualNet is not available to check for the correct nodes yet, it only knows if it itself is correct or faulty.
Which means that if the first node is faulty we cannot initialize the Broadcast Algorithm with a correct proposer id!

dforsten:
> > > > or (in fewer cases) the adversary's broadcast would actually win.
> > > 
> > > 
> > > Is it still guaranteed that the adversary doesn't actually control the proposer node? Maybe the new
> > 
> > 
> > In the old framework the input was just sent to the node with index 0, that may not be a good idea with the new net simulator framework. I will change that and make sure the index is actually from a correct node, and see if that fixes the issue.
> 
> Tricky though, on Broadcast Algorithm creation the VirtualNet is not available to check for the correct nodes yet, it only knows if it itself is correct or faulty.
> Which means that if the first node is faulty we cannot initialize the Broadcast Algorithm with a correct proposer id!

By reading the source of net/mod.rs we can use arcane knowledge to pick a node Id which is valid, since all ids lower than the number of faulty ids are actually faulty:

`let is_faulty = idx < faulty;`

Not a very robust approach though, if that assumption breaks in net simulator code, the test breaks as well.

afck:
Yes, that's annoying. I guess we'll need to extend the framework to allow excluding specific IDs from the faulty set.
Anyway, in the long run the tests should definitely cover both cases: A faulty proposer and a correct proposer.
For now, we could just _detect_ whether the proposer is faulty, and if it is, we'd only check whether all correct nodes output the same value (or none of them output anything), and add a TODO for splitting up the test later.

afck:
You're right, we could also add that fact to the documentation (that the faulty nodes come first), and then use it in the test.

dforsten:
> You're right, we could also add that fact to the documentation (that the faulty nodes come first), and then use it in the test.

And that actually fixed the tests, 
if I pick the first non-faulty idx as proposer the test runs stable even with random swapping.

The solution you suggested to check if the proposer is faulty seems more elegant and stable to me compared to using net-simulator internal behaviors, so I will implement that, picking the proposer at random.

After that we should convert that test to use proptest and seeds as soon as possible.

dforsten:
> > You're right, we could also add that fact to the documentation (that the faulty nodes come first), and then use it in the test.
> 
> And that actually fixed the tests,
> if I pick the first non-faulty idx as proposer the test runs stable even with random swapping.
> 
> The solution you suggested to check if the proposer is faulty seems more elegant and stable to me compared to using net-simulator internal behaviors, so I will implement that, picking the proposer at random.
> 
> After that we should convert that test to use proptest and seeds as soon as possible.

Well, that would fix the issue with the faulty proposer winning, but it would not fix the queue from stalling, which is the more common case when the proposer is a faulty node.

afck:
> it would not fix the queue from stalling, which is the more common case when the proposer is a faulty node.

Probably because the proposer actually sends _both_ values. In that case, the algorithm is just supposed to make all correct nodes do the same: output nothing, or output one of those values.

If the queue is empty, that probably means they output nothing. In that case, we just need to check that none of the correct nodes has an output.

dforsten:
> > it would not fix the queue from stalling, which is the more common case when the proposer is a faulty node.
> 
> Probably because the proposer actually sends _both_ values. In that case, the algorithm is just supposed to make all correct nodes do the same: output nothing, or output one of those values.
> 
> If the queue is empty, that probably means they output nothing. In that case, we just need to check that none of the correct nodes has an output.

The changes to randomize the proposer (which can be faulty) and to check all three outcomes is now committed and the tests run stable even with the Adversary using random_swap!

tests/broadcast.rs
```diff
@@ -125,24 +125,48 @@ fn test_broadcast<A: Adversary<Broadcast<NodeId>>>(
     mut net: VirtualNet<Broadcast<NodeId>, A>,
     proposed_value: &[u8],
     rng: &mut ThreadRng,
+    proposer_id: NodeId,
 ) {
     // This returns an error in all but the first test.
     let _ = env_logger::try_init();
 
+    let proposer_is_faulty = net.get(proposer_id).unwrap().is_faulty();;
```
afck: Duplicated `;`.

dforsten: Oh, peculiar that clippy did not catch that .

dforsten: Fixed

dforsten:
Is there anything I can do from my end to push this pull request along?
I would like to continue with resolving the issue with NetworkInfo access and port queueing_honey_badger.rs to be able to remove the old network simulator completely.

afck:
From my side, feel free to add on top of this PR, and dismiss my approval whenever you add a major change. What do you prefer, @vkomenda?

tests/net/adversary.rs
```diff
@@ -435,9 +525,83 @@ where
...
+
+    #[inline]
+    fn tamper<R: Rng>(
+        &mut self,
+        mut net: NetMutHandle<'_, D, Self>,
+        msg: NetMessage<D>,
+        rng: &mut R,
+    ) -> Result<CpStep<D>, CrankError<D>> {
+        // only replay a message in some cases
+        if randomly(self.p_replay) {
```
vkomenda: The old `randomly` doesn't fit in the `net` framework because it leads to non-reproducible tests due to the use of `thread_rng`. Can you use `rng` here instead?

tests/net/adversary.rs
```diff
@@ -435,9 +525,83 @@ where
...
+            if let Some(picked_node) = random_node(&mut net, rng) {
+                let mut new_msg = msg.clone();
+                new_msg.from = new_msg.to;
+                new_msg.to = picked_node;
+                net.inject_message(QueuePosition::Back, new_msg);
+            }
         }
+
+        // Possibly inject more messages
+        while randomly(self.p_inject) {
```
vkomenda: Please use the `rng` argument instead of `thread_rng`. `rng` could be passed to `randomly` as an argument.

tests/subset.rs
```diff
@@ -1,39 +1,54 @@
...
+    mut net: VirtualNet<Subset<NodeId, u8>, A>,
     inputs: &BTreeMap<NodeId, ProposedValue>,
-) {
-    let ids: Vec<NodeId> = network.nodes.keys().cloned().collect();
+) where
+    A: Adversary<Subset<NodeId, u8>>,
+{
+    let ids: Vec<NodeId> = net.nodes().map(|node| *node.id()).collect();
+
+    let mut rng = rand::thread_rng();
```
vkomenda: Please use `proptest::TestRng` to make tests reproducible given a seed. See `net_dynamic_hb.rs` and `binary_agreement_mitm.rs` for the relevant `proptest` macros that allow that.

tests/subset.rs
```diff
@@ -56,38 +71,52 @@ fn test_subset<A: Adversary<Subset<NodeId, u8>>>(
...
-    F: Fn(BTreeMap<NodeId, Arc<NetworkInfo<NodeId>>>) -> A,
+    F: Fn() -> A,
 {
     // This returns an error in all but the first test.
     let _ = env_logger::try_init();
 
-    let new_subset =
-        |netinfo: Arc<NetworkInfo<NodeId>>| Subset::new(netinfo, 0).expect("new Subset instance");
-    TestNetwork::new(good_num, bad_num, adversary, new_subset)
+    let mut rng = rand::thread_rng();
```
vkomenda: Likewise, please don't use `thread_rng` but use `TestRng` instead.

tests/threshold_sign.rs
```diff
@@ -1,46 +1,43 @@
...
 where
     A: Adversary<ThresholdSign<NodeId>>,
 {
-    let mut rng = rand::thread_rng();
-
-    network.input_all(());
+    /// Tests a network of threshold signing instances with an optional expected value. Outputs the
+    /// computed signature if the test is successful.
+    fn test_threshold_sign(&mut self) -> Signature {
+        let mut rng = rand::thread_rng();
```
vkomenda: Use `TestRng`.

tests/net/util.rs
```diff
@@ -16,3 +18,12 @@ macro_rules! try_some {
         }
     };
 }
+
+/// Return true with a certain `probability` ([0 .. 1.0]).
+pub fn randomly(probability: f32) -> bool {
+    assert!(probability <= 1.0);
+    assert!(probability >= 0.0);
+
+    let mut rng = rand::thread_rng();
```
vkomenda: This is not reproducible. Please pass an RNG as an argument and use it instead.

tests/broadcast.rs
```diff
@@ -1,164 +1,285 @@
...
+fn test_broadcast_different_sizes<A, F>(
+    new_adversary: F,
+    proposed_value: &[u8],
+    adversary_netinfo: &Arc<Mutex<NetworkInfoMap>>,
+) where
     A: Adversary<Broadcast<NodeId>>,
-    F: Fn(BTreeMap<NodeId, Arc<NetworkInfo<NodeId>>>) -> A,
+    F: Fn() -> A,
 {
     let mut rng = rand::thread_rng();
```
vkomenda: Use `TestRng`.

tests/broadcast.rs
```diff
@@ -1,164 +1,285 @@
...
+
+        test_broadcast(net, proposed_value, &mut rng, proposer_id);
     }
 }
 
 #[test]
 fn test_8_broadcast_equal_leaves_silent() {
-    let adversary = |_| SilentAdversary::new(MessageScheduler::Random);
+    let new_adversary = || ReorderingAdversary::new();
+    let mut rng = rand::thread_rng();
```
vkomenda: Use `TestRng`.

tests/honey_badger.rs
```diff
@@ -14,149 +13,165 @@ use rand::{seq::SliceRandom, Rng};
...
     A: Adversary<UsizeHoneyBadger>,
 {
-    let new_queue = |id: &NodeId| (*id, (0..num_txs).collect::<Vec<usize>>());
-    let mut queues: BTreeMap<_, _> = network.nodes.keys().map(new_queue).collect();
+    let mut queues: BTreeMap<_, _> = net
+        .correct_nodes()
+        .map(|node| (*node.id(), (0..num_txs).collect::<Vec<usize>>()))
+        .collect();
+
+    let mut rng = rand::thread_rng();
```
vkomenda: Use `TestRng`

tests/honey_badger.rs
```diff
@@ -175,71 +190,80 @@ fn new_honey_badger(
...
+    adversary_netinfo: &Arc<Mutex<NetworkInfoMap>>,
+) where
     A: Adversary<UsizeHoneyBadger>,
-    F: Fn(usize, usize, BTreeMap<NodeId, Arc<NetworkInfo<NodeId>>>) -> A,
+    F: Fn() -> A,
 {
     // This returns an error in all but the first test.
     let _ = env_logger::try_init();
 
     let mut rng = rand::thread_rng();
```
vkomenda: Use `TestRng`.

dforsten:
> Looks good to me. The part that I think still needs to be done is making these test reproducible as it in `net_dynamic_hb.rs` and `binary_agreement_mitm.rs`.

One issue I am running into right now is that the tests using the proptest/TestRng setup now take a *loooong* time to run. Half an hour for the (release) test_broadcast_random_adversary test alone, and that on a 9900K CPU at 5GHz.

dforsten:
> > Looks good to me. The part that I think still needs to be done is making these test reproducible as it in `net_dynamic_hb.rs` and `binary_agreement_mitm.rs`.
> 
> One issue I am running into right now is that the tests using the proptest/TestRng setup now take a _loooong_ time to run. Half an hour for the (release) test_broadcast_random_adversary test alone, and that on a 9900K CPU at 5GHz.

Limiting the "cases" field in the proptest config to a low number is probably the best way to go with elaborate tests like these, I will see to constrain that value for tests which would cause excessive run times otherwise.

tests/net/adversary.rs
dforsten: Switched to TestRng

tests/net/adversary.rs
dforsten: Passing rng to randomly now

tests/subset.rs
dforsten: Switched to proptest and TestRng

tests/subset.rs
dforsten: Switched to TestRng

tests/threshold_sign.rs
dforsten: Switched to TestRng

tests/net/util.rs
dforsten: Eliminated thread_rng

tests/broadcast.rs
dforsten: Using TestRng

tests/broadcast.rs
dforsten: Using TestRng

tests/honey_badger.rs
dforsten: Using TestRng

tests/honey_badger.rs
dforsten: Using TestRng

dforsten:
The drop_and_re_add test failed again, which was not touched by the changes in this PR.

Could we re-start TravisCI please?

afck:
Are you running the tests with [mock crypto](https://github.com/poanetwork/hbbft/blob/3336fa772dc27abc15c731de073a99e11d1f38e4/Cargo.toml#L64)? I know that they are very slow otherwise.

dforsten:
When limiting the number of cases to 1 in the proptest config the tests run as fast as before.

What is more concerning are the random failures in the drop_and_re_add test.
I cannot reproduce any of the failures reported by Travis CI on my local machine, and there do not seem to be any changes in my branch which would alter the behavior of that test.

tests/broadcast.rs
```diff
@@ -214,10 +216,57 @@ fn test_broadcast_different_sizes<A, F>(
...
+    #[allow(clippy::unnecessary_operation)]
+    fn test_broadcast_random_adversary(seed in gen_seed()) {
+        do_test_broadcast_random_adversary(seed)
+    }
+}
+
+fn do_test_8_broadcast_equal_leaves_silent(seed: TestRngSeed) {
     let new_adversary = || ReorderingAdversary::new();
-    let mut rng = rand::thread_rng();
+    let mut rng: TestRng = TestRng::from_seed(seed);
```
afck: There's another unnecessary closure above that line.

dforsten: Fixed

dforsten:
I checked out the **master** branch and ran the drop_and_re_add test two times, and got the same error as on TravisCI:

```
thread 'drop_and_re_add' panicked at 'crank: node failed to process step: Fault { reported_by: 6, faulty_id: 6, fault_kind: SyncKeyGenPart(MultipleParts) }', libcore/result.rs:1009:5
proptest: FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs
proptest: Saving this and future failures in /home/davidf/branches/honeybadger/hbbft_original/tests/net_dynamic_hb.proptest-regressions
proptest: If this test was run on a CI system, you may wish to add the following line to your copy of the file. (You may need to create it.)
xs 3923584502 2719340074 1697030405 1410331824 
thread 'drop_and_re_add' panicked at 'Test failed: crank: node failed to process step: Fault { reported_by: 6, faulty_id: 6, fault_kind: SyncKeyGenPart(MultipleParts) }; minimal failing input: cfg = TestConfig { dimension: NetworkDimension { size: 7, faulty: 1 }, total_txs: 24, batch_size: 13, contribution_size: 3, seed: [106, 18, 211, 205, 212, 173, 194, 179, 76, 21, 231, 223, 189, 174, 65, 72] }
	successes: 0
	local rejects: 0
	global rejects: 0
', tests/net_dynamic_hb.rs:78:1

```

This issue does **not** seem to be introduced with my branch.

dforsten:
Could you try running:

`time cargo test --release drop_and_re_add`

on the **master** branch?

I am reproducing the error in about 50% of the runs on my local machine.

afck:
> `Fault { reported_by: 6, faulty_id: 6, fault_kind: SyncKeyGenPart(MultipleParts) }`

So node 6 reported itself for producing multiple `Part` messages in the key generation process…? And if this makes the test fail, that means that node 6 was a correct one in that case, right? :thinking: 
I'll also try it out locally. I don't have an idea yet what might be going wrong here.

dforsten:
The reason I can now reproduce the issue on master so easily is because it replays the regression case file the first fail created, here it is:

[net_dynamic_hb.proptest-regressions.zip](https://github.com/poanetwork/hbbft/files/2966062/net_dynamic_hb.proptest-regressions.zip)

afck:
I also get it on `master`, yes. Sometimes `MessageLimitExceeded` (even if I increase the limit) and sometimes the `MultipleParts`. We have a lot of debugging to do… :sweat_smile:

dforsten:
Should we do the debugging in the course of this PR though?

afck:
> Should we do the debugging in the course of this PR though?

No, I'm looking into it on `master` now. Not sure yet whether it's a problem with the test or with DHB itself.

vkomenda:
> The reason I can now reproduce the issue on master so easily is because it replays the regression case file the first fail created

@dforsten, `net_dynamic_hb` passes for me with that seed you gave, both in master and in your branch.  Maybe we should compare our `Cargo.lock` files in case of different dependent package versions.

vkomenda:
I got `drop_and_re_add` to fail with this seed:

```
xs 2949746464 2777896929 710146962 3572918699 # shrinks to cfg = TestConfig { dimension: NetworkDimension { size: 15, faulty: 3 }, total_txs: 23, batch_size: 17, contribution_size: 3, seed: [184, 35, 113, 127, 105, 136, 58, 230, 126, 31, 22, 139, 100, 194, 239, 19] }
```

afck:
OK, I think #388 fixes _most_ of the test failures, at least. Still running the test in a loop to find more…

vkomenda:
@dforsten, to fix the current CI build failure, I think the following change would be OK:
```
diff --git a/tests/threshold_sign.rs b/tests/threshold_sign.rs
index 55a8a97..a904313 100644
--- a/tests/threshold_sign.rs
+++ b/tests/threshold_sign.rs
@@ -50,7 +50,7 @@ const GOOD_SAMPLE_SET: f64 = 400.0;
 /// size.
 fn check_coin_distribution(num_samples: usize, count_true: usize, count_false: usize) {
     // Maximum 40% expectation in case of 400 samples or more.
-    const EXPECTED_SHARE: f64 = 0.4;
+    const EXPECTED_SHARE: f64 = 0.33;
     let max_gain = GOOD_SAMPLE_SET.log2();
     let num_samples_f64 = num_samples as f64;
     let gain = num_samples_f64.log2().min(max_gain);
```
So instead of a 2/5 share of either `true` or `false` throws, we'd expect only a 1/3 share. This is to counteract non-uniformity of the distribution with a low number of samples, in this case 50.

dforsten:
I increased the tolerance in the coin distribution test to avoid random test failures and updated the feature branch with the latest changes from master.

Should I continue to check in work to this branch or will the current PR commits be merged/rebased to master?

vkomenda:
Let's continue with the remaining tests in other PRs.

PR #391, opened 2019-03-26
Don't drop re-added peers from sender queue.

### LINKED ISSUE #390: Test `drop_and_re_add` fails.
The `drop_and_re_add` test fails with the following `net_dynamic_hb.proptest_regressions` file:
```
xs 3766095200 4099363815 241143382 3155674404 # shrinks to cfg = TestConfig { dimension: NetworkDimension { size: 10, faulty: 3 }, total_txs: 27, batch_size: 12, contribution_size: 7, seed: [73, 242, 153, 97, 138, 68, 78, 181, 145, 37, 155, 185, 206, 103, 141, 115] }
```
Node 5's `DynamicHoneyBadger` never receives the first batch after getting re-added. The `Subset` instances for the subsequent batches complete, but the node is stalled waiting for the one in epoch 5.

The reason is that the `Broadcast` instances with nodes 3 and 7 as proposers don't terminate (the corresponding `BinaryAgreement` instances both returned `true`). I haven't found out why yet:
* It could be a bug in `Broadcast`,
* or a problem with the `SenderQueue` (possibly it drops the removed nodes _after_ they have already been re-added),
* or a problem in the test itself (which needs to be refactored anyway, since it's really hard to read).

afck:
If a previously removed peer gets added back as a validator, `SenderQueue` now removes that peer from `last_epochs`, so it doesn't drop it later.

(Also minor cleanups in the `net_dynamic_hb` test module.)

Closes #390.

### DIFF
--- a/tests/net_dynamic_hb.rs
+++ b/tests/net_dynamic_hb.rs
@@ -17,7 +17,7 @@ use threshold_crypto::PublicKey;
 
 type DHB = SenderQueue<DynamicHoneyBadger<Vec<usize>, usize>>;
 
-/// Choose a node's contribution for an epoch.
+/// Chooses a node's contribution for an epoch.
 ///
 /// Selects randomly out of a slice, according to chosen batch and contribution sizes. The function
 /// will not fail to do so, even if the queue is empty, returning a smaller or empty slice
@@ -89,6 +89,9 @@ proptest! {
 // TODO: Add an observer node to the test network.
 #[allow(clippy::needless_pass_by_value, clippy::cyclomatic_complexity)]
 fn do_drop_and_re_add(cfg: TestConfig) {
+    // This returns an error in all but the first test.
+    let _ = env_logger::try_init();
+
     let mut rng: TestRng = TestRng::from_seed(cfg.seed);
 
     // First, we create a new test network with Honey Badger instances.
@@ -122,8 +125,7 @@ fn do_drop_and_re_add(cfg: TestConfig) {
     let nodes_for_remove = state.subset_for_remove(&mut rng);
     println!("Will remove and re-add nodes {:?}", nodes_for_remove);
 
-    // We generate a list of transaction we want to propose, for each node. All nodes will propose
-    // a number between 0..total_txs, chosen randomly.
+    // We generate a list of total_txs transactions we want to propose, for each node.
     let mut queues: BTreeMap<_, Vec<usize>> = state
         .net
         .nodes()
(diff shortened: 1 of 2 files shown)


tests/net_dynamic_hb.rs
```diff
@@ -108,11 +111,7 @@ fn do_drop_and_re_add(cfg: TestConfig) {
                 id
             );
             let dhb = DynamicHoneyBadger::builder().build(node.netinfo.clone());
-            SenderQueue::builder(
-                dhb,
-                node.netinfo.all_ids().filter(|&&them| them != id).cloned(),
-            )
-            .build(node.id)
+            SenderQueue::builder(dhb, node.netinfo.all_ids().cloned()).build(node.id)
```
vkomenda: This might work because `SenderQueue` filters out its own ID internally. But conceptually the IDs here are all IDs other than the ID of the node for which the `SenderQueue` is being built.

afck: I'm not sure about that: Isn't it usually convenient to be able to just pass all the network's IDs to the constructor/builder?
If we decide against that, should we add a check instead, and make the builder fail if it contains our own ID?

(I also found some branches in the sender queue code that looked like we were expected to actually have an entry for our own ID…)

vkomenda: I'm not sure about convenience. Our epoch is known from `algo.epoch()`. You would duplicate it by including the pair of our ID and our epoch into `peer_epochs`.

afck: That's true. But to my mind, that means we should either filter our own ID, or fail if we find it. And the sender queue shouldn't contain code that checks for our own ID in the map.

vkomenda: `peer_epochs` is updated on a message by the `SenderQueue`. On the contrary, `self.algo.epoch` is updated internally by the algorithm. That's why I think these two things should not overlap and we should not duplicate own epoch by including it in `peer_epochs`.

afck: Yes, I completely agree. The question is just how we prevent it:
* Should the constructor/builder panic if it sees our own ID among the peer IDs?
* Or should it just filter our won ID from them?

vkomenda: An assertion in the constructor/builder would be great. After that, if own ID appears in `peer_epochs` that must be due to user code sending a message with the sender being us. That's arguably a bug in the user code. Could be a panic too, what do you think?

afck: I still feel it would be a more convenient API to allow the user to just pass in all node IDs (and automatically filter out the own ID). E.g. imagine you're implementing a blockchain client and there's a chain spec configuring the initial set of validators. It would be convenient to be able to just pass that list into the `SenderQueue` constructor. I'd actually guess that in the majority of use-cases, the user would have to do this manual filtering if we add that check.

Anyway, it's definitely unrelated to the actual test fix, so I reverted my changes for now. We can modify the `SenderQueueBuilder` in a separate PR.

PR #393, opened 2019-04-01
Use the serde derive feature and update other dependencies

vkomenda:
This addresses the problem in poanetwork/threshold_crypto#78. The `threshold_crypto` dependency in this PR needs to be changed after `threshold_crypto` 0.3.1 is uploaded to crates.io.


afck:
Looks good to me! This can switch to the published threshold_crypto 0.3.1 now.

PR #404, opened 2019-04-25
Extract test framework into hbbft_testing crate.

afck:
Also update Rust and some dependencies.


dforsten:
After a couple of fix-ups of dependencies in the hbbft_engine crate it compiled fine against both, the hbbft and the hbbft_testing crates.
Thank you!

Issue #309, opened 2018-10-29
Examine `Broadcast` for optimization potential.


mbr:
While debugging an issue with failing tests, I encountered some inefficiency in the reliable broadcast algorithm, at least inside the edge-case of a two node network. Consider the following partial log:

```
B: [0] -> [1]: HoneyBadger(0, Message { epoch: 0, content: Subset(Broadcast(0, Value(Proof { #1, root_hash: c002..beaa, value: b808..6d14, .. }))) })
D: [0] -> [1]: HoneyBadger(0, Message { epoch: 0, content: Subset(Broadcast(0, Echo(Proof { #0, root_hash: c002..beaa, value: 0000..3a39, .. }))) })
E: [1] -> [0]: HoneyBadger(0, Message { epoch: 0, content: Subset(Broadcast(0, Echo(Proof { #1, root_hash: c002..beaa, value: b808..6d14, .. }))) })
G: [1] -> [0]: HoneyBadger(0, Message { epoch: 0, content: Subset(Broadcast(0, Ready(c002..beaa))) })
J: [0] -> [1]: HoneyBadger(0, Message { epoch: 0, content: Subset(Broadcast(0, Ready(c002..beaa))) })
```

This is almost in line with the algorithm description on [page 7](https://eprint.iacr.org/2016/199.pdf); we are sending `VAL(h, b_i, s_i)` to every node in the network and expect an `ECHO(h, b_i, s_i)` back for each, finally a `READY(h)`. The full text-book message log would look like this:

```
A: [0->0] VAL(h, b_0, s_0)
B: [0->1] VAL(h, b_1, s_1)
C: [0->0] ECHO(h, b_0, s_0)
D: [0->1] ECHO(h, b_0, s_0)
E: [1->0] ECHO(h, b_1, s_1)
F: [1->1] ECHO(h, b_1, s_1)
G: [1->0] READY(h)
H: [1->1] READY(h)
I: [0->0] READY(h)
J: [0->1] READY(h)
```

This is twice as many messages, because `A`, `C`, `F`, `H` and `I` are messages that need not leave the node and are short-circuited inside the code already (instead of returning it an `Step` output, it is directly handled by calling the respective `handle_` function).

## Possible optimizations

[...]

Issue #396, opened 2019-04-03
More tests: Broadcast


afck:
Add an adversary that sends random (but "valid looking") messages. The distributions should be general, but somewhat intelligent, so that they are more likely to uncover bugs.

Also, the original idea behind the `ProposeAdversary` was that the faulty nodes would act as if they were part of another network with the same proposer but a different proposed value. (Currently they're just all proposing, which in most cases will simply be ignored.)
So the attacker should simulate a separate network with the same IDs and the same proposer but a different input value. It should then send in the real network the same messages that its nodes sent in the simulation.

PR #405, opened 2019-04-29
Optimized broadcast #309

pawanjay176:
WIP for #309 
- [x] Add the extra CanDecode and EchoHash messages in the Messages enum.
- [x] Have pessimism factor g as a configurable parameter in the Broadcast struct.
- [x] Write handle_echo_hash and handle_can_decode functions and modify other handle functions accordingly.
- [ ] Change handle_value to prevent echoing a message back to proposer. Would save (N-1) messages regardless of g (original idea of the issue).
- [x] Test normal flow of messages
- [x] Test for expected reduction in number of messages with different values of g.
- [ ] Additional testing

### DIFF
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -395,7 +395,7 @@ impl<N: NodeIdT> Broadcast<N> {
         // after arranging all node ids in a circular list.
         let left = self
             .netinfo
-            .all_ids_except(self.proposer_id())
+            .all_ids()
             .cycle()
             .skip_while(|x| *x != self.our_id())
             .take(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
@@ -438,7 +438,7 @@ impl<N: NodeIdT> Broadcast<N> {
         // after arranging all node ids in a circular list.
         let right = self
             .netinfo
-            .all_ids_except(self.proposer_id())
+            .all_ids()
             .cycle()
             .skip_while(|x| *x != self.our_id())
             .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
@@ -466,7 +466,7 @@ impl<N: NodeIdT> Broadcast<N> {
         // after arranging all node ids in a circular list.
         let right = self
             .netinfo
-            .all_ids_except(self.proposer_id())
+            .all_ids()
             .cycle()
             .skip_while(|x| *x != self.our_id())
             .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -242,7 +242,7 @@ impl<N: NodeIdT> Broadcast<N> {
         }
         // Send the proof in an `Echo` message to left nodes
         // and `EchoHash` message to right nodes and handle the response.
-        let echo_steps = self.send_echo(p.clone())?;
+        let echo_steps = self.send_echo_left()?;
         let echo_hash_steps = self.send_echo_hash(p.root_hash())?;
         Ok(echo_steps.join(echo_hash_steps))
     }
@@ -362,44 +362,55 @@ impl<N: NodeIdT> Broadcast<N> {
         }
         // Upon receiving 2f + 1 matching Ready(h) messages, send full
         // `Echo` message to every node who hasn't sent us a `CanDecode`
-        let our_value = self.echos.get(self.our_id()).unwrap();
         if self.count_readys(hash) == 2 * self.netinfo.num_faulty() + 1 {
             // `send_echo` function should take care of which nodes to send the message
-            step.extend(self.send_echo(our_value)?);
+            step.extend(self.send_echo_remaining()?);
         }
 
         Ok(step.join(self.compute_output(hash)?))
     }
 
-    /// Sends an `Echo` message and handles it. Does nothing if we are only an observer.
-    fn send_echo(&mut self, p: Proof<Vec<u8>>) -> Result<Step<N>> {
-       if !self.netinfo.is_validator() {
+    /// Sends `Echo` message to all left nodes.
+    fn send_echo_left(&mut self) -> Result<Step<N>> {
+        if !self.netinfo.is_validator() {
             return Ok(Step::default());
         }
+        // Caller should make sure message is inserted. unwrap should never panic.
+        let p = self.echos.get(self.our_id()).unwrap().clone();
         let echo_msg = Message::Echo(p.clone());
         let mut step = Step::default();
-        // TODO: iterator not correctly describing right of our_id. Use cycle method on
+        // TODO: iterator not correctly describing left of our_id. Use cycle method on
         // iterator to correctly specify.
-        let right = self.netinfo.all_ids().skip(self.netinfo.num_correct()+self.pessimism_factor);
+        let left = self.netinfo.all_ids_except(self.proposer_id()).take(self.netinfo.num_correct()+self.pessimism_factor);
+        for id in left {
+            let msg = Target::Node(id.clone()).message(echo_msg.clone());
+            step.messages.push(msg);
+        }
+        let our_id = &self.our_id().clone();
+        Ok(step.join(self.handle_echo(our_id, p)?))
+    }
 
-        // Send `Echo` to all left nodes.
-        if !self.echo_sent_partial {
-            self.echo_sent_partial = true;
-            for id in right {
+    /// Sends `Echo` message to remaining nodes who haven't sent `CanDecode`
+    fn send_echo_remaining(&mut self) -> Result<Step<N>> {
+        // No need for echo_sent_partial check
+        self.echo_sent_full = true;
+        if !self.netinfo.is_validator() {
+            return Ok(Step::default());
+        }
+        // Caller should make sure message is inserted. unwrap should never panic.
+        let p = self.echos.get(self.our_id()).unwrap().clone();
+        let echo_msg = Message::Echo(p.clone());
+        let mut step = Step::default();
+        // TODO: iterator not correctly describing left of our_id. Use cycle method on
+        // iterator to correctly specify.
+        let right = self.netinfo.all_ids_except(self.proposer_id()).skip(self.netinfo.num_correct()+self.pessimism_factor);
+        for id in right {
+            if !self.can_decodes.contains_key(id) {
                 let msg = Target::Node(id.clone()).message(echo_msg.clone());
                 step.messages.push(msg);
             }
         }
-        // Send `Echo` to remaining right nodes who haven't sent us `CanDecode`.
-        else {
-            self.echo_sent_full = true;
-            for id in right {
-                if !self.can_decodes.contains_key(id) {
-                    let msg = Target::Node(id.clone()).message(echo_msg.clone());
-                    step.messages.push(msg);
-                }
-            }
-        }
+        // Is this required?!
         let our_id = &self.our_id().clone();
         Ok(step.join(self.handle_echo(our_id, p)?))
     }
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -44,6 +44,8 @@ pub struct Broadcast<N> {
     pessimissm_factor: usize,
     /// The proofs we have received via `Echo` messages, by sender ID.
     echos: BTreeMap<N, Proof<Vec<u8>>>,
+    /// The hashes we have received via `EchoHash` messages, by sender ID.
+    echo_hashes: BTreeMap<N, Digest>,
     /// The root hashes we received via `Ready` messages, by sender ID.
     readys: BTreeMap<N, Vec<u8>>,
 }
@@ -104,6 +106,7 @@ impl<N: NodeIdT> Broadcast<N> {
             decided: false,
             pessimissm_factor: g,
             echos: BTreeMap::new(),
+            echo_hashes: BTreeMap::new(),
             readys: BTreeMap::new(),
 
         })
@@ -270,7 +273,12 @@ impl<N: NodeIdT> Broadcast<N> {
         // Save the proof for reconstructing the tree later.
         self.echos.insert(sender_id.clone(), p);
 
-        if self.ready_sent || self.count_echos(&hash) < self.netinfo.num_correct() {
+        // Upon receiving `N - 2f` `Echo`s with this root hash, multicast `CanDecode`
+        if !self.can_decode_sent && self.count_echos(&hash) >= self.netinfo.num_correct() - self.netinfo.num_faulty() {
+            return self.send_can_decode(&hash);
+        }
+
+        if self.ready_sent || self.count_echos(&hash) + self.count_echo_hashes(&hash) < self.netinfo.num_correct() {
             return self.compute_output(&hash);
         }
 
@@ -472,6 +480,14 @@ impl<N: NodeIdT> Broadcast<N> {
         self.echos
             .values()
             .filter(|p| p.root_hash() == hash)
+            .count()
+    }
+
+    /// Returns the number of nodes that have sent us an `EchoHash` message with this hash.
+    fn count_echo_hashes(&self, hash: &Digest) -> usize {
+        self.echo_hashes
+            .values()
+            .filter(|h| *h == hash)
             .count()
     }
 
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -92,6 +92,7 @@ impl<N: NodeIdT> Broadcast<N> {
             decided: false,
             echos: BTreeMap::new(),
             readys: BTreeMap::new(),
+
         })
     }
 
@@ -123,6 +124,8 @@ impl<N: NodeIdT> Broadcast<N> {
             Message::Value(p) => self.handle_value(sender_id, p),
             Message::Echo(p) => self.handle_echo(sender_id, p),
             Message::Ready(ref hash) => self.handle_ready(sender_id, hash),
+            Message::CanDecode(ref hash) => self.handle_can_decode(sender_id, hash),
+            Message::EchoHash(ref hash) => self.handle_echo_hash(sender_id, hash),
         }
     }
 
@@ -286,6 +289,15 @@ impl<N: NodeIdT> Broadcast<N> {
         }
         Ok(step.join(self.compute_output(hash)?))
     }
+
+    fn handle_can_decode(&mut self, sender_id: &N, hash: &Digest) -> Result<Step<N>> {
+        unimplemented!()
+    }
+
+    fn handle_echo_hash(&mut self, sender_id: &N, hash: &Digest) -> Result<Step<N>> {
+        unimplemented!()
+    }
+
 
     /// Sends an `Echo` message and handles it. Does nothing if we are only an observer.
     fn send_echo(&mut self, p: Proof<Vec<u8>>) -> Result<Step<N>> {
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -495,9 +495,12 @@ impl<N: NodeIdT> Broadcast<N> {
         let mut step = Step::default();
 
         for id in self.netinfo.all_ids() {
-            if let Some(EchoContent::Hash(_)) = self.echos.get(id) {
-                let msg = Target::Node(id.clone()).message(can_decode_msg.clone());
-                step.messages.push(msg);
+            match self.echos.get(id) {
+                Some(EchoContent::Hash(_)) | None => {
+                    let msg = Target::Node(id.clone()).message(can_decode_msg.clone());
+                    step.messages.push(msg);
+                }
+                _ => ()
             }
         }
         let our_id = &self.our_id().clone();
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -443,21 +443,20 @@ impl<N: NodeIdT> Broadcast<N> {
         let echo_msg = Message::Echo(p);
         let mut step = Step::default();
 
-        if let Some(senders) = self.can_decodes.get(hash) {
-            // Remaining node ids to the right of our_id
-            // after arranging all node ids in a circular list.
-            let right = self
-                .netinfo
-                .all_ids()
-                .cycle()
-                .skip_while(|x| *x != self.our_id())
-                .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
-                .take_while(|x| *x != self.our_id());
-            let msgs = right
-                .filter(|id| !senders.contains(id))
-                .map(|id| Target::Node(id.clone()).message(echo_msg.clone()));
-            step.messages.extend(msgs);
-        }
+        let senders = self.can_decodes.get(hash);
+        // Remaining node ids to the right of our_id
+        // after arranging all node ids in a circular list.
+        let right = self
+            .netinfo
+            .all_ids()
+            .cycle()
+            .skip_while(|x| *x != self.our_id())
+            .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
+            .take_while(|x| *x != self.our_id());
+        let msgs = right
+            .filter(|id| senders.map_or(true, |s| !s.contains(id)))
+            .map(|id| Target::Node(id.clone()).message(echo_msg.clone()));
+        step.messages.extend(msgs);
         Ok(step)
     }
 
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -17,7 +17,7 @@ use crate::fault_log::Fault;
 use crate::{ConsensusProtocol, NetworkInfo, NodeIdT, Target};
 
 type RseResult<T> = result::Result<T, rse::Error>;
-
+type EchoMessage = Option<Proof<Vec<u8>>>;
 /// Broadcast algorithm instance.
 #[derive(Debug)]
 pub struct Broadcast<N> {
@@ -42,7 +42,7 @@ pub struct Broadcast<N> {
     /// /// Estimate as to how many nodes we think are faulty.
     fault_estimate: usize,
     /// The hashes and proofs we have received via `Echo` and `EchoHash` messages, by sender ID.
-    echos: BTreeMap<N, (Digest, Option<Proof<Vec<u8>>>)>,
+    echos: BTreeMap<N, (Digest, EchoMessage)>,
     /// The hashes we have received from nodes via `CanDecode` messages, by hash.
     /// A node can receive conflicting `CanDecode`s from the same node.
     can_decodes: BTreeMap<Digest, BTreeSet<N>>,
@@ -261,6 +261,8 @@ impl<N: NodeIdT> Broadcast<N> {
             }
         }
 
+        // Case where we have received an earlier `EchoHash`
+        // messages from sender_id with different root_hash.
         if let Some((hash, None)) = self.echos.get(sender_id) {
             if hash != p.root_hash() {
                 return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleEchos).into());
@@ -307,11 +309,10 @@ impl<N: NodeIdT> Broadcast<N> {
         }
 
         // If the sender has already sent an `Echo` for the same hash, ignore.
-        if let Some((_, Some(proof))) = self.echos.get(sender_id) {
-            if proof.root_hash() == hash {
+        if let Some((_, Some(p))) = self.echos.get(sender_id) {
+            if p.root_hash() == hash {
                 return Ok(Step::default());
             } else {
-                // Should be a new error? Conflicting Echo and EchoHashes.
                 return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleEchoHashes).into());
             }
         }
@@ -331,7 +332,7 @@ impl<N: NodeIdT> Broadcast<N> {
         if let Some(nodes) = self.can_decodes.get(hash) {
             if nodes.contains(sender_id) {
                 warn!(
-                    "Node {:?} received CanDecode({:?}) multiple times from {:?}.",
+                    "Node {:?} received same CanDecode({:?}) multiple times from {:?}.",
                     self.our_id(),
                     hash,
                     sender_id,
@@ -377,7 +378,6 @@ impl<N: NodeIdT> Broadcast<N> {
         // Upon receiving 2f + 1 matching Ready(h) messages, send full
         // `Echo` message to every node who hasn't sent us a `CanDecode`
         if self.count_readys(hash) == 2 * self.netinfo.num_faulty() + 1 {
-            // `send_echo` function should take care of which nodes to send the message
             step.extend(self.send_echo_remaining()?);
         }
 
@@ -415,13 +415,15 @@ impl<N: NodeIdT> Broadcast<N> {
             return Ok(Step::default());
         }
 
+        // Simply return as we can decode the message anyway as we have N - 2f shards.
         if !self.echos.contains_key(self.our_id()) {
             return Ok(Step::default());
         }
+
         if let Some((_, None)) = self.echos.get(self.our_id()) {
             return Ok(Step::default());
         }
-        let (_, p) = self.echos.get(self.our_id()).unwrap();
+        let (_, p) = &self.echos[self.our_id()];
         let p = p.clone().unwrap();
         let echo_msg = Message::Echo(p.clone());
         let mut step = Step::default();
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -40,10 +40,8 @@ pub struct Broadcast<N> {
     decided: bool,
     /// /// Estimate as to how many nodes we think are faulty.
     fault_estimate: usize,
-    /// The proofs we have received via `Echo` messages, by sender ID.
-    echos: BTreeMap<N, Proof<Vec<u8>>>,
-    /// The hashes we have received via `EchoHash` messages, by sender ID.
-    echo_hashes: BTreeMap<N, Digest>,
+    /// The hashes and proofs we have received via `Echo` and `EchoHash` messages, by sender ID.
+    echos: BTreeMap<N, (Digest, Option<Proof<Vec<u8>>>)>,
     /// The hashes we have received via `CanDecode` messages, by sender ID.
     can_decodes: BTreeMap<N, Digest>,
     /// The root hashes we received via `Ready` messages, by sender ID.
@@ -105,7 +103,6 @@ impl<N: NodeIdT> Broadcast<N> {
             decided: false,
             fault_estimate: g,
             echos: BTreeMap::new(),
-            echo_hashes: BTreeMap::new(),
             can_decodes: BTreeMap::new(),
             readys: BTreeMap::new(),
 
@@ -219,17 +216,21 @@ impl<N: NodeIdT> Broadcast<N> {
             let fault_kind = FaultKind::ReceivedValueFromNonProposer;
             return Ok(Fault::new(sender_id.clone(), fault_kind).into());
         }
+        // Don't need the first if statement. Check!
         if self.echos.contains_key(self.our_id()) {
-            if self.echos.get(self.our_id()) == Some(&p) {
-                warn!(
-                    "Node {:?} received Value({:?}) multiple times from {:?}.",
-                    self.our_id(),
-                    HexProof(&p),
-                    sender_id
-                );
-                return Ok(Step::default());
-            } else {
-                return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleValues).into());
+            if let Some((_, Some(proof))) = self.echos.get(self.our_id()) {
+                if *proof == p {
+                    warn!(
+                        "Node {:?} received Value({:?}) multiple times from {:?}.",
+                        self.our_id(),
+                        HexProof(&p),
+                        sender_id
+                    );
+                    return Ok(Step::default());
+                }
+                else {
+                    return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleValues).into());
+                }
             }
         }
 
@@ -240,7 +241,7 @@ impl<N: NodeIdT> Broadcast<N> {
 
         // Insert the value received into echos map before sending `Echo` and `EchoHash` to peers.
         // Will get warning in `handle_echo` call for our_id.
-        self.echos.insert(sender_id.clone(), p.clone());
+        self.echos.insert(sender_id.clone(), (p.root_hash().clone(), Some(p.clone())));
 
         // Send the proof in an `Echo` message to left nodes
         // and `EchoHash` message to right nodes and handle the response.
@@ -252,7 +253,7 @@ impl<N: NodeIdT> Broadcast<N> {
     /// Handles a received `Echo` message.
     fn handle_echo(&mut self, sender_id: &N, p: Proof<Vec<u8>>) -> Result<Step<N>> {
         // If the sender has already sent `Echo`, ignore.
-        if let Some(old_p) = self.echos.get(sender_id) {
+        if let Some((_, Some(old_p))) = self.echos.get(sender_id) {
             if *old_p == p {
                 warn!(
                     "Node {:?} received Echo({:?}) multiple times from {:?}.",
@@ -274,14 +275,14 @@ impl<N: NodeIdT> Broadcast<N> {
         let hash = *p.root_hash();
 
         // Save the proof for reconstructing the tree later.
-        self.echos.insert(sender_id.clone(), p);
+        self.echos.insert(sender_id.clone(), (hash.clone(), Some(p.clone())));
 
         // Upon receiving `N - 2f` `Echo`s with this root hash, send `CanDecode`
-        if !self.can_decode_sent && self.count_echos(&hash) >= self.netinfo.num_correct() - self.netinfo.num_faulty() {
+        if !self.can_decode_sent && self.count_echos_full(&hash) >= self.netinfo.num_correct() - self.netinfo.num_faulty() {
             return self.send_can_decode(&hash);
         }
 
-        if self.ready_sent || self.count_echos(&hash) + self.count_echo_hashes(&hash) < self.netinfo.num_correct() {
+        if self.ready_sent || self.count_echos(&hash) < self.netinfo.num_correct() {
             return self.compute_output(&hash);
         }
 
@@ -291,7 +292,8 @@ impl<N: NodeIdT> Broadcast<N> {
 
     fn handle_echo_hash(&mut self, sender_id: &N, hash: &Digest) -> Result<Step<N>> {
         // If the sender has already sent `EchoHash`, ignore.
-        if let Some(old_hash) = self.echo_hashes.get(sender_id) {
+        // Should there be an error case for sending an `EchoHash` after an `Echo` (2nd element not None)
+        if let Some((old_hash, _)) = self.echos.get(sender_id) {
             if old_hash == hash {
                 warn!(
                     "Node {:?} received EchoHash({:?}) multiple times from {:?}.",
@@ -305,9 +307,9 @@ impl<N: NodeIdT> Broadcast<N> {
             }
         }
         // Save the hash for counting later.
-        self.echo_hashes.insert(sender_id.clone(), *hash);
+        self.echos.insert(sender_id.clone(), (*hash, None));
 
-        if self.ready_sent || self.count_echos(&hash) + self.count_echo_hashes(&hash) < self.netinfo.num_correct() {
+        if self.ready_sent || self.count_echos(&hash) < self.netinfo.num_correct() {
             return self.compute_output(&hash);
         }
         // Upon receiving `N - f` `Echo`s with this root hash, multicast `Ready`.
@@ -378,7 +380,9 @@ impl<N: NodeIdT> Broadcast<N> {
             return Ok(Step::default());
         }
         // Caller should make sure message is inserted. unwrap should never panic.
-        let p = self.echos.get(self.our_id()).unwrap().clone();
+        // Could also have another error type if it panics.
+        let (_, p) = self.echos.get(self.our_id()).unwrap();
+        let p = p.clone().unwrap();
         let echo_msg = Message::Echo(p.clone());
         let mut step = Step::default();
         // TODO: iterator not correctly describing left of our_id. Use cycle method on
@@ -389,7 +393,7 @@ impl<N: NodeIdT> Broadcast<N> {
             step.messages.push(msg);
         }
         let our_id = &self.our_id().clone();
-        Ok(step.join(self.handle_echo(our_id, p)?))
+        Ok(step.join(self.handle_echo(our_id, p.clone())?))
     }
 
     /// Sends `Echo` message to remaining nodes who haven't sent `CanDecode`
@@ -399,7 +403,9 @@ impl<N: NodeIdT> Broadcast<N> {
             return Ok(Step::default());
         }
         // Caller should make sure message is inserted. unwrap should never panic.
-        let p = self.echos.get(self.our_id()).unwrap().clone();
+        // Could also have another error type for this case.
+        let (_, p) = self.echos.get(self.our_id()).unwrap();
+        let p = p.clone().unwrap();
         let echo_msg = Message::Echo(p.clone());
         let mut step = Step::default();
         // TODO: iterator not correctly describing right of our_id. Use cycle method on
@@ -413,7 +419,7 @@ impl<N: NodeIdT> Broadcast<N> {
         }
         // Is this required?!
         let our_id = &self.our_id().clone();
-        Ok(step.join(self.handle_echo(our_id, p)?))
+        Ok(step.join(self.handle_echo(our_id, p.clone())?))
     }
 
     /// Sends an `EchoHash` message and handles it. Does nothing if we are only an observer.
@@ -482,7 +488,9 @@ impl<N: NodeIdT> Broadcast<N> {
             .netinfo
             .all_ids()
             .map(|id| {
-                self.echos.get(id).and_then(|p| {
+                self.echos.get(id)
+                    .and_then(|(_, x)| x.as_ref())
+                    .and_then(|p| {
                     if p.root_hash() == hash {
                         Some(p.value().clone().into_boxed_slice())
                     } else {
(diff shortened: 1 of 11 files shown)


src/broadcast/broadcast.rs
```diff
@@ -26,16 +26,30 @@ pub struct Broadcast<N> {
...
-    /// Whether we have already multicast `Echo`.
-    echo_sent: bool,
+    /// Whether we have already multicast `Echo` to our left nodes.
+    echo_sent_partial: bool,
+    /// Whether we have already multicast `Echo` to our right nodes who haven't sent `CanDecode`.
+    echo_sent_full: bool,
     /// Whether we have already multicast `Ready`.
     ready_sent: bool,
+    /// Whether we have already multicast `EchoHash` to the right nodes.
+    echo_hash_sent: bool,
```
afck: That should always be identical to `echo_sent_partial`, I think?

src/broadcast/broadcast.rs
```diff
@@ -26,16 +26,30 @@ pub struct Broadcast<N> {
...
+    /// Whether we have already multicast `CanDecode`.
+    can_decode_sent: bool,
     /// Whether we have already output a value.
     decided: bool,
+    /// Pessimissm factor
+    pessimissm_factor: usize,
```
afck: Typo: `pessimism`. (Not sure if it's a good word anyway… `redundancy_factor`? `fault_estimate`?)

src/broadcast/broadcast.rs
```diff
@@ -218,9 +243,13 @@ impl<N: NodeIdT> Broadcast<N> {
         if !self.validate_proof(&p, &self.our_id()) {
             return Ok(Fault::new(sender_id.clone(), FaultKind::InvalidProof).into());
         }
-
-        // Otherwise multicast the proof in an `Echo` message, and handle it ourselves.
-        self.send_echo(p)
+        // Store value received to send back at later stage.
+        self.value_message = p.clone();
+        // Multicast the proof in an `Echo` message to left nodes
```
afck: Let's reserve the word "multicast" for instances where messages are actually sent to _all_ other validators.

src/broadcast/broadcast.rs
```diff
@@ -250,14 +279,66 @@ impl<N: NodeIdT> Broadcast<N> {
...
+            if old_hash == hash {
+                warn!(
+                    "Node {:?} received CanDecode({:?}) multiple times from {:?}.",
+                    self.our_id(),
+                    hash,
+                    sender_id,
+                );
+                return Ok(Step::default());
+            } else {
+                return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleCanDecodes).into());
```
afck: I don't think we can attribute this fault to the sender; it could also be due to a faulty proposer: If the proposer sent one kind of matching `Value`s to one set of _N - f_ correct nodes, and another to another set of _N - f_ correct nodes, then poor `sender_id` could have received two sets of _N - f_ `Echo`s with different root hashes. I think in that case they actually have to send both `CanDecode`s, and we have to keep track of both! (Please double-check the proof; I think it only works that way.) So we'd need to make `can_decodes` a `BTreeSet<(N, Digest)>`, I guess?

src/broadcast/broadcast.rs
```diff
@@ -26,16 +26,30 @@ pub struct Broadcast<N> {
...
+    /// Original value message received from proposer.
+    value_message: Proof<Vec<u8>>,
     /// If we are the proposer: whether we have already sent the `Value` messages with the shards.
     value_sent: bool,
-    /// Whether we have already multicast `Echo`.
-    echo_sent: bool,
+    /// Whether we have already multicast `Echo` to our left nodes.
+    echo_sent_partial: bool,
+    /// Whether we have already multicast `Echo` to our right nodes who haven't sent `CanDecode`.
+    echo_sent_full: bool,
```
afck: Isn't that equivalent to `self.echos.contains_key(self.our_id())`?

src/broadcast/broadcast.rs
```diff
@@ -26,16 +26,30 @@ pub struct Broadcast<N> {
...
     /// Whether we have already multicast `Ready`.
     ready_sent: bool,
+    /// Whether we have already multicast `EchoHash` to the right nodes.
+    echo_hash_sent: bool,
+    /// Whether we have already multicast `CanDecode`.
+    can_decode_sent: bool,
```
afck: Is that equivalent to `self.can_decodes.contains_key(self.our_id())`?

src/broadcast/broadcast.rs
```diff
@@ -284,21 +365,88 @@ impl<N: NodeIdT> Broadcast<N> {
...
+        // Send `Echo` to remaining right nodes who haven't sent us `CanDecode`.
+        else {
+            self.echo_sent_full = true;
+            for id in right {
+                if !self.can_decodes.contains_key(id) {
+                    let msg = Target::Node(id.clone()).message(echo_msg.clone());
+                    step.messages.push(msg);
+                }
+            }
+        }
```
afck: Instead of making this distinction, it might be cleaner to just have two methods: `send_echo_to_left` and `send_echo_to_rest` or something (maybe with better names :sweat_smile:). The former would be called in `handle_value` and the latter in `handle_ready`. Neither of them needs an argument: Currently `self.value_message` is cloned and passed in at both call sites.

src/broadcast/broadcast.rs
```diff
@@ -284,21 +365,88 @@ impl<N: NodeIdT> Broadcast<N> {
...
+        self.can_decode_sent = true;
+        if !self.netinfo.is_validator() {
+            return Ok(Step::default());
+        }
+
+        let can_decode_msg = Message::CanDecode(*hash);
+        let mut step = Step::default();
+
+        for id in self.netinfo.all_ids() {
+            if self.echos.contains_key(id) {
```
afck: Should this be negated? `CanDecode` needs to go to everyone who _hasn't_ sent us their `Echo` yet.
(Also, should we filter out ourselves here? Possibly even add a `other_ids` method to `NetworkInfo`?)

src/broadcast/broadcast.rs
```diff
@@ -250,14 +279,66 @@ impl<N: NodeIdT> Broadcast<N> {
...
+                );
+                return Ok(Step::default());
+            } else {
+                return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleEchoHashes).into());
+            }
+        }
+        // Save the hash for counting later.
+        self.echo_hashes.insert(sender_id.clone(), *hash);
+
+        if self.ready_sent || self.count_echos(&hash) + self.count_echo_hashes(&hash) < self.netinfo.num_correct() {
```
afck: We can't just take this sum (sorry, I formulated this the wrong way in the issue!), because then we'd count a node that has sent us both an `Echo` and `EchoHash` twice!
I wonder whether it would be easier to just count every `Echo` as an `EchoHash` implicitly, too, i.e. also insert their hash into `echo_hashes`?

src/broadcast/broadcast.rs
```diff
@@ -26,16 +26,30 @@ pub struct Broadcast<N> {
     proposer_id: N,
     /// The Reed-Solomon erasure coding configuration.
     coding: Coding,
+    /// Original value message received from proposer.
+    value_message: Proof<Vec<u8>>,
```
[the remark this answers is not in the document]
pawanjay176: Yes makes sense. 
I was thinking of removing the `value_message` field and simply adding the message to the `echos` map with `our_id` as key and fetching from the map later. 
Also, currently, `our_id` node's shard never factors into the `decode_from_shards` function. So doing it this way might be better right?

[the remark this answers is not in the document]
pawanjay176: Yes makes sense. 
However, if the `send_echo` function sends the `value_message` to itself, we could also simply get the message from the `echos` map right? We won't need an extra variable in the struct that way.

src/broadcast/broadcast.rs
pawanjay176: Yup. Don't need `echo_sent_partial`.

src/broadcast/broadcast.rs
pawanjay176: Yup I agree its pretty messy right now :sweat_smile: 
Will create separate functions as you suggested :)

src/broadcast/broadcast.rs
pawanjay176: Yup you are right. Will fix this.

src/broadcast/broadcast.rs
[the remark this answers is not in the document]
afck: You're right, that's even better!

src/broadcast/broadcast.rs
pawanjay176: `fault_estimate` seems best suited to me.

src/broadcast/broadcast.rs
```diff
@@ -26,16 +26,30 @@ pub struct Broadcast<N> {
...
     /// If we are the proposer: whether we have already sent the `Value` messages with the shards.
     value_sent: bool,
-    /// Whether we have already multicast `Echo`.
-    echo_sent: bool,
+    /// Whether we have already multicast `Echo` to our left nodes.
+    echo_sent_partial: bool,
```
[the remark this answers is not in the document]
pawanjay176: Yes that is correct. Isn't `echo_sent` in the original code also `true` iff `echos.contains_key(self.our_id()) == true`?

src/broadcast/broadcast.rs
pawanjay176: Yes you are right. It also serves no purpose in the rest of the code as of now.

src/broadcast/broadcast.rs
pawanjay176: Wouldn't sending to all other validators be broadcast?!

src/broadcast/broadcast.rs
pawanjay176: We could take the length of the set union of `echo` and `echo_hash` keys where the value is `hash`. Counting every `Echo` as an `EchoHash` implicitly and then counting just the `EchoHash`s would work too but I think it would lead to some confusing code.

src/broadcast/broadcast.rs
pawanjay176: Yes it is equivalent. Checking the `BTreeMap` would have worked for all these `bool`s in the original code as well right?

src/broadcast/broadcast.rs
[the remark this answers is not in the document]
afck: Yes, I think you're right. Anyway, I'm always in favor of keeping redundancy out of data structures, to avoid all kinds of bugs where equivalent fields go out of sync…

src/broadcast/broadcast.rs
afck: I think the Honey Badger BFT paper uses the term "multicast" to avoid confusion with the "Reliable Broadcast Protocol", and we adopted that. Since we call the latter just "Broadcast", we should be even more careful.

src/broadcast/broadcast.rs
afck: That's true; but let's avoid actually allocating the set, and just count the number of node IDs that are present in either.
Alternatively, we could replace the two maps with one:
```rust
echo_hashes: BTreeMap<N, (Digest, Option<Proof<Vec<u8>>>)>,
```

src/broadcast/broadcast.rs
pawanjay176: Makes sense. If a faulty proposer sends the conflicting messages `h0` and `h1` to the honest nodes such that neither set of honest nodes get `N-f` `Echo`s, then the protocol simply would stall. If one set gets `N-f` `Echos`, they propagate their `Ready`s and eventually, all honest nodes will be able to decode the majority `h_{i}`. The honest nodes have no way of knowing which `h_{i}`  eventually gets the majority votes (if there is a majority). Hence, they need to advertise their ability to decode both the `h_{i}`s to all the correct nodes to ensure that all the honest nodes can agree upon a value if possible.
In the first case(stalling) , does the protocol terminate somehow? We can't have timeouts here because of the asynchronous nature of the protocol I assume?

src/broadcast/broadcast.rs
pawanjay176: Single map idea is awesome. Thanks!

src/broadcast/broadcast.rs
afck: You're exactly right, and the protocol can stall if the sender is malicious: if they don't send anything, it's guaranteed to stall, and if they send different messages to different groups of nodes, either all of them agree on one of those messages, or none of them terminates.

Inside Honey Badger, it's fine and expected if some Broadcast instances stall: A batch can be produced when at least ⅔ of them have terminated in the current epoch, which is guaranteed if at least ⅔ of nodes are correct.

src/broadcast/broadcast.rs
```diff
@@ -362,44 +362,55 @@ impl<N: NodeIdT> Broadcast<N> {
...
-    /// Sends an `Echo` message and handles it. Does nothing if we are only an observer.
-    fn send_echo(&mut self, p: Proof<Vec<u8>>) -> Result<Step<N>> {
-       if !self.netinfo.is_validator() {
+    /// Sends `Echo` message to all left nodes.
+    fn send_echo_left(&mut self) -> Result<Step<N>> {
+        if !self.netinfo.is_validator() {
             return Ok(Step::default());
         }
+        // Caller should make sure message is inserted. unwrap should never panic.
+        let p = self.echos.get(self.our_id()).unwrap().clone();
```
afck: But at this point we haven't inserted it yet! That only happens below (`self.handle_echo(our_id, p)`).
I actually think `send_echo_left` needs to keep its `p` argument.

src/broadcast/broadcast.rs
```diff
@@ -362,44 +362,55 @@ impl<N: NodeIdT> Broadcast<N> {
...
-            for id in right {
+    /// Sends `Echo` message to remaining nodes who haven't sent `CanDecode`
+    fn send_echo_remaining(&mut self) -> Result<Step<N>> {
+        // No need for echo_sent_partial check
+        self.echo_sent_full = true;
+        if !self.netinfo.is_validator() {
+            return Ok(Step::default());
+        }
+        // Caller should make sure message is inserted. unwrap should never panic.
+        let p = self.echos.get(self.our_id()).unwrap().clone();
```
afck: That's not true: The proposer might not have sent us a value (maliciously) or the message might be delayed. I _think_ it's correct to just return here if we don't have a value, i.e. if that `echos` entry doesn't exist: The fact that we arrived here means that the algorithm is going to be able to terminate even without our own `Echo` anyway, doesn't it?

src/broadcast/broadcast.rs
```diff
@@ -362,44 +362,55 @@ impl<N: NodeIdT> Broadcast<N> {
...
-        else {
-            self.echo_sent_full = true;
-            for id in right {
-                if !self.can_decodes.contains_key(id) {
-                    let msg = Target::Node(id.clone()).message(echo_msg.clone());
-                    step.messages.push(msg);
-                }
-            }
-        }
+        // Is this required?!
```
afck: No, that can be removed: We handled our own `Echo` in `send_echo_left` already.

src/broadcast/broadcast.rs
```diff
@@ -219,17 +216,21 @@ impl<N: NodeIdT> Broadcast<N> {
             let fault_kind = FaultKind::ReceivedValueFromNonProposer;
             return Ok(Fault::new(sender_id.clone(), fault_kind).into());
         }
+        // Don't need the first if statement. Check!
```
afck: Yes, the next line is redundant.

src/broadcast/broadcast.rs
```diff
@@ -252,7 +253,7 @@ impl<N: NodeIdT> Broadcast<N> {
     /// Handles a received `Echo` message.
     fn handle_echo(&mut self, sender_id: &N, p: Proof<Vec<u8>>) -> Result<Step<N>> {
         // If the sender has already sent `Echo`, ignore.
-        if let Some(old_p) = self.echos.get(sender_id) {
+        if let Some((_, Some(old_p))) = self.echos.get(sender_id) {
```
afck: Actually, even if this is `Some((hash, None))`, we should probably check that `hash` agrees with `p.root_hash()`: If it doesn't, the sender must be faulty, too.

src/broadcast/broadcast.rs
```diff
@@ -274,14 +275,14 @@ impl<N: NodeIdT> Broadcast<N> {
         let hash = *p.root_hash();
 
         // Save the proof for reconstructing the tree later.
-        self.echos.insert(sender_id.clone(), p);
+        self.echos.insert(sender_id.clone(), (hash.clone(), Some(p.clone())));
 
         // Upon receiving `N - 2f` `Echo`s with this root hash, send `CanDecode`
-        if !self.can_decode_sent && self.count_echos(&hash) >= self.netinfo.num_correct() - self.netinfo.num_faulty() {
+        if !self.can_decode_sent && self.count_echos_full(&hash) >= self.netinfo.num_correct() - self.netinfo.num_faulty() {
```
afck: Shouldn't this be `self.netinfo.num_correct() - 2 * self.netinfo.num_faulty()`? Or, even better, `self.coding.data_shard_num()`? (That makes the meaning more clear: We have enough shards to decode the data.)

src/broadcast/broadcast.rs
```diff
@@ -291,7 +292,8 @@ impl<N: NodeIdT> Broadcast<N> {
 
     fn handle_echo_hash(&mut self, sender_id: &N, hash: &Digest) -> Result<Step<N>> {
         // If the sender has already sent `EchoHash`, ignore.
-        if let Some(old_hash) = self.echo_hashes.get(sender_id) {
+        // Should there be an error case for sending an `EchoHash` after an `Echo` (2nd element not None)
```
afck: It's debatable whether we should expect messages from a single peer to arrive in the right order. The algorithms should work fine if we don't. 
For now, I wouldn't consider this a fault, as long as the hashes match.
However, we need to make sure we don't insert `None` again below, if we already have a `Some(_)`!

src/broadcast/broadcast.rs
```diff
@@ -378,7 +380,9 @@ impl<N: NodeIdT> Broadcast<N> {
             return Ok(Step::default());
         }
         // Caller should make sure message is inserted. unwrap should never panic.
-        let p = self.echos.get(self.our_id()).unwrap().clone();
+        // Could also have another error type if it panics.
+        let (_, p) = self.echos.get(self.our_id()).unwrap();
+        let p = p.clone().unwrap();
```
afck: We should not assume that we have our own `Echo` here. (See my earlier comment, which probably became outdated.)

src/broadcast/broadcast.rs
```diff
@@ -399,7 +403,9 @@ impl<N: NodeIdT> Broadcast<N> {
             return Ok(Step::default());
         }
         // Caller should make sure message is inserted. unwrap should never panic.
-        let p = self.echos.get(self.our_id()).unwrap().clone();
+        // Could also have another error type for this case.
+        let (_, p) = self.echos.get(self.our_id()).unwrap();
+        let p = p.clone().unwrap();
```
afck: Same here: We can't expect that.

src/broadcast/broadcast.rs
```diff
@@ -413,7 +419,7 @@ impl<N: NodeIdT> Broadcast<N> {
         }
         // Is this required?!
         let our_id = &self.our_id().clone();
-        Ok(step.join(self.handle_echo(our_id, p)?))
+        Ok(step.join(self.handle_echo(our_id, p.clone())?))
```
afck: Why do we need to clone that?

src/broadcast/broadcast.rs
```diff
@@ -43,7 +44,8 @@ pub struct Broadcast<N> {
     /// The hashes and proofs we have received via `Echo` and `EchoHash` messages, by sender ID.
     echos: BTreeMap<N, (Digest, Option<Proof<Vec<u8>>>)>,
     /// The hashes we have received via `CanDecode` messages, by sender ID.
-    can_decodes: BTreeMap<N, Digest>,
+    /// A node can receive conflicting `CanDecode`s from the same node.
+    can_decodes: BTreeSet<(N, Digest)>,
```
afck: It might make sense to make this a `BTreeMap<Digest, BTreeSet<N>>` instead: We probably always need to access it by hash? That way we don't need to create those tuples, and clone the root hash inside the loop.

src/broadcast/broadcast.rs
```diff
@@ -380,9 +380,12 @@ impl<N: NodeIdT> Broadcast<N> {
...
         let mut step = Step::default();
-        // TODO: iterator not correctly describing left of our_id. Use cycle method on
-        // iterator to correctly specify.
-        let left = self.netinfo.all_ids_except(self.proposer_id()).take(self.netinfo.num_correct()+self.fault_estimate);
+        // `N - 2f + g` node ids to the left of our_id (including our_id)
+        // after arranging all node ids in a circular list.
+        let left = self.netinfo.all_ids_except(self.proposer_id())
+                               .cycle()
+                               .skip_while(|x| *x == self.our_id())
+                               .take(self.netinfo.num_correct() - self.netinfo.num_faulty()+self.fault_estimate);
```
afck: `.skip_while(|x| *x == self.our_id())` would always skip 0 elements except if we're the first entry. You probably wanted to say `!=` and then skip one more?

But I think we can actually count the proposer towards the _N - 2f + g_, since it definitely does have the data.
The same goes for ourselves. So I'd do something like:
```rust
let left = self.netinfo.all_ids()
    .cycle()
    .skip_while(|x| *x != self.our_id())
    .take(self.netinfo.num_correct() - self.netinfo.num_faulty()+self.fault_estimate)
    .skip(1);
```
…and maybe in `send_echo` filter out the proposer?

src/broadcast/broadcast.rs
```diff
@@ -403,9 +406,13 @@ impl<N: NodeIdT> Broadcast<N> {
...
-        // TODO: iterator not correctly describing right of our_id. Use cycle method on
-        // iterator to correctly specify.
-        let right = self.netinfo.all_ids_except(self.proposer_id()).skip(self.netinfo.num_correct()+self.fault_estimate);
+        // Remaining `2f - g` node ids to the right of our_id (including our_id)
+        // after arranging all node ids in a circular list.
+        let right = self.netinfo.all_ids_except(self.proposer_id())
+                               .cycle()
+                               .skip_while(|x| *x == self.our_id())
+                               .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
+                               .take_while(|x| *x == self.our_id());
```
afck: This also looks wrong to me. (Also below.)

src/broadcast/broadcast.rs
pawanjay176: Yes it should be `!=` and then skip one for ourself. I didn't get why we would count the proposer towards the N - 2f + g though. Seems wasteful since he already has the data.

Sorry for all the unchecked code btw. I was changing laptops so committed everything I had without checking. Forgot that pull request would get updated. :sweat_smile:

src/broadcast/broadcast.rs
pawanjay176: Yes will fix this.

src/broadcast/broadcast.rs
pawanjay176: I had inserted the message in the `handle_value` function before calling `send_echo_left` . 
```
// Insert the value received into echos map before sending `Echo` and `EchoHash` to peers.
// Will get warning in `handle_echo` call for our_id.
self.echos.insert(sender_id.clone(), (p.root_hash().clone(), Some(p.clone())));

// Send the proof in an `Echo` message to left nodes
// and `EchoHash` message to right nodes and handle the response.
let echo_steps = self.send_echo_left()?;
let echo_hash_steps = self.send_echo_hash(p.root_hash())?;
Ok(echo_steps.join(echo_hash_steps))
```

But yeah, passing the `p` argument to `send_echo_left` probably makes more sense.

src/broadcast/broadcast.rs
pawanjay176: You are right. `self.coding.data_shard_num()` is better.
Will fix this.

src/broadcast/broadcast.rs
pawanjay176: Yeah we should be able to terminate since we wouldn't have reached this function if we don't have atleast `N-2f` shards. Returning `Step::default()` should be okay I think.

src/broadcast/broadcast.rs
pawanjay176: You mean same sender sent different `hash` with `Echo` and `EchoHash` messages. Makes sense.

src/broadcast/broadcast.rs
pawanjay176: Will pass `p` as an argument here.

src/broadcast/broadcast.rs
pawanjay176: Sorry, will fix this.

src/broadcast/broadcast.rs
```diff
@@ -261,6 +261,8 @@ impl<N: NodeIdT> Broadcast<N> {
             }
         }
 
+        // Case where we have received an earlier `EchoHash`
+        // messages from sender_id with different root_hash.
```
afck: Typo: "message" (singular)

src/broadcast/broadcast.rs
```diff
@@ -42,7 +42,7 @@ pub struct Broadcast<N> {
     /// /// Estimate as to how many nodes we think are faulty.
     fault_estimate: usize,
     /// The hashes and proofs we have received via `Echo` and `EchoHash` messages, by sender ID.
-    echos: BTreeMap<N, (Digest, Option<Proof<Vec<u8>>>)>,
+    echos: BTreeMap<N, (Digest, EchoMessage)>,
```
afck: Actually, since the `Proof` already contains the root hash, the additional `Digest` is redundant:
```rust
enum EchoContent {
    Hash(Digest),
    Full(Proof<Vec<u8>>>),
}
// …
    echos: BTreeMap<N, EchoContent>,
```
(Maybe `EchoContent` could have a `hash` method for convenience.)

src/broadcast/broadcast.rs
```diff
@@ -420,6 +420,13 @@ impl<N: NodeIdT> Broadcast<N> {
             return Ok(Step::default());
         }
 
+        // Can't send `Echo` if we haven't received a `Value` message for that hash from Proposer.
+        if let Some((h, _)) = self.echos.get(self.our_id()) {
+            if h != hash {
+                return Ok(Step::default());
+            }
+        }
+
```
afck: Why access `self.echos` four times?
```rust
let p = match self.echos.get(self.our_id()) {
    None | Some((_, None)) => return Ok(Step::default()),
    Some((_, p)) if p.root_hash() != hash => return Ok(Step::default()),
    Some((_, p)) => p.clone(),
}
```
No `unwrap`s or index accesses needed! :tada:

src/broadcast/broadcast.rs
```diff
@@ -284,21 +380,130 @@ impl<N: NodeIdT> Broadcast<N> {
...
+        if let Some((_, None)) = self.echos.get(self.our_id()) {
+            return Ok(Step::default());
+        }
+        let (_, p) = &self.echos[self.our_id()];
+        let p = p.clone().unwrap();
+        let echo_msg = Message::Echo(p.clone());
```
afck: No need to clone `p` again: We only use its hash below, which is already in the `hash` variable.

src/broadcast/broadcast.rs
```diff
@@ -449,9 +443,7 @@ impl<N: NodeIdT> Broadcast<N> {
             .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
             .take_while(|x| *x != self.our_id());
         for id in right {
-            if self.can_decodes.contains_key(p.root_hash())
-                && !self.can_decodes[p.root_hash()].contains(id)
-            {
+            if self.can_decodes.contains_key(hash) && !self.can_decodes[hash].contains(id) {
```
afck: That could also be done without an index access, and only a single map lookup:
```rust
if let Some(senders) = self.can_decodes.get(hash) {
    for id in right {
        if !senders.contains(id) {
```
Also, no reason to create `right` if `self.can_decodes.get(hash)` is `None`.
Alternatively, in a functional style:
```rust
if let Some(senders) = self.can_decodes.get(hash) {
    let right = …;
    let msgs = right.filter(|id| !senders.contains(id))
        .map(|id| Target::Node(id.clone()).message(echo_msg.clone()));
    step.messages.extend(msgs);
}
```

src/broadcast/broadcast.rs
```diff
@@ -604,15 +593,15 @@ impl<N: NodeIdT> Broadcast<N> {
     fn count_echos_full(&self, hash: &Digest) -> usize {
         self.echos
             .values()
-            .map(|(_, p)| p)
+            .map(EchoContent::proof)
```
afck: With `.filter_map(EchoContent::proof)` you don't need the `flatten`.

pawanjay176:
@afck the `dynamic_honey_badger` and `queueing_honey_badger` tests fail right after node restarts in the tests. Trying to figure out why by looking into the `Subset` crate to check the lifecycle of a `Broadcast` instance.

afck:
That's weird: Do the Broadcast tests pass? If they do, but the higher-level tests fail, that shows that we need to extend the Broadcast tests (#396).

src/broadcast/broadcast.rs
```diff
@@ -248,16 +277,86 @@ impl<N: NodeIdT> Broadcast<N> {
...
+                );
+            }
+        }
+        let mut values = self
+            .can_decodes
+            .get(hash)
+            .unwrap_or(&BTreeSet::new())
+            .clone();
+        values.insert(sender_id.clone());
+        self.can_decodes.insert(*hash, values);
```
afck:
```rust
self.can_decodes
    .entry(*hash)
    .or_default()
    .insert(sender_id.clone());
```

src/broadcast/broadcast.rs
```diff
@@ -284,21 +383,120 @@ impl<N: NodeIdT> Broadcast<N> {
...
-        self.echo_sent = true;
+    /// Sends `Echo` message to all left nodes and handles it.
+    fn send_echo_left(&mut self, p: Proof<Vec<u8>>) -> Result<Step<N>> {
         if !self.netinfo.is_validator() {
             return Ok(Step::default());
         }
         let echo_msg = Message::Echo(p.clone());
-        let step: Step<_> = Target::All.message(echo_msg).into();
+        let mut step = Step::default();
+        // `N - 2f + g` node ids to the left of our_id (excluding our_id and proposer_id)
```
afck: The proposer isn't currently excluded, is it?
(I think it's better that way: That causes additional complications and should probably be a separate PR, if at all.)

pawanjay176:
@afck I think i found the problem. The current implementation uses `Target::All` to send the `Echo` and `Ready` messages and `Target::All` sends to both observer and validator nodes. As soon as we restart a node in the `DynamicHoneyBadger` tests, the restarted node becomes an observer and is not sent any message except `Target::All`. In my implementation, the nodes to send the messages to are derived from `self.netinfo.all_ids()` which don't include observer nodes and they don't receive any of `Echo`, `EchoHash` and `CanDecode` messages.

Will try to push a fix for this by tonight.

src/broadcast/broadcast.rs
```diff
@@ -28,14 +29,23 @@ pub struct Broadcast<N> {
     coding: Coding,
     /// If we are the proposer: whether we have already sent the `Value` messages with the shards.
     value_sent: bool,
-    /// Whether we have already multicast `Echo`.
+    /// Whether we have already send `Echo` left nodes and right nodes who haven't sent `CanDecode`.
     echo_sent: bool,
     /// Whether we have already multicast `Ready`.
     ready_sent: bool,
+    /// Whether we have already sent `EchoHash` to the right nodes.
```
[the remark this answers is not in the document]
pawanjay176: Will update the docs once I get the tests working :)

src/broadcast/broadcast.rs
pawanjay176: No its not :sweat_smile: Forgot to change the comment.

afck:
Right, I hadn't thought of that!
I'm almost thinking that for now, the easiest solution would be a `Target::Observers` variant, and just always sending a full `Echo` there.

Slightly more optimized: Only send a full `Echo` to the observers if you are among the _2 f + 1_ nodes to the left of the proposer.

What do you think, @vkomenda?

src/broadcast/broadcast.rs
```diff
@@ -200,8 +217,9 @@ impl<N: NodeIdT> Broadcast<N> {
             let fault_kind = FaultKind::ReceivedValueFromNonProposer;
             return Ok(Fault::new(sender_id.clone(), fault_kind).into());
         }
-        if self.echo_sent {
-            if self.echos.get(self.our_id()) == Some(&p) {
+
+        if let Some(EchoContent::Full(proof)) = self.echos.get(self.our_id()) {
+            if *proof == p {
```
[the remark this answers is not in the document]
pawanjay176: If we only have an `EchoContent::Hash`, then we can't return from the `if let` statement right? We need to update the `echos` map with the actual full proof. So it would have to be

```
match self.echos.get(self.our_id()) {
    Some(val) if val.hash() != p.root_hash() => {
        return Ok(Fault::new(sender_id.clone(), FaultKind::MultipleValues).into())
    }
    Some(EchoContent::Full(proof)) if *proof == p => {
        warn!(
            "Node {:?} received Value({:?}) multiple times from {:?}.",
            self.our_id(),
            HexProof(&p),
            sender_id
        );
        return Ok(Step::default());
    },
    _ => ()
};
```

src/broadcast/broadcast.rs
```diff
@@ -28,14 +29,23 @@ pub struct Broadcast<N> {
     coding: Coding,
     /// If we are the proposer: whether we have already sent the `Value` messages with the shards.
     value_sent: bool,
-    /// Whether we have already multicast `Echo`.
+    /// Whether we have already send `Echo` left nodes and right nodes who haven't sent `CanDecode`.
```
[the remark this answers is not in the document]
vkomenda: The comment is still missing a "to". How do you visualise "left" and "right" on a circular list, I wonder? Can you have a one-dimensional counterpart of these terms instead? Possibly predecessor and successor?

src/broadcast/broadcast.rs
```diff
@@ -284,21 +383,120 @@ impl<N: NodeIdT> Broadcast<N> {
...
+        // after arranging all node ids in a circular list.
+        let left = self
+            .netinfo
+            .all_ids()
+            .cycle()
+            .skip_while(|x| *x != self.our_id())
+            .take(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
+            .skip(1);
+        for id in left {
+            let msg = Target::Node(id.clone()).message(echo_msg.clone());
```
vkomenda: This cloning of message content makes me wish for a different `Target`. At least a target that allows sending the same content to a list of nodes.

vkomenda:
> I'm almost thinking that for now, the easiest solution would be a Target::Observers variant, and just always sending a full Echo there.

What if the new variant allowed sending the same message content to a given list of nodes _plus_ any observers not on that list? This would allow to reduce cloning of message content inside Broadcast although later that content would have to be cloned anyway when sending it to each of those nodes. But at least we would have eliminated `for`-loops over lists of nodes by replacing them with pushing a single message with that list of recipients.

With such a new message variant, we would change the sender queue slightly to marshal the content to the correct list of messages that includes any observers not known to Broadcast.

afck:
Yes, that's true, especially for broadcast, where some of the messages are relatively large. And I don't even think they necessarily have to be cloned later: The application logic can just serialize it once (instead of the current duplicated serialization), and then send those same bytes to all the recipients. No need to clone anything.

On the other hand, maybe it's not `Target` that should contain a list of nodes, but the `TargetedMessage`, that should contain a list of targets?

Anyway, I think that's beyond the scope of this PR and needs some design and discussion first. I'd just add `Target::Observers` for now.

vkomenda:
> Anyway, I think that's beyond the scope of this PR and needs some design and discussion first. I'd just add Target::Observers for now.

Take care to filter the non-observers from `Target::Node` messages in that case to avoid overlap with `Target::Observers`.

pawanjay176:
So just to clarify, what you guys are suggesting is having an additional 

`Observers(Vec<N>)` type in the `Target` enum. After doing that, we would have to add a `match` arm to `defer_messages` in `traits.rs`. In there, we would have to filter out the nodes that we have passed in the `Observers` list and all observer nodes from the `peer_epochs` map. But how do we check if a node in the `peer_epochs` map is an observer or a validator in that function?

afck:
No, just `Observers`, without parameters. The `Broadcast` protocol should not have to know about specific observers.

But that's a good point: I'm not sure whether we have the information in `traits.rs` about who is or isn't an observer…
We could add a target `AllExcept(Vec<N>)` instead, so `Broadcast` could put into the target the nodes that it _does_ know about. Not a nice solution, I know… :grimacing:

pawanjay176:
My mistake. I thought you meant something like `NodesPlusObservers(Vec<N>)` where the `SenderQueue` would send it to all the passed nodes + all observer nodes. So instead of having the loop for the `Echo` and `EchoHash` messages, we just do `Target::NodesPlusObservers(nodes)`. 

But again, I don't think `traits.rs` has info on who is or isn't a validator. `AllExcept` seems to work but is kind of hacky :sweat_smile:

vkomenda:
> My mistake. I thought you meant something like NodesPlusObservers(Vec<N>) where the SenderQueue would send it to all the passed nodes + all observer nodes. So instead of having the loop for the Echo and EchoHash messages, we just do Target::NodesPlusObservers(nodes).

Right, that sounds like my initial reading of the new `Target` variant which @afck said could be a different PR where we could make a `TargetedMessage` to contain a list of recipients. That also has some unclear points to me:

- Do we have to ensure that that list of recipients is non-empty? E.g., 
```
pub struct TargetedMessage<M, N> {
    pub target: Target<N>,
    pub more_targets: Vec<Target<N>>,
    pub message: M,
}
```
- If the intended list of targets may include extra nodes as defined by the `Target` variant (e.g., extra observer nodes as decided by the sender queue), wouldn't that be clearer to actually record that extra hint in `TargetedMessage` as well? Or maybe not. I'm not so sure about this one.

Anyhow, if `AllExcept` repairs the tests then let's go ahead with it. The resulting set operation seems to be the same, but with a set difference added inside. However, maybe I'm complicating things and no difference is needed? Otherwise we would be filtering `all_ids` quite a bit.

pawanjay176:
Another thing we could possibly do would be to have an additional function in the `ConsensusProtocol` trait.
```rust
fn is_node_validator(&self) -> bool;
``` 
Every implementation of the `ConsensusProtocol` trait ought to have an `Arc<NetworkInfo<N>>` from which we can query if the passed Node is a validator or not. Though I'm not super convinced if having an `is_node_validator` in the `ConsensusProtocol` trait makes sense.

vkomenda:
> Another thing we could possibly do would be to have an additional function in the `ConsensusProtocol` trait.

Sounds dubious. `ConsensusProtocol` instances don't represent nodes. Can you use `NetworkInfo::is_validator()` instead?

afck:
> Do we have to ensure that that list of recipients is non-empty?

I don't think so. I'd prefer if `Broadcast` simply didn't have to know whether there are observers and who they are. It should just be able to occasionally say "send this to the observers, too, if there are any", so that it's guaranteed that they can also verify the protocol.
So e.g. in the situation where it already sent full `Echo`s to all its fellow validators and is now ready to decode and terminate, it would want to send an additional `Echo` to any observers, to guarantee that they can decode, too. (Not sure whether we'd want to do that in the beginning or in the end.)
In that case, it would return an `Echo` with target `AllExcept(list_of_all_the_validators)`. If there is no observer, that message would go nowhere, which is fine.

`AllExcept` is not elegant, but I think it's the simplest solution for now.

vkomenda:
> > Do we have to ensure that that list of recipients is non-empty?

> I don't think so. I'd prefer if Broadcast simply didn't have to know whether there are observers and who they are.

You've answered a slightly different question but I'm with you on this one :)

`AllExcept(Vec<N>)` does fit into the current `TargetedMessage` well.

src/traits.rs
```diff
@@ -283,6 +283,30 @@ where
                         }
                     }
                 }
+                Target::AllExcept(known) => {
```
afck: I wouldn't call the argument `known`: It's not necessarily all nodes known to the algorithm. Maybe `exclude`?

pawanjay176: Will change it to `exclude`.

pawanjay176:
@afck ran the simulation with the new code. I'm getting ~20% improvement in average message size handled by a node for message sizes > 100 bytes. I think the reason for this sub optimal improvement is because in many instances, a node receives `2f + 1` `Ready`s before it receives `CanDecode` from its 'right' nodes. Hence, it ends up sending the `Echo`s to those nodes before terminating even if those nodes have enough shards to decode. Does that make sense?

afck:
Awesome, thanks! That makes sense and I think 20% already means it's worth making this optimization! :tada: 
I imagine that if you simulate a larger batch size, lower bandwidth and more nodes, the number might go up further? But maybe not. Of course the 66% improvement is only in an unrealistic best case scenario where everything apart from the payload is negligible, and everyone manages to send their `CanDecode` before they get their `Echo`.

pawanjay176:
Added a few minor optimizations to ensure more `CanDecode` messages are delivered.

1. **Send `EchoHash` messages before `Echo` messages** - We send a `CanDecode` message to all nodes who have sent us a `EchoHash` message after we have the minimum shares to decode. This is to ensure that we have as many `EchoHash` messages possible before we start sending out `CanDecodes`. If `EchoHash` messages are delivered later than `Echo`s, it is possible that a node sends `Ready` and terminates before it sent out necessary `CanDecodes`. This will lead to more heavy `Echo` messages being sent.
2. **Modify `send_can_decode` to set can_decode_sent = true only after receiving *all* `EchoHash`s: Earlier, we set `can_decode_sent=true` after we have minimum shares and any number of `EchoHash`s and Modify `handle_echo_hash` to call `send_can_decode` everytime we receive a new `EchoHash`**:  This is to ensure that we send all the `EchoHash`s possible before terminating.

afck:
I don't agree with those two changes:
We should just send `CanDecode` as soon as we have _N - 2 f_ `Echo`s, and we should send it to everyone who hasn't sent us an `Echo`.
Whether they've sent us an `EchoHash` or not doesn't matter.

pawanjay176:
You are right. I overcomplicated the issue for no reason. I just needed to send `CanDecode` to everyone who didn't send us `Echo` instead of sending to everyone who sent us `EchoHash`.

Should I reset the changes and force push the fix or revert and apply the fix on top of the commit?

afck:
Sure, feel free to force-push to remove the latest commit.

src/broadcast/mod.rs
```diff
@@ -39,12 +40,40 @@
...
+//! We use a modified version of the algorithm to save on bandwith which provides the same
+//! security guarantees as the original. The main idea of the optimized algorithm is that in the
+//! optimisitic case, a node only needs _N - 2 f_ chunks to decode a value and every additional
+//! `Echo` message over that is wasteful.
+//! The modified algorithm introduces two new message types:
+//! * `CanDecode(h)` - Indicates node has enough shards to recover message with merkle root `h`.
+//! * `EchoHash(h)` - Indicates node can send an `Echo(p[i])` message upon request.
+//!
+//! Let `g` be the `fault_estimate` i.e. the estimate of number of faulty nodes in the network
+//! that we want to optimize for.
```
afck: We should state explicitly that the algorithm still works correctly if more than `g` are faulty.
Also, let's be consistent with the underscores (italic) and backticks (fixed width): I'd use underscores for formulas involving the single-letter identifiers (like _N_, _g_, _f_, …) and backticks only for identifiers that occur in the code or pseudocode (`Echo(p[i])`).

src/broadcast/mod.rs
```diff
@@ -39,12 +40,40 @@
...
+//! * A validator that has received _N - f_ `Echo`s plus `EchoHash`s **or** _f + 1_ `Ready`s with root hash `h`,
+//! sends `Ready(h)` to everyone.
+//! * A validator that has received _N - 2 f_ `Echo`s with root hash `h`, sends `CanDecode(h)` to all nodes
+//! who haven't sent them a full `Echo` message.
+//! * A validator that has received _2 f + 1_ `Ready`s with root hash `h` sends a full  `Echo` message to the
+//! remaining nodes who haven't sent them `CanDecode(h)`.
+//! * A node that has received _2 f + 1_ `Ready`s **and** _N - 2 f_ `Echo`s with root hash `h`
+//! decodes and outputs the value, and then terminates.
+//!
+//!
```
afck: Duplicate new line.

src/broadcast/broadcast.rs
```diff
@@ -248,16 +281,83 @@ impl<N: NodeIdT> Broadcast<N> {
         let hash = *p.root_hash();
 
         // Save the proof for reconstructing the tree later.
-        self.echos.insert(sender_id.clone(), p);
+        self.echos.insert(sender_id.clone(), EchoContent::Full(p));
+
+        let mut step = Step::default();
+
+        // Upon receiving `N - 2f` `Echo`s with this root hash, send `CanDecode`
+        if !self.can_decode_sent && self.count_echos_full(&hash) >= self.coding.data_shard_count() {
```
afck: Do we need to store `can_decode_sent` by hash?
Because we might have to send more than one of them, e.g. if the proposer is faulty, it might send a value _v<sub>1</sub>_ with hash _h<sub>1</sub>_ to _N - f_ nodes _and_ value _v<sub>2</sub>_ with hash _h<sub>2</sub>_ to _f + 1_ nodes, with the overlapping nodes all malicious. So you may end up receiving _f + 1_ `Echo`s with _h<sub>2</sub>_, even though in the end _v<sub>1</sub>_ will get accepted. In that case, you should send `CanDecode` with both hashes eventually.

src/broadcast/broadcast.rs
```diff
@@ -284,21 +384,129 @@ impl<N: NodeIdT> Broadcast<N> {
...
         let echo_msg = Message::Echo(p.clone());
-        let step: Step<_> = Target::All.message(echo_msg).into();
+        let mut step = Step::default();
+        // `N - 2f + g` node ids to the left of our_id (excluding our_id)
+        // after arranging all node ids in a circular list.
+        let left = self
+            .netinfo
+            .all_ids()
+            .cycle()
+            .skip_while(|x| *x != self.our_id())
```
afck: `.skip(self.netinfo.node_index(self.our_id()))` should be faster and equivalent? (Also below.)

src/network_info.rs
```diff
@@ -82,10 +82,22 @@ impl<N: NodeIdT> NetworkInfo<N> {
 
     /// ID of all nodes in the network.
     #[inline]
-    pub fn all_ids(&self) -> impl Iterator<Item = &N> {
+    pub fn all_ids(&self) -> impl Iterator<Item = &N> + Clone {
```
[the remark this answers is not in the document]
pawanjay176: It is required for applying the `cycle` method when constructing the `left` and `remaining` iterators.

src/broadcast/broadcast.rs
pawanjay176: Yeah you are right. Will fix this.

src/broadcast/broadcast.rs
pawanjay176: `node_index` returns an `Option<usize>`. This way seems easier rather than unwrapping.

vkomenda:
Yet another failing `Broadcast` seed (possibly related to the one above:
```
cc c7c972d3cefc07093bdde048ff8a005dc72f6afca9911131bda6172511a0976c # shrinks to seed = [15, 50, 223, 151, 165, 51, 17, 57, 238, 60, 25, 199, 137, 121, 37, 120]
```

afck:
Great, thanks! Yes, we should focus on the failing _broadcast_ test first then, since it's probably the reason for the failing DHB test. :+1:

afck:
@vkomenda: If I put that line in `tests/broadcast.proptest-regressions`, the tests still pass. Or is that seed for a different test?

vkomenda:
Strange... That is my `broadcast.proptest-regresstions`.

pawanjay176:
@vkomenda the tests pass for me too for the seed that you provided.
looking into why the seed @afck provided is failing.

pawanjay176:
@afck the test is failing because the time_limit was exceeded. Removed the time_limit and the tests passed.
Checking why that particular test is taking so freakishly long now.

afck:
It make sense that it takes _a bit_ longer: We are sending a few more messages now.
But if there is a huge difference to `master`, we should find out the cause, yes.

@vkomenda: Are you using @pawanjay176's current `optimized_broadcast` branch?

vkomenda:
> @vkomenda: Are you using @pawanjay176's current optimized_broadcast branch?

Yes. I'm also running the tests with `use-insecure-test-only-mock-crypto`. `test_broadcast_random_delivery_adv_propose` fails quickly, without a timeout.

vkomenda:
That failing test actually passes without `use-insecure-test-only-mock-crypto`.

afck:
You're right, thanks! I can reproduce it now.

afck:
OK, I found the bug: When sending the full `Echo`s at the end we were filtering out _everyone_ instead of _no one_ if we hadn't received any `CanDecode`s yet. This fixes the test for me:
```diff
--- a/src/broadcast/broadcast.rs
+++ b/src/broadcast/broadcast.rs
@@ -134,6 +134,7 @@ impl<N: NodeIdT> Broadcast<N> {
         if !self.netinfo.is_node_validator(sender_id) {
             return Err(Error::UnknownSender);
         }
+        debug!("{}: Handling {:?} from {:?}", self, &message, sender_id);
         match message {
             Message::Value(p) => self.handle_value(sender_id, p),
             Message::Echo(p) => self.handle_echo(sender_id, p),
@@ -443,21 +444,20 @@ impl<N: NodeIdT> Broadcast<N> {
         let echo_msg = Message::Echo(p);
         let mut step = Step::default();
 
-        if let Some(senders) = self.can_decodes.get(hash) {
-            // Remaining node ids to the right of our_id
-            // after arranging all node ids in a circular list.
-            let right = self
-                .netinfo
-                .all_ids()
-                .cycle()
-                .skip_while(|x| *x != self.our_id())
-                .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
-                .take_while(|x| *x != self.our_id());
-            let msgs = right
-                .filter(|id| !senders.contains(id))
-                .map(|id| Target::Node(id.clone()).message(echo_msg.clone()));
-            step.messages.extend(msgs);
-        }
+        let senders = self.can_decodes.get(hash);
+        // Remaining node ids to the right of our_id
+        // after arranging all node ids in a circular list.
+        let right = self
+            .netinfo
+            .all_ids()
+            .cycle()
+            .skip_while(|x| *x != self.our_id())
+            .skip(self.netinfo.num_correct() - self.netinfo.num_faulty() + self.fault_estimate)
+            .take_while(|x| *x != self.our_id());
+        let msgs = right
+            .filter(|id| senders.map_or(true, |s| !s.contains(id)))
+            .map(|id| Target::Node(id.clone()).message(echo_msg.clone()));
+        step.messages.extend(msgs);
         Ok(step)
     }
```

afck:
(With this change, it passed one hour of the test suite in a loop. :tada:)

pawanjay176:
Awesome! Will push the change.
Why did the broadcast test fail only for `mock-crypto` and not for the normal case though?

afck:
It would eventually have failed with normal crypto, too, just not with the same seed. The crypto implementations create different keys, so the order of the keys is different, which changes iteration order in some places, I think.

pawanjay176:
Ohh I thought it would be same for same seeds. Makes sense now.

PR #410, opened 2019-06-18
Change the `Target` variants.

### LINKED ISSUE #408: Change the `Target` variants for less cloning and less duplicated code.
With #405, `Target::All` is somewhat redundant, since it is identical to `Target::AllExcept(BTreeSet::new())`.

Possibly we should also replace the `Node` variant with a `Nodes(BTreeSet<N>)`, to avoid unnecessarily cloning messages.

And we should double-check whether `BTreeSet` is the right choice; often `Vec` is faster.

### LINKED ISSUE #415: Vulnerability in memoffset.
`cargo audit` warns:
```
error: Vulnerable crates found!

ID:	 RUSTSEC-2019-0011
Crate:	 memoffset
Version: 0.2.1
Date:	 2019-07-16
URL:	 https://github.com/Gilnaa/memoffset/issues/9#issuecomment-505461490
Title:	 Flaw in offset_of and span_of causes SIGILL, drops uninitialized memory of arbitrary type on panic in client code
Solution: upgrade to: >= 0.5.0

error: 1 vulnerability found!
```
However, this is an indirect dependency:
```
$ cargo tree -ip memoffset
memoffset v0.2.1
└── crossbeam-epoch v0.7.1
    ├── crossbeam v0.7.1
    │   [dev-dependencies]
    │   └── hbbft v0.1.1 (/home/andreas/git/hbbft)
    │       └── hbbft_testing v0.1.0 (/home/andreas/git/hbbft/hbbft_testing)
    │           [dev-dependencies]
    │           └── hbbft v0.1.1 (/home/andreas/git/hbbft) (*)
    ├── crossbeam-deque v0.6.3
    │   ├── rayon v1.1.0
    │   │   └── reed-solomon-erasure v3.1.1
    │   │       └── hbbft v0.1.1 (/home/andreas/git/hbbft) (*)
    │   └── rayon-core v1.5.0
    │       └── rayon v1.1.0 (*)
    └── crossbeam-deque v0.7.1
        └── crossbeam v0.7.1 (*)
```
Let's disable `cargo audit` in `ci.sh` for now, until we can fix this.

Hopefully we won't need to do anything, if crossbeam releases a minor version with the fix: https://github.com/crossbeam-rs/crossbeam/pull/402

afck:
`Target` now only has a `Nodes` and an `AllExcept` variant, to specify
a message's target via a whitelist or blacklist. This avoids cloning
the message content and simplifies the code in several places.

Closes #408, closes #415.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -285,9 +285,7 @@ where
                             }
                         }
                     }
-                    if exclude.len() < peer_epochs.len() {
-                        passed_msgs.push(Target::AllExcept(exclude).message(msg.message));
-                    }
+                    passed_msgs.push(Target::AllExcept(exclude).message(msg.message));
                 }
             }
         }
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -255,56 +255,38 @@ where
         let mut passed_msgs: Vec<_> = Vec::new();
         for msg in self.messages.drain(..) {
             match msg.target.clone() {
-                Target::Node(id) => {
-                    if let Some(&them) = peer_epochs.get(&id) {
-                        if msg.message.is_premature(them, max_future_epochs) {
-                            deferred_msgs.push((id, msg.message));
-                        } else if !msg.message.is_obsolete(them) {
-                            passed_msgs.push(msg);
-                        }
-                    }
-                }
-                Target::All => {
-                    let is_accepted = |&them| msg.message.is_accepted(them, max_future_epochs);
+                Target::Nodes(mut ids) => {
                     let is_premature = |&them| msg.message.is_premature(them, max_future_epochs);
                     let is_obsolete = |&them| msg.message.is_obsolete(them);
-                    if peer_epochs.values().all(is_accepted) {
-                        passed_msgs.push(msg);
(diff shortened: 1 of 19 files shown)


afck:
That finally worked: I increased the Travis timeout.

src/traits.rs
```diff
@@ -255,57 +255,39 @@ where
...
-                            } else if !is_obsolete(them) {
-                                passed_msgs
-                                    .push(Target::Node(id.clone()).message(msg.message.clone()));
+                                exclude.insert(id.clone());
+                            } else if is_obsolete(them) {
+                                exclude.insert(id.clone());
                             }
                         }
                     }
+                    if exclude.len() < peer_epochs.len() {
```
vkomenda: Could it be that there is a node that is not known through `peer_epochs` but which needs to observe `msg.message`? In other words, maybe pushing the `Target::AllExcept(exclude)` message unconditionally would be better?

afck: You're right, that way we still allow for observer nodes unknown to the sender queue.
Done.

PR #414, opened 2019-07-23
Fix unused error in test and fix markdown format error

kigawas:


kigawas:
@afck 
Formated by https://github.com/DavidAnson/vscode-markdownlint

afck:
Thanks! There are also [Node.js](https://github.com/DavidAnson/markdownlint) and [Ruby](https://github.com/markdownlint/markdownlint) versions. I made a note for later to consider adding them to CI: https://github.com/poanetwork/hbbft/issues/416

The test failure is due to a [security vulnerability](https://github.com/poanetwork/hbbft/issues/415) found by `cargo audit`. I'll look into it today…

afck:
Wait… it actually isn't!
```
running 1 test
test run_binary_agreement ... FAILED
```

afck:
The problem is that your change in `sbv_broadcast.rs` actually _did_ change behavior: If `2 * self.netinfo.num_faulty() + 1` is equal to `self.netinfo.num_faulty() + 1`, both branches should be executed, which is why we didn't use an `else` here.

kigawas:
Ah, thanks for pointing out. Already fixed.

afck:
I'm "fixing" CI in #417; please rebase once that's merged. Sorry for the inconvenience!

afck:
Merged; you can rebase now.

PR #417, opened 2019-07-23
Add HoneyBadger::skip_to_epoch and NetworkInfo::other_ids.

### LINKED ISSUE #413: Add a method for skipping the current epoch.
We could add a `*HoneyBadger::skip_epoch` method that would delete all information about the current epoch and move on to the next, without ever outputting the skipped epoch's batch.

This is dangerous, of course, and needs to be well-documented: It must only be called if the upper layer can guarantee that every instance that hasn't output the batch yet will call this method.

An example is the use in a blockchain: If our `HoneyBadger` instance is still working on block number 42, but we've just imported the complete, finalized block 42, we can probably be sure that every node will eventually import it and can call `skip_epoch`.

afck:
This also disables `cargo audit` until #415 is fixed.

Closes #413.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-    - 1.35.0
+    - 1.36.0
 cache:
   cargo: true
   timeout: 1200
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ _**Note:** Additional examples are currently in progress._
 
 ### Build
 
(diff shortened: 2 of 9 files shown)


afck:
I realized my `cargo audit` was outdated. The new version has an `--ignore` option, so we can still run it in CI:
```
cargo audit --ignore RUSTSEC-2019-0011
```

PR #423, opened 2019-08-22
Remove NetworkInfo from Broadcast and keys from NetworkInfo.

### LINKED ISSUE #311: Leave threshold cryptography out of Broadcast.
The `Broadcast` algorithm can also be used as stand-alone, independently of Honey Badger. However, it currently requires a `NetworkInfo`, with threshold key shares and all.

To make it easier to use in a context without threshold encryption, it should be possible to instantiate `Broadcast` without any keys. All it needs to know is the set of node IDs (and possibly observer IDs).

afck:
Closes #311.

### DIFF
--- a/src/network_info.rs
+++ b/src/network_info.rs
@@ -2,7 +2,7 @@ use std::borrow::Borrow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::sync::Arc;
 
-use crate::crypto::{self, PublicKey, PublicKeySet, PublicKeyShare, SecretKey, SecretKeyShare};
+use crate::crypto::{self, PublicKeySet, PublicKeyShare, SecretKeyShare};
 use rand;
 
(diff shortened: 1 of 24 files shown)


src/network_info.rs
```diff
@@ -1,33 +1,102 @@
...
+    N: NodeIdT,
+{
+    fn from(i: I) -> Self {
+        let indices: BTreeMap<N, usize> = i
+            .into_iter()
+            .enumerate()
+            .map(|(n, id)| (id.borrow().clone(), n))
+            .collect();
+        let num_faulty = util::max_faulty(indices.len());
+        assert!(3 * num_faulty < indices.len(), "3 f >= N. This is a bug!");
```
vkomenda: Instead of `<` you probably meant `>=`.

vkomenda: No, actually the code is correct. Nevermind!

PR #425, opened 2019-09-03
Clean up threshold sign tests.

afck:
Add a test that actually verifies the signature, and refactor the tests a bit.

### DIFF
--- a/tests/threshold_sign.rs
+++ b/tests/threshold_sign.rs
@@ -3,7 +3,7 @@
 
 use std::sync::Arc;
 
-use hbbft::{crypto::Signature, threshold_sign::ThresholdSign, util, ConsensusProtocol};
+use hbbft::{threshold_sign::ThresholdSign, util, ConsensusProtocol};
 use hbbft_testing::adversary::{Adversary, NodeOrderAdversary, ReorderingAdversary};
 use hbbft_testing::proptest::{gen_seed, TestRng, TestRngSeed};
 use hbbft_testing::{NetBuilder, NewNodeInfo, VirtualNet};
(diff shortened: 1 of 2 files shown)


tests/threshold_sign.rs
```diff
@@ -124,24 +156,20 @@ proptest! {
     })]
 
     #[test]
-    #[allow(clippy::unnecessary_operation)]
     fn test_threshold_sign_random_silent_200_samples(seed in gen_seed()) {
```
vkomenda: This function should probably now be called `do_test_coin_random_silent_200_samples`.

tests/threshold_sign.rs
```diff
@@ -124,24 +156,20 @@ proptest! {
...
-        do_test_threshold_sign_random_silent_200_samples(seed)
+        let new_adversary = || ReorderingAdversary::new();
+        test_coin_different_sizes(new_adversary, 200, seed);
     }
 
     #[test]
-    #[allow(clippy::unnecessary_operation)]
     fn test_threshold_sign_first_silent_50_samples(seed in gen_seed()) {
```
vkomenda: This function should probably now be called `do_test_coin_random_silent_50_samples`.

tests/threshold_sign.rs
```diff
@@ -13,30 +13,65 @@ use rand::{Rng, SeedableRng};
...
+        // Handle messages until all good nodes have terminated.
+        while !net.nodes().all(|node| node.algorithm().terminated()) {
+            let _ = net.crank_expect(&mut rng);
+        }
+        let node0 = net.correct_nodes().nth(0).unwrap();
+        // Verify that all instances output the same value.
+        assert_eq!(1, node0.outputs().len());
+        assert!(net.nodes().all(|node| node.outputs() == node0.outputs()));
+        let pk = node0.algorithm().netinfo().public_key_set().public_key();
+        assert!(pk.verify(&node0.outputs()[0], "message"));
```
vkomenda: Maybe it's a minor detail but the expected value `"message"` could be cloned from the initial one to guarantee equality.

tests/threshold_sign.rs
afck: Done. (But without the `do_`.)

PR #433, opened 2020-03-11
Fix QHB loop with 1 validator.

### LINKED ISSUE #432: Infinite loop in QueueingHoneyBadger
I'm trying to use the library, and seem to have encountered an infinite loop when calling `handle_input` on an instance of QueueingHoneyBadger with a single validator, in particular in the propose function.
https://github.com/poanetwork/hbbft/blob/4857b7f9c7a0f513caca97c308d352c6a77fe5c2/src/queueing_honey_badger/mod.rs#L341-L353

The `can_propose` function keeps returning true even after the step is extended with the contents of the current queue. It seems that the items proposed are not removed from the queue, and this is intended behavior, but the `dyn_hb.has_input()` isn't returning true properly.
https://github.com/poanetwork/hbbft/blob/4857b7f9c7a0f513caca97c308d352c6a77fe5c2/src/queueing_honey_badger/mod.rs#L333-L338

This seems to be because the internal `HoneyBadger` instance in dyn_hb is setting `has_input` when proposing, but is then unsetting it and advancing the epoch during `try_output_batches` and QHB and DHB expect it to still be set.
https://github.com/poanetwork/hbbft/blob/4857b7f9c7a0f513caca97c308d352c6a77fe5c2/src/honey_badger/honey_badger.rs#L88-L95

https://github.com/poanetwork/hbbft/blob/4857b7f9c7a0f513caca97c308d352c6a77fe5c2/src/honey_badger/honey_badger.rs#L164-L185

afck:
QHB needs to remove the output from the queue, even if it's during a `propose` call. In the case of a single validator, the `propose` call always immediately outputs, so forgetting about this caused an infinite loop.

The QHB test is extended to the case of 2 validators, which also tests a single one, since the QHB test removes and re-adds one of its validators.

Closes #432

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-    - 1.36.0
+    - 1.42.0
 cache:
   cargo: true
   timeout: 2400
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ travis-ci = { repository = "poanetwork/hbbft" }
 [dependencies]
 bincode = "1.2.0"
 byteorder = "1.3.2"
-derivative = "1.0.3"
+derivative = "2.0.2"
 env_logger = "0.7.1"
 failure = "0.1.6"
 hex_fmt = "0.3"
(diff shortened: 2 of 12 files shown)


afck:
Looks like some dependencies used unstable features that don't work anymore, and others violated semver. I'll work on fixing that…

jlandrews:
Tested proposed fix against my code and it does resolve the loop. Thanks! :+1:

PR #434, opened 2020-03-20
Version 0.2.0, using threshold_crypto 0.4.0

afck:


afck:
I'll update this once https://github.com/poanetwork/threshold_crypto/pull/102 is merged. But before we do that, let's make sure this version will work with OpenEthereum.

